PR #7, opened 2017-04-06
Make `items` optional for `connectHitsPerPage`

bjrnt:
**Summary**

When creating a virtual `HitsPerPage` widget, the developer will only provide a `defaultRefinement` and no items as obviously no items will be rendered.

To achieve this effect you currently have to do something like:
`
<VirtualHitsPerPage items={[]} defaultRefinement={this.props.hitsPerPage} />
`
this PR allows you to simplify this to:
`
<VirtualHitsPerPage defaultRefinement={this.props.hitsPerPage} />
`

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectHitsPerPage.js
+++ b/packages/react-instantsearch/src/connectors/connectHitsPerPage.js
@@ -33,7 +33,7 @@ function getCurrentRefinement(props, searchState, context) {
  * @name connectHitsPerPage
  * @kind connector
  * @propType {number} defaultRefinement - The number of items selected by default
- * @propType {{value: number, label: string}[]} items - List of hits per page options.
+ * @propType {{value: number, label: string}[]} [items] - List of hits per page options.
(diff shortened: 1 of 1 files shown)


vvo:
Hi @bjrnt I think ultimately what you could be using is this:

```jsx
<Configure hitsPerPage={20} />
```

Rather than doing a VirtualRefinement for hitsPerPage. This seems easier.

What do you think?

If you ever saw that VirtualRefinement was used for hitsPerPage or advised to be used, please let me know where so we update the documentation.

bjrnt:
@vvo 
That does seem simpler.

After looking at the example [here](https://community.algolia.com/react-instantsearch/widgets/HitsPerPage.html), trying to do the same with a VirtualRefinement seemed like the most obvious way forward. 

Also, in the example I linked to `items` are not given even though it is listed as a required prop, so I assumed that both widgets were meant to work with only a `defaultRefinement`.

vvo:
I'll update this part, thanks for the heads up! Sorry for confusion :)

PR #11, opened 2017-04-10
feat(Highlighter): allow rendering to custom tag

Haroenv:
**Summary**

`<em>` isn't actually very suited for highlighting, a `<mark>` tag is way better, because it brings its own accessibility, default style etc. with it. For legacy reasons we can't just change the default tag, since that would break all implementations, but it is something we can do in a future major release.

To allow this, a new prop (`HighlightTagName`) was added to Highlight, Snippet and Highlighter. 

See also https://github.com/algolia/instantsearch.js/pull/2092

Compared to the previous pull request I also added a small paragraph in the [guide](https://github.com/algolia/react-instantsearch/compare/feat/custom-highlight-tag?expand=1#diff-dd7bc35948fc032c33147e1ef644ba4a) on why to change the tagname.

I didn't add a deprecation warning to rendering Highlighter without a tagname, because there are lots of use cases where you'd want a custom tag, so I don't think there should be a warning.

**Result**

I added stories for `Highlight` and `Snippet` to make this clear.

### DIFF
--- a/packages/react-instantsearch/src/components/Highlighter.test.js
+++ b/packages/react-instantsearch/src/components/Highlighter.test.js
@@ -44,4 +44,43 @@ describe('Highlighter', () => {
     );
     expect(tree.toJSON()).toMatchSnapshot();
   });
+
+  it('renders a hit with a custom tag correctly', () => {
+    const hitFromAPI = {
+      objectID: 0,
+      deep: { attribute: { value: 'awesome highlighted hit!' } },
+      _highlightProperty: {
+        deep: {
+          attribute: {
+            value: {
+              value: 'awesome <ais-highlight>hi</ais-highlight>ghlighted <ais-highlight>hi</ais-highlight>t!',
+              fullyHighlighted: true,
+              matchLevel: 'full',
+              matchedWords: [''],
+            },
+          },
+        },
+      },
+    };
+
+    const highlight = ({ hit, attributeName, highlightProperty }) =>
+      parseAlgoliaHit({
+        preTag: '<ais-highlight>',
+        postTag: '</ais-highlight>',
+        attributeName,
+        hit,
+        highlightProperty,
+      });
+
+    const tree = renderer.create(
+      <Highlighter
+        attributeName="deep.attribute.value"
+        hit={hitFromAPI}
+        highlight={highlight}
+        highlightProperty="_highlightProperty"
+        tagName="mark"
+      />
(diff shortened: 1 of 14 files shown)


packages/react-instantsearch/src/components/Highlighter.test.js
```diff
@@ -44,4 +44,43 @@ describe('Highlighter', () => {
     );
     expect(tree.toJSON()).toMatchSnapshot();
   });
+
+  it('renders a hit with a custom tag correctly', () => {
```
mthuret: Why just modifying the Highlighter test and not the Highlight/Snippet?

stories/Highlight.stories.js
```diff
@@ -0,0 +1,45 @@
...
+  hit: React.PropTypes.object.isRequired,
+};
+
+stories
+  .add('default', () => (
+    <WrapWithHits hasPlayground={false} linkedStoryGroup="Highlight">
+      <Hits hitComponent={Default} />
+    </WrapWithHits>
+  ))
+  .add('custom tag', () => (
```
mthuret: I would probably call it playground. Might not be the best name but emphasis/consistent with stories that allows to play with props.

Haroenv: I noticed that the link to "play with props" doesn't work on this, but I'm not really sure why

packages/react-instantsearch/src/components/Highlighter.test.js
Haroenv: Added the same type of test to the Highlight and Snippet tests.

stories/Highlight.stories.js
```diff
@@ -0,0 +1,45 @@
...
+  </p>
+);
+
+StrongHits.propTypes = {
+  hit: React.PropTypes.object.isRequired,
+};
+
+stories
+  .add('default', () => (
+    <WrapWithHits hasPlayground={false} linkedStoryGroup="Highlight">
```
mthuret: If you put this prop to true, then the link should appear and it should redirect to the playground story.

At first, those props that indice if you want the "play with props" and "view source code" links where made for the `default` story only because it was embed as an iframe on our doc website. 

I think we should still keep those links only for the default one (being the entry point).

Haroenv: Ah thanks, I misunderstood the point of this prop 😄

PR #12, opened 2017-04-10
fix(InstantSearch): dont fire request/onsearchStateChange when unmounting

mthuret:
See https://github.com/algolia/instantsearch.js/issues/2061

Also fix issue with SPA => you want to move to another URL, but because of the unmount, `onSearchStateChange` is called and then you go back.


packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -1,14 +1,14 @@
-import React, { PropTypes, Component, Children } from 'react';
+import React, {PropTypes, Component, Children} from 'react';
```
Haroenv: I think you're not running the post-commit `prettier` hook, or with other settings.

PR #29, opened 2017-04-12
fix(util): remove empty key was removing non object key

mthuret:
remove empty key wasn't working correctly as it removed key that were not objects.

### DIFF
--- a/packages/react-instantsearch/src/core/utils.js
+++ b/packages/react-instantsearch/src/core/utils.js
@@ -1,4 +1,4 @@
-import { isEmpty } from 'lodash';
+import { isEmpty, each, isPlainObject } from 'lodash';
 
 // From https://github.com/reactjs/react-redux/blob/master/src/utils/shallowEqual.js
 export function shallowEqual(objA, objB) {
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch/src/core/utils.js
```diff
@@ -64,14 +64,12 @@ export const defer = f => {
...
-  Object.keys(obj).forEach(
-    key =>
-      (obj[key] &&
-        typeof obj[key] === 'object' &&
-        !isEmpty(obj[key]) &&
-        removeEmptyKey(obj[key])) ||
-      obj[key] === undefined ||
-      (isEmpty(obj[key]) && delete obj[key])
-  );
+  each(obj, (v, k) => {
```
Haroenv: why not `Object.keys(obj).forEach()`? It's longer to write but more standard and less lodash 😄

packages/react-instantsearch/src/core/utils.js
```diff
@@ -64,14 +64,12 @@ export const defer = f => {
...
+    if (isEmpty(v) && isObject(v)) {
+      delete obj[k];
+    } else if (isObject(v)) {
```
Haroenv: I'd use `isPlainObject` instead of `isObject` here.

packages/react-instantsearch/src/core/utils.js
mthuret: Yes true :)

packages/react-instantsearch/src/core/utils.js
mthuret: Which corner case this will fix?

Haroenv: * isObject: Checks if value is the language type of Object. (e.g. arrays, functions, objects, regexes, new Number(0), and new String(''))
* isPlainObject: Checks if value is a plain object, that is, an object created by the Object constructor or one with a [[Prototype]] of null.

So most of the things are Objects, but in this case we only want "real" objects (maps)

packages/react-instantsearch/src/core/utils.js
```diff
@@ -1,4 +1,4 @@
-import { isEmpty } from 'lodash';
+import { isEmpty, each, isPlainObject } from 'lodash';
```
Haroenv: lodash `each` import isn't needed now, right?

mthuret: Arf. The linter should have detect this...

PR #30, opened 2017-04-12
chore: remove jest notifications

mthuret:
See: facebook/jest#3177

On OSX and watch mode, if a test is failing, the terminal-notifier process stays open.

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "node": "^6.x.x || ^7.x.x"
   },
   "jest": {
-    "notify": true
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -119,12 +119,9 @@
   "engines": {
     "node": "^6.x.x || ^7.x.x"
   },
-  "jest": {
-    "notify": true
```
Haroenv: What about `notify: false` to remind us that we want to enable it again later?

PR #35, opened 2017-04-13
chore(Travis): don't build docs

### LINKED ISSUE #33: Don't build the docs in Travis

A bit of both

Travis and Netlify both run the docs

Travis runs the tests, Netlify the docs

**Feature: What is your use case for such a feature?**

Faster builds

Change the command ran by Travis to only do the tests

4.0.0-beta.1

Haroenv:
**Summary**

Seeing if the docs are able to build can be done with Netlify now

closes #33

**Result**

Travis will run faster because it doesn't run the docs anymore

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: node_js
 node_js: stable
 before_script:
   - yarn boot
-branches:		
-  only:		
+branches:
+  only:
     - master
 cache:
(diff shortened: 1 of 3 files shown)


Haroenv:
Last commit should have crashed netlify, but it didn't. Can I see the logs/get credentials to log in to this netlify instance?

PR #38, opened 2017-04-14
chore(tests): run jest with --runInBand in CI

Haroenv:
**Summary**

Run jest with [--runInBand](http://facebook.github.io/jest/docs/cli.html#runinband) in CI. I heard a few times that this makes the tests faster on Travis, so let's give it a try

**Result**

before|after
---|---
±5 min | 3 min 55 sec
4 min 38 sec | 3 min 55 sec
||4 min 7 sec
||3 min 56 sec

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "release": "./scripts/release.sh",
     "lint": "if [ \"$CI\" = \"true\" ]; then eslint .; else eslint . --cache; fi",
     "docs:doctoc": "doctoc --maxlevel 1 README.md",
-    "test": "jest && yarn lint && NODE_ENV=production yarn test:build",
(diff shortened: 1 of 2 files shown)


mthuret:
Is the bench realised on travis or locally?

Haroenv:
Tested on Travis, I'm not 100% sure if it's always faster, but we can give it a try, and remove it again if it's not working as expected. 

some link: https://twitter.com/thejameskyle/status/852988092898099200

mthuret:
Yes I just want to make sure it's not slower locally. I prefer the CI to be a little bit slower than my last check with `npm run test` before pushing.

Haroenv:
It's only enabled on CI

Issue #15, opened 2017-04-11
Remove need for VirtualHits with connectAutoComplete


mthuret:
# Use case

The recipe demonstrating the connectAutoComplete uses `<VirtualHits/>`. It's because without them, no particular search parameters are computed for every `<Index/>`. 

# Need

This syntax is not very straightforward. Ideally we want to get rid of them. 

# Proposal

`<Index>` should be seen as a widget, like the `<Hits>`. That way it will be taken into account when computing search parameters.

PR #45, opened 2017-04-20
feat(MultiIndex): remove the need for virtual hits when using connectAutoComplete

mthuret:
see #15.

Now you can use `connectAutoComplete` like:

```js
<InstantSearch
      appId=""
      apiKey=""
      indexName="ikea"
    >
      <Configure hitsPerPage={1} />
      <Index indexName="bestbuy" />
      <Index indexName="airbnb" />
      <AutoComplete />
</InstantSearch>
```

### DIFF
--- a/packages/react-instantsearch/examples/autocomplete/src/App-Multi-Index.js
+++ b/packages/react-instantsearch/examples/autocomplete/src/App-Multi-Index.js
@@ -1,9 +1,6 @@
 import React from 'react';
 import { InstantSearch, Configure, Index } from 'react-instantsearch/dom';
-import {
-  connectAutoComplete,
-  connectHits,
-} from 'react-instantsearch/connectors';
+import { connectAutoComplete } from 'react-instantsearch/connectors';
 import Autosuggest from 'react-autosuggest';
 import 'react-instantsearch-theme-algolia/style.css';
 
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch/src/core/Index.js
```diff
@@ -58,8 +84,11 @@ Index.propTypes = {
 };
 
 Index.childContextTypes = {
-  // @TODO: more precise widgets manager propType
```
Haroenv: Why was this todo removed, is it solved?

packages/react-instantsearch/src/core/Index.test.js
```diff
@@ -0,0 +1,62 @@
...
+  it('exposes multi index context', () => {
+    const wrapper = mount(
+      <Index {...DEFAULT_PROPS}>
+        <div />
+      </Index>,
+      { context }
+    );
+
+    const childContext = wrapper.instance().getChildContext();
+    expect(childContext.multiIndexContext.targetedIndex).toBe('foobar');
```
Haroenv: `toBe(DEFAULT_PROPS.indexName)`

vvo:
There's no example of connectAutoComplete in your PR description, am I missing smg?

mthuret:
@vvo What do you mean? You have the recipe (link in the guide) and the story?

packages/react-instantsearch/examples/autocomplete/src/App-Multi-Index.js
```diff
@@ -15,17 +12,11 @@ const App = () => (
   >
     <AutoComplete />
     <Configure hitsPerPage={1} />
-    <Index indexName="bestbuy">
-      <VirtualHits />
-    </Index>
-    <Index indexName="airbnb">
-      <VirtualHits />
```
vvo: 👍

vvo:
LGTM, does it have any impact on other usages?

mthuret:
@vvo I don't think so. If you don't need the `connectAutoComplete` then you still use `<Index/>` as usual. The only thing now, is that if you're using it completely alone in that use case (but why?), then a request will be fired (right now it was fired if a least one widget was nested).

Issue #55, opened 2017-04-25
Multi-index search


gl0gl0:
Hi,

When implementing multi-index search, following the example:
https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/examples/multi-index/src/App.js

My code:
https://screencast.com/t/emslqzaPHH

The hit results are showing correctly for the second index (artists), but for the first index (songs) no results are being displayed. And I see that in the Network tab (Developer Tools), the REST calls shows hit results for both of them.

Any help would be very much appreciated.
Thanks!

mthuret:
Thanks for spotting this regression @gl0gl0! 

Right now, It will work with this syntax: 

```js
<InstantSearch
      appId=""
      apiKey=""
      indexName="index1"
    >
      <SearchBox />
      <Index indexName="index1">
        <CustomHits />
      </Index>
      <Index indexName="index2">
        <CustomHits />
      </Index>
    </InstantSearch>
```

But definitely the syntax you used should work! I will fix it, and it will be part of a new beta.

mthuret:
Should be fix with 4.0.0-beta.4

PR #56, opened 2017-04-25
fix(MultIndex): no need to nest hits, if those are from main index.

mthuret:
This was a regression due to: https://github.com/algolia/react-instantsearch/pull/36

See: https://github.com/algolia/react-instantsearch/issues/55

Allows this syntax: 

```js
<InstantSearch
      appId=""
      apiKey=""
      indexName="index1"
>
  <SearchBox />
      <CustomHits />
  <Index indexName="index2">
      <CustomHits />
  </Index>
</InstantSearch>
```

### DIFF
--- a/packages/react-instantsearch/src/core/indexUtils.js
+++ b/packages/react-instantsearch/src/core/indexUtils.js
@@ -7,8 +7,8 @@ export function getIndex(context) {
 }
 
 export function getResults(searchResults, context) {
-  if (hasMultipleIndex(context)) {
-    return searchResults.results && searchResults.results[getIndex(context)]
+  if (searchResults.results && !searchResults.results.hits) {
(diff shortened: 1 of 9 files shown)


packages/react-instantsearch/src/core/indexUtils.js
```diff
@@ -1,4 +1,4 @@
-import { has, omit, get } from 'lodash';
+import { has, omit, get, isObject } from 'lodash';
```
Haroenv: Why has this been added and not used?

mthuret: I believe that because of prettier we don't have this eslint rules anymore? That's annoying, do you think you can have a look about it? I will fix this.

Issue #52, opened 2017-04-24
ToggleWidget defaultChecked doesnt work


davidfurlong:
Im tired of filling and reading the issue guidelines. Title is self explanatory

mthuret:
Hi @davidfurlong 

This a typo on the documentation, you should use `defaultRefinement={true}`.

Thanks for spotting this, I will update the documentation and produce a new story to demonstrate it.

PR #57, opened 2017-04-25
docs: defaultChecked doesn't exist anymore

mthuret:
See: https://github.com/algolia/react-instantsearch/issues/52

I also add a story about it.

### DIFF
--- a/packages/react-instantsearch/src/widgets/Toggle.js
+++ b/packages/react-instantsearch/src/widgets/Toggle.js
@@ -8,7 +8,7 @@ import ToggleComponent from '../components/Toggle.js';
  * @propType {string} attributeName - Name of the attribute on which to apply the `value` refinement. Required when `value` is present.
  * @propType {string} label - Label for the toggle.
  * @propType {string} value - Value of the refinement to apply on `attributeName` when checked.
- * @propType {boolean} [defaultChecked=false] - Default state of the widget. Should the toggle be checked by default?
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch/src/widgets/Toggle.js
```diff
@@ -8,7 +8,7 @@ import ToggleComponent from '../components/Toggle.js';
  * @propType {string} attributeName - Name of the attribute on which to apply the `value` refinement. Required when `value` is present.
  * @propType {string} label - Label for the toggle.
  * @propType {string} value - Value of the refinement to apply on `attributeName` when checked.
- * @propType {boolean} [defaultChecked=false] - Default state of the widget. Should the toggle be checked by default?
+ * @propType {boolean} [defaultRefinement=false] - Default state of the widget. Should the toggle be checked by default?
```
Haroenv: When did the actual change of this prop happen, I think `defaultChecked` is clearer, no?

mthuret: A while ago we decided to homogenize the api of all components, meaning that if you know how to use one, you know how to user the others. So every default* where renamed to defaultRefinement :)

Haroenv: All right makes sense

mthuret:
@Haroenv Good catch, thanks :)

Issue #59, opened 2017-04-25
Unused import are no more reported


mthuret:
When an unused import is present, the build is not failing anymore. 

This might be related to the addition of prettier.

Haroenv:
Probably a removal too much in algolia/eslint-confit-algolia indeed

vvo:
Fixed, please upgrade eslint-config-algolia:

```
yarn add eslint-config-algolia # no --dev for react-instantsearch root folder
```

PR #63, opened 2017-04-26
chore: update eslint version

mthuret:
see #59

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectHits.js
+++ b/packages/react-instantsearch/src/connectors/connectHits.js
@@ -1,5 +1,5 @@
 import createConnector from '../core/createConnector';
-import { getIndex, getResults } from '../core/indexUtils';
+import { getResults } from '../core/indexUtils';
 
 /**
  * connectHits connector provides the logic to create connected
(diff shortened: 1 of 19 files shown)


packages/react-instantsearch/src/connectors/connectHits.js
```diff
@@ -17,17 +17,18 @@ export default createConnector({
   displayName: 'AlgoliaHits',
 
   getProvidedProps(props, searchState, searchResults) {
-    const index = getIndex(this.context);
     const results = getResults(searchResults, this.context);
     const hits = results ? results.hits : [];
 
-    return { hits };
+    return {
+      hits,
```
Haroenv: prettier keeps it like this because you once wrote it in full, can you try writing it on one line and see what happens?

packages/react-instantsearch/src/core/Index.js
```diff
@@ -42,7 +42,7 @@ class Index extends Component {
     */
     this.unregisterWidget = widgetsManager.registerWidget({
       getSearchParameters: searchParameters =>
-        this.getSearchParameters.call(this, searchParameters, this.props),
+        this.getSearchParameters(searchParameters, this.props),
```
Haroenv: I had a small doubt if this was 100% the same

PR #70, opened 2017-04-27
fix(core/InstantSearch): update algoliaClient when it change

Mystraht:
**Summary**

ApiKey parameter is not changed in search query when InstantSearch component is updated with new apiKey props.
Algolia client is updated with new values if appId or apiKey is changed (at createInstantSearch.js:39) but helper in InstantSearchManager (created in InstantSearch component) is not updated.

**Result**

I created component that render InstantSearch with apiKey "exampleApiKey1" then i changed component state (with setTimeout for the example) to render InstantSearch again but with a new apiKey props. (exampleApiKey2)

Without my fix, ajax call is always called with exampleApiKey1

http://image.prntscr.com/image/c5a1670088ad4e7b9edfaff72ec9479b.png

With my fix, ajax get is called with exampleApiKey1 then exampleApiKey2 after rendering with the new props (see screens)

http://image.prntscr.com/image/a68d4384ff16488f83813e98b91478cb.png

Thanks

### DIFF
--- a/packages/react-instantsearch/src/core/InstantSearch.js
+++ b/packages/react-instantsearch/src/core/InstantSearch.js
@@ -68,6 +68,10 @@ class InstantSearch extends Component {
     if (this.props.indexName !== nextProps.indexName) {
       this.aisManager.updateIndex(nextProps.indexName);
     }
+    
+    if (this.props.algoliaClient !== nextProps.algoliaClient) {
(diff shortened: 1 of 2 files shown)


Mystraht:
Thank you for this quick response, i just added the unit test

Mystraht:
Ok thanks

mthuret:
@Mystraht Thank you for your contribution :)

PR #76, opened 2017-05-02
docs(Autocomplete): add @mentions to possible use of Autocomplete 

### LINKED ISSUE #71: Building @mentions with createConnector
Hi all,

I'm trying to build an @mention into a textarea, using antd Mention component. But it seems that new props aren't communicated properly to the connected component.

I created a small Fiddle to show the issue : https://jsfiddle.net/8r928c4y/16/

If it is a misuse rather than a bug, feel free to close the ticket of course ;)

I believe new props are not communicated to child component. If the Fiddle, you can see that `getProvidedProps` is called on state change and returns hits as it should.

To retrieve matching hits from the Mention component

Chrome v 49 at least

Thanks!

Haroenv:
**Summary**

closes #71

Same dataset as the [guide](https://www.algolia.com/doc/guides/search/autocomplete-textarea/) that uses the helper etc.

I accidentally branched off the previous branch instead of master, and I'm not really sure how to fix that 📦 

**Result**


mthuret:
I really like this example. Good alternative as an example for the `connectAutoComplete` connector! 🎉 

Will be merged, once deployed on netlify!

vvo:
That's awesome! I wish those were live though :D

Haroenv:
#88 @vvo

PR #81, opened 2017-05-03
chore(build system): make es2015 modules available

### LINKED ISSUE #27: Verify App Build size big due to export strategy and add TreeShaking
The goal here is to create a new application using React InstantSearch and very if the build contains only imported functions.

For example by importing `<InstantSearch>`, `<SearchBox>` and `<connectHits>`.

If not, actions should be taken to fix it.

Haroenv:
After spending a few hours and trying a few options this is the needed solution:

1. add three BABEL_ENV cases
    1. develop: same as before
    2. commonjs: same, but with `babili` added
    3. es: same, but with `babili` and `modules: false`

   I tried a *lot* of things, and this is the only order of things that made it work to me, but at least it worked!

2. edit two node scripts that aren't being babelified anymore (simpler in this case, let's wait until it's supported on node)
3. add /dist/es/ with es module build
    - also mark that build as the `module` in package.json

This will give the same experience for everyone using webpack 1 or browserify, but will have tree shaking for webpack 2 and rollup

fixes #27


rayrutjes:
Hey @Haroenv , please note that by doing this change, we would impact existing webpack2 and rollup users.
We might break existing builds, this is what happened when we tried to add modules to the JS helper.
This might require a major bump.

Haroenv:
4.0.0 is in beta, so that doesn't matter.

mthuret:
@Haroenv @rayrutjes : If this could break builds, what our users will need to do to get it work again? We'll need to update the migration guide.

Haroenv:
I don't think this can break builds like the helper, since react-instantsearch is written in es2015 modules, so here its

es -> es

while in the helper it was

commonjs -> es

packages/react-instantsearch/scripts/build.sh
```diff
@@ -2,15 +2,32 @@
...
 rm -rf dist/* &&
 cp package.json dist/ &&
-cp README.md dist/ &&
+cp README.md dist/
+
+NODE_ENV=production
+
+# first make es module build
+export BABEL_ENV=es
+mkdir -p dist/es &&
```
rayrutjes: According to package.json it should now be 'es' directly I think.

Haroenv: no, the published directory on npm is `dist` alone, see https://unpkg.com/react-instantsearch@beta/

Haroenv:
FYI, tested it on: 

- windtoday (cc @kikobeats)
- examples folder
- yarn website

and it all creates smaller builds, and works without configuration change

iam4x:
Are you open to use UglifyJS2 instead of babili? It will produces smaller builds.

package.json
```diff
@@ -41,10 +41,12 @@
     "babel-eslint": "^7.2.1",
     "babel-jest": "^19.0.0",
     "babel-loader": "^6.3.2",
+    "babel-plugin-external-helpers": "^6.22.0",
```
mthuret: Needs to be removed (not used anymore)

packages/react-instantsearch/scripts/build.sh
```diff
@@ -2,15 +2,32 @@
...
 set -e # exit when error
 
+# copy files
 mkdir -p dist/ &&
 rm -rf dist/* &&
 cp package.json dist/ &&
-cp README.md dist/ &&
+cp README.md dist/
+
+NODE_ENV=production
```
mthuret: remove this

mthuret:
LGTM. Let's try this with a new beta :)

iam4x:
Nice work @Haroenv 👏

harshmaur:
Let me try it out on my app as well.

harshmaur:
It doesnt work with webpack 1? Do I need webpack 2?

Haroenv:
That might be the case @harshmaur, there's no way that I know of we can make sure you have a correctly treeshook bundle with webpack 1.

harshmaur:
@Haroenv, I have used material-ui and It has some export strategy, webpack then does not have to do tree shaking for us. Can you check if that is possible? I still have to migrate to webpack 2 as my config is quite complex and I have to figure out a lot of things before I upgrade.

Haroenv:
I understand that @harshmaur, can you open a new issue specifically for webpack 1? I hope to be able to tackle that too, similar to how lodash/material-ui has done it, but that's a different thing to solve.

harshmaur:
@Haroenv I moved to webpack 2.5.0, My build size is still the same. Do I need to change the way I import?

Haroenv:
That does not really sound right, can I take a look at your repo?

harshmaur:
@Haroenv Can you check your twitter? I have sent you a message.

PR #96, opened 2017-05-16
fix nested attributes for faceting

orthes:
**Summary**

Since the https://github.com/algolia/react-instantsearch/commit/09a4e1dfd98c8989aa292bae4bf792aaf9aa7c48 commit, nested attributes for faceting are no longer working. 

Before, condition if id exists in a state looked like this: [```if (searchState[namespace] && typeof searchState[namespace][id] !== 'undefined') {```](https://github.com/algolia/react-instantsearch/commit/09a4e1dfd98c8989aa292bae4bf792aaf9aa7c48#diff-ff7d5a81ec7fdde10efeed49448d38d5L14). 

Now namespace and id are merged together [```${namespace}.${getId(props)}```](https://github.com/algolia/react-instantsearch/commit/09a4e1dfd98c8989aa292bae4bf792aaf9aa7c48#diff-ff7d5a81ec7fdde10efeed49448d38d5R13) and condition moved to ```indexUtils.getCurrentRefinementValue()``` function and looks like this [```has(searchState, id)```](https://github.com/algolia/react-instantsearch/commit/09a4e1dfd98c8989aa292bae4bf792aaf9aa7c48#diff-9d3b5ecf5b23605ad4be1550c962c5a4R62).

This works in most cases, but if you use nested attributes for faceting, your id looks something like this: ```category.name```. So with the namespace prepended it's ```refinementList.category.name```. And this when run through lodash' ```has()``` function returns `false` even when searchState looks like this:

```
{
  "page": 1,
  "query": "",
  "refinementList": {
    "category.name": ["my category"]
  }
}
```

**Result**

My pull requests just changes ```getCurrentRefinementValue()``` and ```cleanUpValue()``` functions. It splits joined namespace and id using regex and uses them separately. If they are not found, it falls back to the old logic.

I also added some tests that prove that without this fix ```getCurrentRefinementValue()``` and ```cleanUpValue()``` don't return correct results.

### DIFF
--- a/packages/react-instantsearch/src/core/indexUtils.js
+++ b/packages/react-instantsearch/src/core/indexUtils.js
@@ -115,6 +115,14 @@ function refineSingleIndexWithNamespace(
   };
 }
 
+function getNamespaceAndId(input) {
+  const parts = input.match(/^([^.]*)\.(.*)/);
+  const namespace = parts && parts[1];
+  const id = parts && parts[2];
+
+  return { namespace, id };
+}
+
 // eslint-disable-next-line max-params
 export function getCurrentRefinementValue(
   props,
@@ -125,13 +133,25 @@ export function getCurrentRefinementValue(
   refinementsCallback
 ) {
   const index = getIndex(context);
+  const namespace = getNamespaceAndId(id);
   const refinements =
     (hasMultipleIndex(context) && has(searchState, `indices.${index}.${id}`)) ||
+    (!hasMultipleIndex(context) &&
+      namespace.namespace &&
+      namespace.id &&
+      has(searchState[namespace.namespace], namespace.id)) ||
     (!hasMultipleIndex(context) && has(searchState, id));
   if (refinements) {
-    const currentRefinement = hasMultipleIndex(context)
-      ? get(searchState.indices[index], id)
-      : get(searchState, id);
+    let currentRefinement;
+
+    if (hasMultipleIndex(context)) {
+      currentRefinement = get(searchState.indices[index], id);
+    } else if (namespace.namespace && namespace.id) {
+      currentRefinement = get(searchState[namespace.namespace], namespace.id);
+    } else {
+      currentRefinement = get(searchState, id);
+    }
+
     return refinementsCallback(currentRefinement);
   }
   if (props.defaultRefinement) {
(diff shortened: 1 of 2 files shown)


mthuret:
Hi @orthes, thanks for spotting this bug and taking the time to fix it!

I'll review this asap. In the meantime can you fix the build error (small eslint issue)?

orthes:
@mthuret Sure. Sorry about that. Fixed.

packages/react-instantsearch/src/core/indexUtils.js
```diff
@@ -125,13 +133,25 @@ export function getCurrentRefinementValue(
...
+      has(searchState[namespace.namespace], namespace.id)) ||
     (!hasMultipleIndex(context) && has(searchState, id));
   if (refinements) {
-    const currentRefinement = hasMultipleIndex(context)
-      ? get(searchState.indices[index], id)
-      : get(searchState, id);
+    let currentRefinement;
+
+    if (hasMultipleIndex(context)) {
+      currentRefinement = get(searchState.indices[index], id);
```
mthuret: This will still not work for multi indices application. We need to replace by something like: has(
        searchState.indices[`${index}`][namespace.namespace],
        `${namespace.id}`
      ))

packages/react-instantsearch/src/core/indexUtils.js
```diff
@@ -125,13 +133,25 @@ export function getCurrentRefinementValue(
...
 ) {
   const index = getIndex(context);
+  const namespace = getNamespaceAndId(id);
   const refinements =
     (hasMultipleIndex(context) && has(searchState, `indices.${index}.${id}`)) ||
+    (!hasMultipleIndex(context) &&
+      namespace.namespace &&
+      namespace.id &&
+      has(searchState[namespace.namespace], namespace.id)) ||
     (!hasMultipleIndex(context) && has(searchState, id));
```
mthuret: This is not needed anymore right? We can handle only two case (one for multi indices, one regular) no?

packages/react-instantsearch/src/core/indexUtils.js
```diff
@@ -125,13 +133,25 @@ export function getCurrentRefinementValue(
   refinementsCallback
 ) {
   const index = getIndex(context);
+  const namespace = getNamespaceAndId(id);
```
mthuret: I believe it would be better to find another name :)

mthuret:
@orthes: as supporting multi indices for this fix will require more work, I can take over to complete this PR if you want :)

orthes:
@mthuret Thank you very much. That would be great. Since I'm not using multi indices in my app, I completely forgot about that. Feel free to take this and refactor as you see fit.

packages/react-instantsearch/src/core/indexUtils.js
orthes: Haha. Absolutely :)

packages/react-instantsearch/src/core/indexUtils.js
orthes: You know the library much better than I do. I left it there just in case I'm missing some use case. I wanted to be 100% sure it's backwards compatible.

mthuret:
@orthes I added the fix for the multi indices part. You can review it before I merge this and publish a new version if you want :)

orthes:
@mthuret Thanks. Looks great!

PR #120, opened 2017-06-04
docs(typo): Fix typo in Menu Widget documentation

pixelastic:
Tiny typo in the `transformItems` description:

> Takes an item__s__ as parameter


pixelastic:
Woops sorry, actually I'm confused by the documentation. Does it take "one item and return it back", or does it take "an items array and returns it back". I assumed the first, but now that i'm testing the code I'm not sure.

mthuret:
@pixelastic `items` is correct, it's all the items displayed by the menu.

pixelastic:
Maybe `items list` or `items array` would be more easy to understand. Just saying `items` seems like internal lingo, but is not very friendly for the new comers to the library (at least, that how I felt it).

PR #122, opened 2017-06-06
docs(Highlight): improve guide and examples

mthuret:


docgen/src/guide/Highlighting_results.md
```diff
@@ -73,13 +73,36 @@ Here is an example of a custom Highlight widget. It can be used the same
...
+      if (part.isHighlighted) return <mark>{part.value}</mark>;
+      return part.value;
+    });
+    return <div>{highlightedHits}</div>;
+  }
+);
+
+const Hit = ({hit}) =>
+<p>
+  <CustomHighlight attributeName="description" hit={hit} highlightProperty="_highlightResult" />
```
mthuret: This connector is also use for the Snippet component so we can't hardcode it anymore. But I can precise those properties.

docgen/src/guide/Highlighting_results.md
```diff
@@ -73,13 +73,36 @@ Here is an example of a custom Highlight widget. It can be used the same
...
-    if(part.isHighlighted) return <mark>{part.value}</mark>;
-    return part.value;
-  });
-});
+import React from 'react';
+import { connectHighlight } from 'react-instantsearch/connectors';
+import { InstantSearch, Hits } from 'react-instantsearch/dom';
+
+const CustomHighlight = connectHighlight(
+  ({ highlight, attributeName, hit, highlightProperty }) => {
```
Haroenv: What about putting here `highlightProperty = '_highlightResult’` Then it would be obvious to me.

docgen/src/guide/Highlighting_results.md
pixelastic: As a user reading the documentation, if I see a component named `CustomHighlight`, I expect it to do Highlight, and Highlight only, not Snippetting. 

My suggestion would then be to have the following example:
```
const CustomHighlight = connectHighlight(
  ({ highlight, attributeName, hit }) => {
    const parsedHit = highlight({ attributeName, hit, highlightProperty: '_highlightResult' });
    const highlightedHits = parsedHit.map(part => {
      if (part.isHighlighted) return <mark>{part.value}</mark>;
      return part.value;
    });
    return <div>{highlightedHits}</div>;
  }
);

const Hit = ({hit}) =>
<p>
  <CustomHighlight attributeName="description" hit={hit} />
</p>;
```

That way, the end-result component, `CustomHighlight` only deals with Highlight. For sure, it's less modular than being able to pass the `highlightProperty`, but I think it's clearer.

Issue #123, opened 2017-06-07
When changing the hitsPerPage configuration on the fly, connectInfiniteHits doesn't react well


mthuret:
See: https://discourse.algolia.com/t/infinite-hits-with-skip-and-limit/1538/3

Basically when you change the number of hitsPerPage on the fly, then you end up with duplicates element.

PR #124, opened 2017-06-07
fix(InfiniteHits): changing hitsPerPage were causing duplicates

mthuret:
see: https://jsfiddle.net/spondbob/sm3cad75/ and https://github.com/algolia/react-instantsearch/issues/123

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectInfiniteHits.js
+++ b/packages/react-instantsearch/src/connectors/connectInfiniteHits.js
@@ -51,16 +51,14 @@ export default createConnector({
       };
     }
 
-    const { hits, page, nbPages, hitsPerPage } = results;
+    const { hits, page, nbPages } = results;
 
     if (page === 0) {
       this._allResults = hits;
     } else {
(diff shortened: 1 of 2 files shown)


bobylito:
I'm not sure about the behaviour here. Now and before the content will be broken:
 - before, changing the hitsPerPage would trigger a rerender (because the previous page was computed) and if hitsPerPage increased then no duplicate but some missing elements, if less some duplicates
 - now, changing the hitsPerPage does not trigger a rerender (because the page is the same), when the user ask for a new page then we have the same issue

In which context, do we want to use a dynamic hitsPerPage? Can't we just say "infiniteHits" is not to be used with HitsPerPageSelector?

bobylito:
I might be wrong on this one, but as a precaution we'll meet with @mthuret on monday :)

mthuret:
This commit only improve readability and doesn't fix the underlying problem. It's currently not possible to change the hitsPerPage value while using the `connectInfiniteHits`.

PR #125, opened 2017-06-12
fix(SFFV): search status we're not inside search state

mthuret:
the search status were not available inside the search state. It means that there's no way to actually display something if this search is currently being performed.


vvo:
Does it needs any docs?

mthuret:
Will add something on the Conditional Display... but ultimately we need to offer a better API for this.

PR #136, opened 2017-06-21
fix(MultiIndex): ensure getResults return only hits matching index in the context

mthuret:


packages/react-instantsearch/src/connectors/connectInfiniteHits.test.js
```diff
@@ -41,7 +42,7 @@ describe('connectInfiniteHits', () => {
       const hits2 = [{}, {}, {}, {}, {}, {}];
       const hits3 = [{}, {}, {}, {}, {}, {}, {}, {}];
       const res1 = getProvidedProps(null, null, {
-        results: { hits, page: 0, hitsPerPage: 6, nbPages: 10 },
+        results: { hits, page: 0, hitsPerPage: 6, nbPages: 10, index: 'index' },
```
vvo: Should we use "indexName" like other parts of the lib? I would say any moment we speak about an index name we should use the "indexName" token as we do in every JS lib.

vvo:
Do you have a small example usage of when this happened?

On a side note, this is not only a fix but also a new feature where the index name is always passed down to the results (multi index or not) right?

packages/react-instantsearch/src/connectors/connectInfiniteHits.test.js
mthuret: `index` is the name used on the `SearchResults` object from the helper.

mthuret:
@vvo Not sure to get what you mean but this bug is happening when switching from one indices to multiple indices. For example we saw it on an example showcasing "tabs". You have one tab that is all, and then one tab per index. 

There is actually a better way to handle this (by namespacing every index, even when only one), but it would imply a breaking change. Initially we were using it, you can see this PR when I revert the breaking change. https://github.com/algolia/react-instantsearch/commit/44f7de08b4b4f971a4feb16efcb84e86af9255cb

packages/react-instantsearch/src/connectors/connectInfiniteHits.test.js
vvo: Does it bubbles up to the React InstantSearch user API?

mthuret: If you're using the `createConnector` API yes, because you access `searchResults` otherwise it shouldn't.

Issue #135, opened 2017-06-20
[theme] SearchBox needs -webkit-appearance: none


Haroenv:
**Do you want to request a *feature* or report a *bug*?**

bug

**Bug: What is the current behavior?**

Safari shows the SearchBox as a default one

<img width="456" alt="screen shot 2017-06-20 at 14 22 22" src="https://user-images.githubusercontent.com/6270048/27332851-ef169d00-55c3-11e7-956c-f414e0d330a9.png">

**Bug: What is the expected behavior?**

our style overriding the webkit appearance

**Bug: What browsers are impacted? Which versions?**

Safari

**What is the version you are using? Always use the latest one before opening a bug issue.**

I was looking at the storybook, so the latest master as of now?

iam4x:
I experienced this bug as well, after a fast investigation its because nothing is prefixed with vendors. Does autoprefixer run correctly with storybook?

Haroenv:
It’s weird that this didn’t happen before, did something in the toolchain change since May-ish?

mthuret:
I switched to storybook 3 and so it now uses webpack 2. This conf is probably missing.

Haroenv:
I can still see the issue live in the storybook 

https://community.algolia.com/react-instantsearch/storybook/?selectedKind=SearchBox&selectedStory=default&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs

`appearance: none` | `-webkit-appearance: none`
---|---
<img width="324" alt="screen shot 2017-06-23 at 19 17 29" src="https://user-images.githubusercontent.com/6270048/27493065-b1be978c-5848-11e7-9ba5-97269b159b34.png">|<img width="350" alt="screen shot 2017-06-23 at 19 17 17" src="https://user-images.githubusercontent.com/6270048/27493067-b2b3d6c0-5848-11e7-869f-7f6a854067b9.png">

mthuret:
The docs wasn't push into prod yet.

PR #138, opened 2017-06-22
fix(storybook): process CSS through autoprefixer

iam4x:
CSS theme is not vendor prefixed into storybook, which leads to the issue #135.

This PR use `postcss-loader` to process css into `autoprefixer` to resolve it 👍


iam4x:
About Argos failing:

* Some search results are not yet loaded so it makes a difference
* Styles changed because they gets applied because of vendors prefix

PR #139, opened 2017-06-22
fix(MultiIndex): handle if namespace isn't in search state

mthuret:
Use case: you handle a tab navigation without any router and you're switching between a multiple indices display and several only one indices display. 
At some point, before the rendering of only one indices, the namespace with the index name is not here yet, so without this check it crashed.

### DIFF
--- a/packages/react-instantsearch/src/core/indexUtils.js
+++ b/packages/react-instantsearch/src/core/indexUtils.js
@@ -138,6 +138,7 @@ export function getCurrentRefinementValue(
     (hasMultipleIndex(context) &&
       searchState.indices &&
       namespace &&
+      searchState.indices[`${index}`] &&
       has(searchState.indices[`${index}`][namespace], `${attributeName}`)) ||
     (hasMultipleIndex(context) &&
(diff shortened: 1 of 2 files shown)


mthuret:
@oliviertassinari I'm not sure to understand the argos error here. If you check our storybook here: https://deploy-preview-139--react-instantsearch.netlify.com/react-instantsearch/storybook/ you can see that the stories are fine.

oliviertassinari:
@mthuret Right, we have a cropping issue with Happo during the screenshot generation.

mthuret:
@oliviertassinari also it seems that sometimes it takes a screenshot before the return of results.

vvo:
When would this happen as a user?

mthuret:
@vvo what do you mean?

oliviertassinari:
@vvo I don't think that it's a react-instant search issue, sounds more like a Happo one, not measuring the component dimensions at the right time. For instance, it's could be linked to asynchronous image loading.

vvo:
Thanks for the explanation.

Issue #137, opened 2017-06-22
infinite loop when rendering a conditional loading message with children


blech75:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**Bug: What is the current behavior?**

(See below for full description.)

Attempting to use a connected "loading message" component (one that renders static content when loading and its children when finished) causes an infinite loop in certain circumstances.

* when `children` contains the `<Stats />` component, `searchResults.searching` is set as expected.
* when `children` contains the `<Hits />` and `<Pagination />` components (possibly others), `searchResults.searching` flips back and forth ad infinitum.

When the "loading message" component is configured to render static content upon completion, the infinite loop does not occur.

**Bug: What is the expected behavior?**

`searchResults.searching` should not flip-flop between `true` and `false` in an infinite loop when rendering `children`.

**Bug: What browsers are impacted? Which versions?**

Observed in Chrome 59.0.3071.109 & Safari 10.1.1.

**What is the version you are using? Always use the latest one before opening a bug issue.**

Observed in react-instantsearch 4.0.3 and 4.0.4.
 
----

### Background

(copied/adapted from [discussion on community forum](https://discourse.algolia.com/t/searching-multiple-indices-with-different-fields-loading-indicator/1647/8).)

Using the example connector provided [in your docs](https://community.algolia.com/react-instantsearch/guide/Conditional_display.html#displaying-content-when-loading):
```
const connectLoading = createConnector({
  displayName: 'ConditionalError',
  getProvidedProps(props, searchState, searchResults) {
    return {
      loading: searchResults.searching
    };
  }
});
```

I then create a component that renders a message when loading, and its children when complete:
```
const LoadingMessage = connectLoading(({ loading, children }) =>
  <div>
    {loading ? <p>Loading...</p> : children }
  </div>
);
```

Now we make use of it in a minimal UI where we just display the stats:
```
<InstantSearch ...>
  <SearchBox />

  <div className="SearchResults">
    <LoadingMessage>
      <p>FINISHED!</p>
      <Stats />
    </LoadingMessage>
  </div>

</InstantSearch>
```

[...]

PR #143, opened 2017-06-26
fix(store): delay call to listener to prevent infinite loops

mthuret:
See https://github.com/algolia/react-instantsearch/issues/137

### DIFF
--- a/packages/react-instantsearch/src/core/createStore.js
+++ b/packages/react-instantsearch/src/core/createStore.js
@@ -2,7 +2,7 @@ export default function createStore(initialState) {
   let state = initialState;
   const listeners = [];
   function dispatch() {
-    listeners.forEach(listener => listener());
+    listeners.forEach(listener => setTimeout(listener, 0));
   }
   return {
     getState() {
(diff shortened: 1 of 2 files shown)


vvo:
Quick note for us to discuss: I also thought that we could delay the update of searching flag to handle the case that is giving us troubles here: when response comes from JavaScript client cache, it's so fast it's not useful to set it to true then false.

mthuret:
@vvo Ultimately this should be hidden with an intermediary API. It will be part of our new "Conditional Display" API.

mthuret:
@oliviertassinari do we have a solution for the happo issue? Otherwise I'll just deactivate argos until we find a solution.

oliviertassinari:
@mthuret Not yet, we can pause argos for now.

oliviertassinari:
I have taken care of the color diff issue https://github.com/argos-ci/argos/issues/198 by rewriting [image-diff](https://github.com/argos-ci/image-difference). Now, I can take care of that windowing issue :).

PR #147, opened 2017-07-03
docs(RN): fix example

mthuret:
Following feedback on twitter: 

- dismiss keyboard when navigating on Android
- Hide back button on Android
- Fix `Nan` display on the price page

To try: 

image

### DIFF
--- a/packages/react-instantsearch/examples/react-native/App.js
+++ b/packages/react-instantsearch/examples/react-native/App.js
@@ -25,7 +25,13 @@ export default class App extends Component {
             }),
           }}
         >
-          <Scene key="Home" component={Home} title="AEKI" initial={true} />
+          <Scene
+            key="Home"
+            component={Home}
+            title="AEKI"
+            initial={true}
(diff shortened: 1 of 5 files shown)


Haroenv:
I don't know if this is like a special deploy preview build, but I find that the transitions are going a bit choppy. Would this change have anything to do with it? The version you showed on Twitter was significantly smoother (probably 60fps)

mthuret:
@Haroenv yep, this one is in development mode.

Haroenv:
While now you changed something in the keyboard, can you make it a search keyboard instead of a regular one?

vvo:
> While now you changed something in the keyboard, can you make it a search keyboard instead of a regular one?

What would this do UI wise?

Haroenv:
`search` instead of `return`, like I did on web in algolia/instantsearch.js#1999

mthuret:
@Haroenv on the Android phone I have, it's written done, which I found actually less mesleading compare to search in a search as you type experience. (also if you select something, the keyboard is dismiss).

PR #149, opened 2017-07-06
fix(results): revert commit that ensure hits are returned only if right indices

mthuret:
I need to revert this commit because it breaks our SortBy feature. 

As for now we don't have a simple way to handle a switch between the multi and mono indices mode. What it means for the user is that a connector can get hits that were meant for another index before getting the right one. 

There's two ways to handle this issue on user side: 
-> Always be in multi indices mode => It means that when you are on a tab that contains only one index you can still add another `<Index/>` but empty. 
-> Check the shape of the hits and decide to proceed with it or not. (i.e => hit.brand ? ... : null). 

Ultimately we will need to find a way to fix this properly. I can see several options: 
-> SearchResults are always namespaced by the indexName. This is a breaking change because the `createConnector` api exposed directly the `SearchResults`. 
-> Clean the hits when performing a new search. (But in a smart way to avoid a blinking effect).

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectHierarchicalMenu.test.js
+++ b/packages/react-instantsearch/src/connectors/connectHierarchicalMenu.test.js
@@ -22,7 +22,6 @@ describe('connectHierarchicalMenu', () => {
         getFacetValues: jest.fn(),
         getFacetByName: () => true,
         hits: [],
-        index: 'index',
       };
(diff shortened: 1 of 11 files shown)


vvo:
Maybe add an explanation on what is not working or why it was not the right way to do it (if you have insights)

PR #152, opened 2017-07-09
fix(argos): address flakyness

oliviertassinari:
**Summary**

Following 1ff96ad6aaa7565d8f49cd209c667ef98dc4d3ca

**Result**

Should have flake less build now.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -8,7 +8,6 @@ if [ "$CI" = "true" ]
   else jest
 fi
 yarn lint
-# issue with happo. Uncomment when fixed. 
-# yarn test:regressions
-# yarn argos
+yarn test:regressions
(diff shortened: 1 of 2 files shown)


mthuret:
@oliviertassinari small question, one test is still in orange (but the argos CI task is green). Is it normal?

oliviertassinari:
@mthuret The three images have changed since we have token the baselines. I have tried [increasing the threshold of the diff](https://github.com/argos-ci/argos/commit/ea2c2d5cc3e94d67623c19c6ef93e02f59561391#diff-77d7e9e173ff37d4de888bc6b305e3d9), there is two much difference in the versions of the images.

mthuret:
@oliviertassinari, okay let's give it a try.

PR #153, opened 2017-07-10
docs(SSR): fix build of SSR example

mthuret:
npm run build was broken.

### DIFF
--- a/packages/react-instantsearch/examples/next-app/package.json
+++ b/packages/react-instantsearch/examples/next-app/package.json
@@ -2,12 +2,12 @@
   "name": "create-next-example-app",
   "scripts": {
     "dev": "next",
-    "build": "next build",
+    "build": "NODE_ENV=development next build",
     "start": "next start"
   },
   "dependencies": {
     "css-loader": "^0.28.1",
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch/examples/next-app/package.json
```diff
@@ -2,12 +2,12 @@
   "name": "create-next-example-app",
   "scripts": {
     "dev": "next",
-    "build": "next build",
+    "build": "NODE_ENV=development next build",
```
vvo: Shouldn't the build be NODE_ENV=production? Usually when hosting it provides better performance

mthuret: Well this is not working since some modifications with the .babelrc file to handle code splitting. Maybe @Haroenv could have a look.

vvo: Hm indeed that babelrc is only setupping things for development. We should setup things as a whole and then provide some differences for production

vvo: I am doing some chore work on the build and upgrades so I will update this one I think.

PR #158, opened 2017-07-12
Initial state will be parsed properly now

harshmaur:
Hello, 

Thank You for quick SSR support for the library! I was trying the example and found that initial state was not parsing properly. So the change fixes it. Rest everything works great!

### DIFF
--- a/packages/react-instantsearch/examples/next-app/pages/index.js
+++ b/packages/react-instantsearch/examples/next-app/pages/index.js
@@ -26,7 +26,7 @@ export default class extends React.Component {
      to initialize the searchState. 
   */
   static async getInitialProps(params) {
-    const searchState = qs.parse(params.asPath.substring(2));
+    const searchState = qs.parse(asPath.substring(asPath.indexOf('?') + 1))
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch/examples/next-app/pages/index.js
```diff
@@ -26,7 +26,7 @@ export default class extends React.Component {
      to initialize the searchState. 
   */
   static async getInitialProps(params) {
-    const searchState = qs.parse(params.asPath.substring(2));
+    const searchState = qs.parse(asPath.substring(asPath.indexOf('?') + 1))
```
mthuret: Not sure to understand where does asPath came from. 

Also, what was the use case were the substring(2) was failing? 

Thanks for you PR @harshmaur!

harshmaur: Oh Sorry! should be `params.asPath`

Also, I am doing custom routing so my asPath is different from href. 

asPath can be `/deals/store/amazon?query=1` 
`.substring` would fail in this case.

PR #166, opened 2017-07-17
docs(pictures): use cloudinary to fix Mixed Content issue.

mthuret:
We have those warnings due to Ikea picture not being under https. 

image

### DIFF
--- a/docgen/src/examples/e-commerce-infinite/App.js
+++ b/docgen/src/examples/e-commerce-infinite/App.js
@@ -174,7 +174,11 @@ const Hit = ({ item }) => {
   return (
     <article className="hit">
       <div className="product-picture-wrapper">
-        <div className="product-picture"><img src={`${item.image}`} /></div>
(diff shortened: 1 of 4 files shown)


docgen/src/examples/e-commerce-infinite/App.js
```diff
@@ -174,7 +174,11 @@ const Hit = ({ item }) => {
   return (
     <article className="hit">
       <div className="product-picture-wrapper">
-        <div className="product-picture"><img src={`${item.image}`} /></div>
+        <div className="product-picture">
+          <img
+            src={`https://res.cloudinary.com/hilnmyskv/image/fetch/h_300,q_100,f_auto/${item.image}`}
```
Haroenv: what are our wloudinary limits, probably high enough, right?

mthuret: I assume so, it's already use for the logo for ex.

PR #170, opened 2017-07-18
docs(RN): Added missing Text import in React Native docs example

ArturKlajnerok:
**Summary**

Example in https://community.algolia.com/react-instantsearch/Getting_started_React_native.html is missing a `Text` import.

**Result**

Simple docs fix.


mthuret:
Thank you very much @ArturKlajnerok!

Issue #177, opened 2017-07-20
href links in NextJS not coming


harshmaur:
Hello, 

I wanted to know why href links as just `#` in the next js example in pagination widget, while it works in the eCommerce demo. 

In order for google to crawl the next pages, its necessary to have the href links inside the <a> tags

mthuret:
Hi @harshmaur,

Can you confirm that this occurs only when entering the page without any refinements made? 

I suspect a wrong usage of the Router. I'll check it out, thanks!

harshmaur:
I am not sure If I understand you correctly. It happens to me on every case. with or without refinements.

mthuret:
Fix with 4.1.0.beta3

PR #182, opened 2017-07-24
docs(next-app): fix createURL in url sync

mthuret:
See: https://github.com/algolia/react-instantsearch/issues/177

next-app-bug

### DIFF
--- a/packages/react-instantsearch/examples/next-app/components/app.js
+++ b/packages/react-instantsearch/examples/next-app/components/app.js
@@ -10,16 +10,22 @@ import {
 } from 'react-instantsearch/dom';
 import { InstantSearch } from './instantsearch';
 
-const HitComponent = ({ hit }) =>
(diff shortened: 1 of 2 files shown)


vvo:
We need to eslint and prettify the whole code soon because it seems there have been some unwanted updates (renovate pining will solve this) and now your editor is reformating while you are fixing a bug (hard to find what you fixed exactly)

packages/react-instantsearch/examples/next-app/components/app.js
```diff
@@ -50,6 +58,7 @@ export default class extends React.Component {
         resultsState={this.props.resultsState}
         onSearchStateChange={this.props.onSearchStateChange}
         searchState={this.props.searchState}
+        createURL={this.props.createURL}
```
vvo: This is the fix

Issue #168, opened 2017-07-17
SSR : bug on highlighting


vvo:
**Do you want to request a *feature* or report a *bug*?**

Bug

**Bug: What is the current behavior?**

The next.js example has a weird behavior on reloading, it shows `co<em>ca</em>` as text instead of the actual co**ca**:

![2017-07-17 15 54 09](https://user-images.githubusercontent.com/123822/28271213-3b3c194a-6b08-11e7-8e8c-61b93131de31.gif)

**Bug: What is the expected behavior?**

It should not show html tags as raw.

**Bug: What browsers are impacted? Which versions?**

all

**What is the version you are using? Always use the latest one before opening a bug issue.**

Latest

mthuret:
The root cause of this bug became for the timestamp present in our highlighted tags. 

A first request is made to recover the first results, they have a tag like: ais-higlight-timestamp1. 

When the app is finally rendered with those injected results, then the highlightedTags changed and are now: ais-highlight-timestamp2. 

I can see several ways to fix this:

1. Removing the timestamp. 
2. Somehow pass the timestamp used for the first request for the first rendering.

mthuret:
cc @vvo @marielaures @Haroenv

Haroenv:
Yes, I think that it's reasonable to expect that people won't put `__ais-highlight__` in their data more than `__ais-highlight-123123__`

in [highlightTags.js](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/src/core/highlightTags.js) I would suggest the following diff: 

```diff
- const timestamp = Date.now().toString();

export default {
-  highlightPreTag: `<ais-highlight-${timestamp}>`,
-  highlightPostTag: `</ais-highlight-${timestamp}>`,
+  highlightPreTag: `__ais-highlight__`,
+  highlightPreTag: `__/ais-highlight__`,
};
```

cc @rayrutjes who discussed this yesterday in Vue InstantSearch, where the conclusion was that we'll probably have to drop XSS safety completely there, in React InstantSearch I think we can still keep this, but should focus on making arrays work too.

vvo:
the timestamp is used to limit the attack surface: it's harder to inject a timestamp at exactly the right moment to be displayed on a victim running APP.

Server-side wide this does not make much sense because we could be in situations where the generated timestamp is exactly the same for all requests to the html content (example: a nodejs service sending html of react-instantsearch).

[...]

PR #183, opened 2017-07-24
fix(highlight): wrong parsing between client and server

mthuret:
See https://github.com/algolia/react-instantsearch/issues/168

The Server-side html is rendered perfectly but then when rendered on the client using the `resultsState` props, results are containing the server old timestamp and `<Highlight>` the new client one. The `<Highlight/>` component then doesn't see any highlighted values and that's why we can see the little glitch. 

For now, the solution is to remove the `timestamp` when dealing with highlighting tags.

### DIFF
--- a/packages/react-instantsearch/src/core/createInstantSearchServer.js
+++ b/packages/react-instantsearch/src/core/createInstantSearchServer.js
@@ -9,6 +9,7 @@ import ReactDom from 'react-dom/server';
 import { getIndex, hasMultipleIndex } from './indexUtils';
 import { isEmpty } from 'lodash';
 import cis from './createInstantSearch';
+import highlightTags from './highlightTags.js';
 
 const createInstantSearch = function(algoliasearch) {
   const InstantSearch = cis(algoliasearch, {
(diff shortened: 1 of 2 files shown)


Haroenv:
I think this is a breaking change though, you could do `s/<ais-highlight-\d{6}/>/<em>/g`, and this would break it, no?

mthuret:
@Haroenv => for people using regex and `setDangerousHtml` (for parsing arrays for instance), yes. Will document that in the migration guide when releasing. Hopefully this shouldn't be done by too many people.

Haroenv:
In Yarn I imported the `highlightTags` module, so it should be safe to upgrade, I think this methods is the most used, so LGTM, other info was internal, so I think it's safe

mthuret:
@Haroenv Is it because you need arrays support? We should be careful if we move the file then.

Haroenv:
yes, for arrays I `import highlightTags from 'react-instantsearch/src/core/highlightTags';`, which will still work after this change

vvo:
Also if people were importing and using stuff that is not documented this is not a breaking change I believe. This was not public API.

PR #189, opened 2017-07-25
fix(nextjs/ssr): parse `params.asPath`

iam4x:
There is an edge case where `params.asPath === '/'` and gets converted to `{'/': null}` as search state.

Checking that we have an `?` before fix this issue.


packages/react-instantsearch/examples/next-app/pages/index.js
```diff
@@ -28,9 +28,9 @@ export default class extends React.Component {
      to initialize the searchState. 
   */
   static async getInitialProps(params) {
-    const searchState = qs.parse(
-      params.asPath.substring(params.asPath.indexOf('?') + 1)
-    );
+    const searchState = params.asPath.includes('?')
```
Haroenv: .includes is not supported everywhere we support. I think you should do `indexOf('?') > -1` 😄

iam4x: In this context with nextjs and babel is it safe to use `.includes` 👍

vvo: available in evergreen browsers, always check https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Browser_compatibility. So it's good

Haroenv: Ah, I forgot that this was node-only

Issue #164, opened 2017-07-17
Scrolling Top for Pagination is also Scrolling Top when a refinement is triggered


mthuret:
What's happening: 

When using `<ScrollTo>` to listen to a pagination change (default behaviour with `scrollOn` being "page"), it also scroll to if a refinement is made on any other widget if pagination was never used or if pagination is >=2. 

What should happen: 

When clicking on another widget than pagination no scroll should occur. 

Causes: 

Every time a refinement is made, the pagination is reset to 1. Therefore, `<ScrollTo/>` is detecting a change. 

Proposal: 

We could compute a boolean that compares if the state has changed (removing page and configure keys). If so, then it means that a refinement was made other than the pagination one, and therefore, no `<ScrollTo/>` should happen. 

What do you think @vvo, @marielaures, @Haroenv  ? 

See: https://discourse.algolia.com/t/want-to-be-able-to-only-scroll-to-top-for-pagination-refinement/1963/2

vvo:
Yes, I think we had the discussion some time ago, it's disturbing that it does it on every widgets. nice find

vvo:
Maybe we went too far at the begining, we should just have coded the scrollTo inside the pagination element with a DOM selector (dirty but that's the only usecase today :D)

mthuret:
Maybe, I don't know if people uses it in another way

Haroenv:
solved in #202 (for future reference)

timkelty:
@Haroenv I'm still experiencing this issue and don't see how #202 has solved it…can you elaborate?

Haroenv:
I'm not completely sure about the context of this issue, could you make a new issue with a sandbox to explain what your problem is @timkelty ?

timkelty:
>>> could you make a new issue with a sandbox to explain what your problem is

/cc @kylewayneluck @kswedberg

timkelty:
@Haroenv Need anything more from me?

VivekNayyar:
This problem is still happening. Even after updating to `"react-instantsearch": "^5.3.2"`.
Any one with a solution for this?

vivek12345:
This problem is still happening. Even after updating to "react-instantsearch": "^5.3.2".
Any one with a solution for this?

samouss:
[...]

PR #202, opened 2017-07-25
fix(ScrollTo): with pagination scroll to only if pagination widget refined

mthuret:
See: https://github.com/algolia/react-instantsearch/issues/164


packages/react-instantsearch/src/connectors/connectScrollTo.js
```diff
@@ -32,6 +32,27 @@ export default createConnector({
...
+      the results. 
+      See: https://github.com/algolia/react-instantsearch/issues/164
+    */
+    const cleanedSearchState = cleanUpValue(
+      cleanUpValue(searchState, this.context, 'configure'),
+      this.context,
+      id
+    );
+
+    const hasNotChanged = isEqual(this._prevSearchState, cleanedSearchState);
```
vvo: Can you confirm that:
- this deepEqual comparison is done at every refinement (not at every scroll event)
- what we are doing deep equal on is react-instantsearch state (not helper state)

mthuret: So it's called twice each time your perform a refinement. (before and after the query is resolved). It's not called when scrolling. 

The isEqual is done on the `searchState` of RIS, not the helper one. 

However I believe we can use the `shallowEqual` here, wdyt?

vvo: > However I believe we can use the shallowEqual here, wdyt?

I think yes since the way widgets are refining is to change references

mthuret:
Indeed the strategy doesn't work with multi-index. 
You would need to also remove the empty multi index namespaces (indices: indexName...) and the configure that can also figure at the top. Otherwise the first time you'll click on the pagination, it will see that a change occurred and don't scroll.

mthuret:
=> Algo updated to work with multi index.

packages/react-instantsearch/src/connectors/connectScrollTo.js
```diff
@@ -32,6 +38,36 @@ export default createConnector({
...
+
+    /* Get the subpart of the state that interest us*/
+    if (hasMultipleIndex(this.context)) {
+      const index = getIndex(this.context);
+      searchState = searchState.indices ? searchState.indices[index] : {};
+    }
+
+    /*
+      We need to keep track of the search state to know if there's a change in the app
+      that was not triggered by the props.scrollOn (id) or the Configure widget.
```
marielaures: Maybe "if there is a change in the app that has been triggered by another element than "props.scrollOn (id) or the Configure widget" to insist on the fact that only those 2 are able to affect scrollTo

packages/react-instantsearch/src/connectors/connectScrollTo.js
```diff
@@ -32,6 +38,36 @@ export default createConnector({
...
+      searchState = searchState.indices ? searchState.indices[index] : {};
+    }
+
+    /*
+      We need to keep track of the search state to know if there's a change in the app
+      that was not triggered by the props.scrollOn (id) or the Configure widget. 
+      This is useful when using ScrollTo in combination of Pagination. As pagination
+      can be change by every widget, we want to scroll only if it cames from the pagination
+      widget itself. We also remove the configure key to do this comparaison because for 
+      now configure values are not present in the search state before a first refinements and will false
```
marielaures: before a first refinement has been made

packages/react-instantsearch/src/connectors/connectScrollTo.js
```diff
@@ -32,6 +38,36 @@ export default createConnector({
...
+      const index = getIndex(this.context);
+      searchState = searchState.indices ? searchState.indices[index] : {};
+    }
+
+    /*
+      We need to keep track of the search state to know if there's a change in the app
+      that was not triggered by the props.scrollOn (id) or the Configure widget. 
+      This is useful when using ScrollTo in combination of Pagination. As pagination
+      can be change by every widget, we want to scroll only if it cames from the pagination
+      widget itself. We also remove the configure key to do this comparaison because for
```
marielaures: Could we specify that it is part of the searchState or is it obvious? :)

packages/react-instantsearch/src/connectors/connectScrollTo.js
```diff
@@ -1,15 +1,21 @@
...
+import { shallowEqual } from '../core/utils';
+import { omit } from 'lodash';
 /**
  * connectScrollTo connector provides the logic to build a widget that will
  * let the page scroll to a certain point.
  * @name connectScrollTo
  * @kind connector
  * @propType {string} [scrollOn="page"] - Widget searchState key on which to listen for changes, default to the pagination widget.
  * @providedPropType {any} value - the current refinement applied to the widget listened by scrollTo
+ * @providedPropType {boolean} hasNotChanged - indicate if the refinement came from the scrollOn observed widget
```
marielaures: indicates whether the refinement came from the scrollOn argument (for instance page if we used scrollOn(page))

Issue #160, opened 2017-07-17
Move to latest eslint-config-algolia


vvo:
The [eslint-config-algolia](https://github.com/algolia/eslint-config-algolia) has been updated in many ways recently.

We should update to the latest version and also remove the commit hook. I have experienced issues with it and so did other contributors. It's slow most of the time and redundant with the CI test script, editor auto formatting that already take cares of this formatting/linting checks.

vvo:
taking this one

vvo:
Removing myself from this issue, might not have time

vvo:
@mthuret if you are doing this, you can also remove husky and lint-staged, no more the need of formatting staged files if we have renovate + lint in CI + autoformatting in editors. Just extra work

PR #205, opened 2017-07-25
chore(eslint): upgrade eslint-config-algolia

mthuret:
See https://github.com/algolia/react-instantsearch/issues/160


mthuret:
@Haroenv => other changes? 

For the record, locally, createInstantSearch.test.js was failing when testing for the right version applied. It's due to jest cache. To clean it, do a jest --showConfig, get the `cacheDirectory` and empty it.

Haroenv:
I see that for example quotes were changed in storybooks.js

PR #206, opened 2017-07-25
docs(react-native): fix broken link

rayrutjes:
**Summary**

**Result**


docgen/src/Getting_started_React_native.md
```diff
@@ -102,7 +102,7 @@ We don't provide any widgets for React native yet, but you’ll still be able to
 
 Connectors are higher order components. They encapsulate the logic for a specific kind of widget and they provide a way to interact with the InstantSearch context.
 
-**Checkout our dedicated [connectors guide](guide/connectors.html). to learn more about them**
+**Checkout our dedicated [connectors guide](guide/Connectors.html). to learn more about them**
```
Haroenv: You can also link to `guide/connectors` it seems, that's what this redirects to. 

Unrelated, but this should be `check out` instead of `checkout`

Issue #144, opened 2017-06-28
When releasing, update all our examples react-instantsearch version


mthuret:
At some point we switched between latest and beta, but it would be better if those versions used on our example could follow the last version released.

Haroenv:
basically the same issue as algolia/instantsearch.js#2211

PR #208, opened 2017-07-26
chore(examples): update ris version when new release

mthuret:
See: https://github.com/algolia/react-instantsearch/issues/144

The idea is to update the `react-instantsearch` version of every examples when doing a new release. 

This script if very naive, if you have improvement idea @vvo, I'll take them :)

Also, with this version, if you still have a directory under examples that doesn't exist remotely anymore, it will still create a lock file...


mthuret:
@Haroenv forget to force push, shouldn't be there ;)

scripts/release.sh
```diff
@@ -118,3 +118,18 @@ npm run build-and-publish
...
+    cd $d
+    yarn upgrade react-instantsearch@$newVersion
+    cd ../../../..
+done
+
+commitMessage="chore(deps): update examples to react-instantsearch v$newVersion"
+git add .
+printf "$commitMessage" | git commit --file -
+
+printf "\n\nUpdate of react-instantsearch in all examples: almost done, check everything in another terminal tab if you want.\n"
```
vvo: Maybe split this two logs so it's easier to get that almost done is not for the react-instantsearch upgrade?

packages/react-instantsearch/examples/react-router/package.json
```diff
@@ -9,7 +9,7 @@
     "qs": "^6.3.0",
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
-    "react-instantsearch": "beta",
+    "react-instantsearch": "4.1.0-beta.3",
```
vvo: Don't we want our examples to rely on the stable version rather than beta?

mthuret: I want examples to rely on the latest existing version possible (latest or beta). Depending of the timeline we'll want them to be either on latest or beta (for instance, now, to be able to run the next-app).

vvo:
LGTM

PR #217, opened 2017-07-27
Format the count of items appearing in a refinement

AndrewNg:
A small fix to improve the UX of navigating refinements. Right now the numbers appear cluttered:
image
With this change, the numbers will look cleaner and will also be shown with language-sensitive representations.
image


Haroenv:
Thank you!

PR #224, opened 2017-07-29
Add an easy way to clone an example

Kikobeats:
Hello,

Inspired in [Next.js examples](https://github.com/zeit/next.js/tree/v3-beta/examples/basic-css) I added a tiny template into an example to make easy clone and try example.

Ideally could be adopted by all the example, what do you think 😄


packages/react-instantsearch/examples/server-side-rendering/README.md
```diff
@@ -1,3 +1,12 @@
+# Server Side Rendering
+
+## How to use
```
Haroenv: Maybe a different title here: 

> See the result first

Kikobeats: Maybe `Clone this example`?

mthuret: We could remove the title and use it as the first step of "to start the example". We can have a sentence like next.js =>  Download the example or clone the repo.

Also the title "The idea behind the example" doesn't feel right as the content doesn't correspond. Either we remove it, or we need to change the content. 

Otherwise I like the idea, but we should do it everywhere within this PR.

Kikobeats:
updated 👍

vvo:
Yo @Kikobeats why not just git clone the react-instantsearch repository? I never saw that kind of way to clone a GitHub repo.

vvo:
Yo @Kikobeats why not just git clone the react-instantsearch repository? I never saw that kind of way to clone a GitHub repo.

Haroenv:
@vvo, this only clones the folder you designate IIRC

mthuret:
The text is in a weird order =>  "This example shows how to do server side rendering and React InstantSearch" should come first. I'll update it as long as other recipes.

Kikobeats:
@vvo 

The command is cloning the repository, but just extracting the example 😄

Issue #23, opened 2017-04-11
Add a HierarchicalBreadCrumb widget


mthuret:
The helper is able to give us the Hierarchical Facet BreadCrumb through https://community.algolia.com/algoliasearch-helper-js/reference.html#AlgoliaSearchHelper#getHierarchicalFacetBreadcrumb

We should add a new widget/connector that allow a user to add this into its website. 

- [x] Propose an API for the connector (Provided Props / Exposed Props)
- [x] Propose an API for the widgets 
- [x] Propose a default style 
- [x] Decide if we go for an option within the connectHerarchicalMenu or with a new connector. 
- [x] Implement connector
- [x] Implement widgets
- [x] Add stories
- [x] Document the need for a virtual Hierarchical Menu for the default refinement

mthuret:
https://www.smashingmagazine.com/2009/03/breadcrumbs-in-web-design-examples-and-best-practices/

marielaures:
# ConnectBreadcrumb

## Exposed props

- attributes : List of attributes to use to generate the hierarchy of the menu. See the example for the convention to follow.
- separator : Symbol used for separating hyperlinks
- rootURL : The originating page (homepage)
- transformItems: Function to modify the items being displayed, e.g. for filtering or sorting them. Takes an items as parameter and expects it back in return.

## Provided props

- refine: A function to toggle a refinement
- createURL: A function to generate a URL for the corresponding search state
- items: the list of items the Breadcrumb can display

marielaures:
## Breadcrumb (Widget)

## Description
A breadcrumb is a secondary navigation scheme that allows the user to see where the current page is in relation to the website or web application’s hierarchy.
In terms of usability, using a breadcrumb reduces the number of actions a visitor needs to take in order to get to a higher-level page.

## Props
- attributes: List of attributes to use to generate the hierarchy of the menu. See the example for the convention to follow. (type: string)
- separator: Symbol used for separating hyperlinks. (type: string or React Component). (default: " > ")
- rootURL: the root element's URL (the originating page). (type: string) (default: null)
- transformItems: Function to modify the items being displayed, e.g. for filtering or sorting them. Takes an items as parameter and expects it back in return.

## Requirements
Breadcrumbs are used for websites with a large amount of content organised in a hierarchical manner. The typical example is an e-commerce website which has a large variety of products grouped into logical categories (with categories, subcategories which also have subcategories).

[...]

PR #228, opened 2017-08-01
feat(Breadcrumb): review

marielaures:
Implementing the Breadcrumb component.
You can see the issue here: https://github.com/algolia/react-instantsearch/issues/23
and check how it works in Storybook!


packages/react-instantsearch-theme-algolia/style.scss
```diff
@@ -3,111 +3,105 @@
...
+        transform: translateY(-50%)
+    }
 }
 
 @mixin openCaret() {
-  &:after {
-    top: 25%;
-    transform: rotate(90deg)
-  }
+    &:after {
```
Haroenv: I'm not sure if these style changes were on purpose. Could you do it in another PR if they're on purpose so that the diff is more readable?

packages/react-instantsearch/src/components/Breadcrumb.js
```diff
@@ -0,0 +1,92 @@
...
+    label: PropTypes.string.isRequired,
+    value: PropTypes.string.isRequired,
+  })
+);
+
+class Breadcrumb extends Component {
+  static propTypes = {
+    canRefine: PropTypes.bool.isRequired,
+    createURL: PropTypes.func.isRequired,
+    items: itemsPropType,
```
Haroenv: Not a proptypes expert, but this should be required too right, since there needs to be an array, with at least one item.

packages/react-instantsearch/src/components/Breadcrumb.js
```diff
@@ -0,0 +1,92 @@
...
+    refine: PropTypes.func.isRequired,
+    rootURL: PropTypes.string,
+    separator: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
Haroenv: With react 16, a string is an element too, does that matter here?

packages/react-instantsearch/src/components/Breadcrumb.js
```diff
@@ -0,0 +1,92 @@
...
+          onClick={() => (!this.props.rootURL ? refine() : null)}
+          href={this.props.rootURL ? this.props.rootURL : createURL()}
+        >
+          <span {...cx('rootLabel')}>
+            {translate('rootLabel')}
+          </span>
+          <span {...cx('separator')}>
+            {this.props.separator}
+          </span>
+
```
Haroenv: nit: I don't think this whitespace is needed

packages/react-instantsearch/src/components/__snapshots__/Breadcrumb.test.js.snap
```diff
@@ -0,0 +1,187 @@
...
+    <span
+      className="ais-Breadcrumb__rootLabel"
+    >
+      Home
+    </span>
+    <span
+      className="ais-Breadcrumb__separator"
+    >
+      <span>
+        🔍
```
Haroenv: The separator should be in-between the two links, no?

packages/react-instantsearch/src/connectors/hierarchicalMenuLogic.js
```diff
@@ -0,0 +1,229 @@
...
+import { SearchParameters } from 'algoliasearch-helper';
+import {
+  cleanUpValue,
+  getCurrentRefinementValue,
+  getIndex,
+  getResults,
+  refineValue,
+} from '../core/indexUtils';
+
+export const getId = props => props.attributes[0];
```
Haroenv: I don't really understand this file, can it be reused in `connectHierarchicalMenu` or is that a stupid idea?

mthuret: This file needs to be removed (was used before, then refactored).

packages/react-instantsearch/src/components/Breadcrumb.js
marielaures: So, I was asking myself the same question, but I checked on the HierarchicalMenu and on the List components, and apparently their `itemsPropType` isn't required either?

packages/react-instantsearch/src/components/Breadcrumb.test.js
```diff
@@ -60,11 +60,11 @@ describe('Breadcrumb', () => {
     );
 
     const items = wrapper.find('.ais-Breadcrumb__itemLink');
-    expect(items.length).toBe(4); //Home + Breadcrumb (3 levels)
+    expect(items.length).toBe(4);
 
     items.first().simulate('click');
     expect(refine.mock.calls.length).toBe(1);
-    expect(refine.mock.calls[0][0]).toEqual(); //Home = empty refinement
+    expect(refine.mock.calls[0][0]).toEqual();
```
Haroenv: I don't understand why you're calling `toEqual()` without any arguments. What about using [`.toBeNull()`](http://facebook.github.io/jest/docs/expect.html#tobenull)

packages/react-instantsearch/src/components/Breadcrumb.js
marielaures: @Haroenv Do you mean there is no need mention `PropTypes.string`?

Haroenv: That's what I mean, but for consistency, let's allow both until we don't have react 15 users anymore.

packages/react-instantsearch/src/components/__snapshots__/Breadcrumb.test.js.snap
```diff
@@ -0,0 +1,223 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Breadcrumb has a separator prop that can be a custom component 1`] = `
+<div
+  className="ais-Breadcrumb__root"
+>
+  <div
```
Haroenv: I think that a breadcrumb item should be a span, so that the default rendering without css is like what we expect: 

div|span
---|---
screen shot 2017-08-03 at 15 45 06 | screen shot 2017-08-03 at 15 45 00

packages/react-instantsearch/src/connectors/connectBreadcrumb.js
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return acc;
+  }, []);
+}
+
+/**
+ * The breadcrumb component is s a type of secondary navigation scheme that 
+ * reveals the user’s location in a website or web application.
+ * 
+ * If you want to use the "defaultRefinement" prop, you can do so by using a Virtual Hierarchical Menu
```
mthuret: I think this should be rephrase a bit. 
The breadcrumb doesn't have any `defaultRefinement` prop, so this might be confusing. 

Maybe have something like => "If you want to select a specific refinement for your Breadcrumb you'll need to use a Virtual Hierarchical Menu..." And then present a usage example. 

This should be present also on the widget guide and maybe only on this one.

PR #230, opened 2017-08-02
chore(lint): fix react instantsearch style

mthuret:
Our styles files are not taken into account by the linter and the fix. 
Will need to fix that.


mthuret:
@Haroenv, this is only for fixing @marielaures PR. I'll investigate later.

PR #231, opened 2017-08-03
Fix CSS classes shown for ais-RefinementList__itemCount

denkristoffer:
The two CSS classes shown for `.RefinementList__itemCount {}` and `.RefinementList__itemCountSelected` on https://community.algolia.com/react-instantsearch/widgets/RefinementList.html are currently missing the `ais-` prefix :)


mthuret:
Good catch @denkristoffer! Thank you very much!

Issue #185, opened 2017-07-25
Error: Cannot declare two hierarchical facets with the same name


harshmaur:
Hello, 

I am using react-instantsearch on next.js with ssr support. 
The problem I am getting is super difficult to reproduce but once it comes, it affects the whole project I can give the code and the stacktrace to help it debug properly. 

``` bash
Error: Cannot declare two hierarchical facets with the same name: `hierarchicalCategories.lvl0`
    at SearchParameters.addHierarchicalFacet (/cm-frontend-react/node_modules/algoliasearch-helper/src/SearchParameters/index.js:946:13)
    at Object.getSearchParameters (/cm-frontend-react/node_modules/react-instantsearch/src/connectors/connectHierarchicalMenu.js:1:3303)
    at c.filter.reduce._algoliasearchHelper.SearchParameters.index (/cm-frontend-react/node_modules/react-instantsearch/src/core/createInstantSearchServer.js:1:4336)
    at Array.reduce (native)
    at findResultsState (/cm-frontend-react/node_modules/react-instantsearch/src/core/createInstantSearchServer.js:1:4286)
    at Function._callee$ (/cm-frontend-react/.next/dist/pages/product-listing.js:411:54)
    at tryCatch (/cm-frontend-react/node_modules/regenerator-runtime/runtime.js:65:40)
    at Generator.invoke [as _invoke] (/cm-frontend-react/node_modules/regenerator-runtime/runtime.js:303:22)
    at Generator.prototype.(anonymous function) [as next] (/cm-frontend-react/node_modules/regenerator-runtime/runtime.js:117:21)
    at step (/cm-frontend-react/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /cm-frontend-react/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13
    at propagateObWrapper (/cm-frontend-react/node_modules/opbeat/lib/instrumentation/async-hooks.js:453:25)
    at instrumented (/cm-frontend-react/node_modules/opbeat/lib/instrumentation/index.js:102:27)
    at /cm-frontend-react/node_modules/opbeat/lib/instrumentation/async-hooks.js:490:70
    at instrumented (/cm-frontend-react/node_modules/opbeat/lib/instrumentation/index.js:102:27)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:169:7)
```

``` jsx
<HierarchicalMenu
     key="categories"
     attributes={[
[...]

PR #238, opened 2017-08-07
fix(SSR): clean SP before rendering again

mthuret:
See #185


mthuret:
Not easy to test as it'll require to expose a variable just for it. Or use the HM + Helper inside it. If you have an idea, I'll take it :)

Haroenv:
It's mostly because I didn't know why this change was made, maybe render twice and see if the result is correct?

mthuret:
I'm not sure to understand. This was failing only when implying HierarchicalMenu as it's the only facet that can be registered twice.

PR #250, opened 2017-08-09
docs(pkg.json): add some keywords

Haroenv:
These are useful metadata, definitely in Yarn


packages/react-instantsearch/package.json
```diff
@@ -7,6 +7,14 @@
   "main": "index.js",
   "module": "es/index.js",
   "description": "⚡ Lightning-fast search for React and React Native apps, by Algolia",
+  "keywords": [
+    "react",
+    "search",
+    "fast",
+    "algolia",
+    "components",
+    "react-native"
```
rayrutjes: Should we add instantsearch?

Haroenv: good call

PR #258, opened 2017-08-14
docs(refining-hits): add a tutorial on how to refine within hits

### LINKED ISSUE #244: Error associating filters per each item
Hello,

I'm trying to build a more mobile friendly way to apply filters.

screen shot 2017-08-09 at 08 36 22

The idea is, instead of have an aside with all filters to apply, each `Hit` show as `Tags` the filters that you can apply when you click on it.

In this case, my hit has:

```
{
  "condition": "New",
  "brand": "Gaastra",
  "model": "Cosmic"
}
```

So when I click in the `brand` Tag, the filter will be applied an the current hits will be refine with `brand:Gaastra` filter.

This make easy apply filter while you are navigating 😄

Implementation looks easy like:
https://gist.github.com/Haroenv/eb37d4c69df56ccbe6adec1877fdfc15

But internally something is not working and it throws warnings everywhere:

```
warning.js?07153df:35 Warning: flattenChildren(...): Encountered two children with the same key, `New`. Child keys must be unique; when two children share a key, only the first child will be used.
    in div (created by CurrentRefinements)
    in div (created by CurrentRefinements)
    in CurrentRefinements (created by Translatable(CurrentRefinements))
    in Translatable(CurrentRefinements) (created by AlgoliaCurrentRefinements(Translatable(CurrentRefinements)))
    in AlgoliaCurrentRefinements(Translatable(CurrentRefinements)) (at App.js:29)
    in div (created by InstantSearch)
    in InstantSearch (created by CreateInstantSearch)
    in CreateInstantSearch (created by CreateInstantSearchServer)
    in CreateInstantSearchServer (at App.js:15)
    in App (at index.js?entry:61)
    in div (created by styled.div)
    in styled.div (created by Provider)
    in ThemeProvider (created by Provider)
    in Provider (at index.js?entry:60)
    in div (at Layout.js:14)
    in Unknown (at index.js?entry:59)
    in _class (created by Container)
    in AppContainer (created by Container)
    in Container (created by App)
    in div (created by App)
    in App
```

I isolated the error into [windtoday@v3-tags](https://github.com/windtoday/windtoday-app/tree/v3-tags) If you want to reproduce it.

Haroenv:
this is a guide that finishes with https://codesandbox.io/embed/oY1klpZYB

fixes #244 and is obviously quite advanced

can be found in the deploy preview under `guide/Refining_hits.html`


docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
+---
+title: Refining within hits
+mainTitle: Guides
+layout: main.pug
+category: guide
+navWeight: # idk what to put here 🚨
```
Haroenv: what do we put here?

Haroenv:
https://deploy-preview-258--react-instantsearch.netlify.com/guide/Refining_hits.html

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+
+## Introduction
+
+In some interfaces, you want to have buttons _inside_ your Hits that will change the refinement of the current search, rather than as standalone components. Since we can't replicate this pattern with widgets, we will need to go with Refinementlists that don't render anything. Please note that this is an advanced tutorial and requires moderate knowledge of how React and React InstantSearch work. Our final result will look something like this, let's get started!
+
+<iframe src="https://codesandbox.io/embed/oY1klpZYB" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+## Regular setup
+
+First, we'll set up a regular project wit create-react-app and react-instantsearch.
```
rayrutjes: Wit => with

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+
+The index that we will use in this example needs to be set up with: 
+
+```jsx
+index.setSettings({
+  attributesForFaceting: ['searchable(keywords)', 'searchable(owner.name)'],
+});
+```
+
+An index that already has these settings is the index for search on [Yarn](https://yarn.fyi), which is used in this guide. We will get started in `App.js` and set up a regular React InstantSearch setup:
```
rayrutjes: In this guide we will use the search index we use on [Yarn](https://yarn.fyi). It already has the mandatory `attributesForFaceting ` configured.
We will get started in `App.js` and set up a regular React InstantSearch app:

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+        {owner.name}
+      </button>
+    </p>
+    <Tags keywords={keywords} />
+  </article>;
+
+export default Hit;
+```
+
+In this component we took the name, description and owner of a certain package, and render these as hits. To enable this as HitComponent, we will change the line in `App.js` that renders the `Hit` component:
```
rayrutjes: Maybe quote `HitComponent`?
Weird tense switch here. I would stay with the present everywhere.

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+        />
+      </InstantSearch>
+    );
+  }
+}
+
+export default App;
+```
+
+We manually render the `hitComponent` as a function, because we want to add extra functions to the `Hit` component later. Now we will also set up two `RefinementList`s for `keywords` and `owner.name`. We do this in a container so that the difference is clear, but obviously you can change the rendering of this however you prefer:
```
rayrutjes: Remove `obviously`

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+  render() {
+    return null;
+  }
+}
+
+const VirtualRefinementList = connectRefinementList(RefinementList);
+export default VirtualRefinementList;
+```
+
+Now this seems pretty useless, except that this component will get the correct props from its connector, each time something in its props change. The props we get from this connector are: `currentRefinement` and `defaultRefinement`. `defaultRefinement` is what we will manually set from within our Hit, `currentRefinement` is the current state of the search, relevant to that refinement, but maybe we should first implement it in our `App.js`:
```
rayrutjes: change => changes

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+        />
+      </InstantSearch>
+    );
+  }
+}
+
+export default App;
+```
+
+We will change the state of `App` from within the `Hit` component later so that it will update the `defaultRefinement` of each respective `VirtualRefinementList`. To be able to handle with that cleanly, let's add a lifecycle hook in `VirtualRefinementList`:
```
rayrutjes: To be able to achieve than in a clean way?

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+    }
+  }
+
+  render() {
+    return null;
+  }
+}
+```
+
+We compare the current refinement with the previous current refinement. If they aren't the same, then we will call a callback `onRefine` from our props with the current `attributeName` and with the new value. We do the same for the `defaultRefinement` so that the correct refinement is always chosen.
```
rayrutjes: then we call the `onRefine` callback

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+          ...state.refinements,
+          [attributeName]: newValue,
+        },
+      };
+    });
+  };
+}
+```
+
+This function works as follows: it has a `attributeName` and `value`, the value is what we want to add to the `VirtualRefinementList`, a string, for example `nodejs` or `react`. We will first check if our array for this attribute already contains that value. If it doesn't contain it, we simply add it to the array, if the array already contains the value we want, we remove it. Finally this value is set as the new state. Correspondingly, that change in state will trigger a change in `defaultRefinement` of the `VirtualRefinementList` for that attribute, and update the search interface.
```
rayrutjes: - it has a**n** `attributeName` and **a** `value`.
- remove the `simply`

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,551 @@
...
+        },
+      };
+    });
+  };
+}
+```
+
+This function works as follows: it has a `attributeName` and `value`, the value is what we want to add to the `VirtualRefinementList`, a string, for example `nodejs` or `react`. We will first check if our array for this attribute already contains that value. If it doesn't contain it, we simply add it to the array, if the array already contains the value we want, we remove it. Finally this value is set as the new state. Correspondingly, that change in state will trigger a change in `defaultRefinement` of the `VirtualRefinementList` for that attribute, and update the search interface. 
+
+But we still need to call this function. What we do is add `this.refine` as a prop to the `Hit` component, and call it there when needed. The inside of the `Hit` component now looks like this:
```
rayrutjes: The `Hit` component implementation now looks like this:

docgen/src/guide/Refining_hits.md
Haroenv: I think the version that’s written here is the correct version

docgen/src/guide/Refining_hits.md
mthuret: It's the number that allow you to positionate the guide in the sidebar. If you want it at the end, it should be bigger than the last one (not just by one to let other guide to be in between).

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,552 @@
...
+title: Refining within hits
+mainTitle: Guides
+layout: main.pug
+category: guide
+navWeight: # idk what to put here 🚨
+---
+
+## Introduction
+
+In some interfaces, you want to have buttons _inside_ your Hits that will change the refinement of the current search, rather than as standalone components. Since we can't replicate this pattern with widgets, we will need to go with Refinementlists that don't render anything. Please note that this is an advanced tutorial and requires moderate knowledge of how React and React InstantSearch work. Our final result will look something like this, let's get started!
```
mthuret: I think we should start directly by evoking the "List of tags" use case that speaks easily to somebody.

mthuret: Refinementlists => RefinementLists

Mention Virtual Widgets and link to the guide

In the end we should have a clear prerequisites list.

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,552 @@
...
+
+First, we'll set up a regular project with create-react-app and react-instantsearch.
+
+```sh
+$ yarn create react-app refining-hits # or npm install -g create-react-app && create-react-app refining-hits
+$ cd refining-hits
+$ yarn add react-instantsearch # or npm install --save react-instantsearch
+```
+
+The index that we will use in this example needs to be set up with:
```
mthuret: This part is confusing. The code sample uses an index that is already configured. If the user wants to use his own he will not have the same attribute name. I guess this should be rephrase a bit.

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,552 @@
...
+export default App;
+```
+
+If we look at what we have now, then we see that we can select any keyword, or any owner as expected, with buttons that don't do anything yet.
+
+<iframe src="https://codesandbox.io/embed/0gJW06BAX" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+## Adding virtual RefinementLists
+
+A "virtual" RefinementList is a RefinementList, just like the ones that are using the regular widgets, but with the exception that they won't render anything. We will make a virtual RefinementList here that will take `defaultRefinement` into account when rendering. We will make a `VirtualRefinementList` component, which uses a `connectRefinementList`:
```
mthuret: I believe we should introduce this paragraph by explaining why we need virtual widgets in the first place

mthuret:
If someone only wants the list of tags on Hits without the other RefinementList does it simplify things?

Haroenv:
You mean just `keywords` and not `owner.name`? Yes, that’s a bit more simple, but was pretty limiting. I like this solution better, since it’s simpler in the end. In the Yarn website it’s still using the “there can only be one attribute” way of doing things, and will probably change that to what I described here. 

Showing the real refinement list doesn’t make it more or less complex

mthuret:
I was talking about this: "Showing the real refinement list doesn’t make it more or less complex".

Haroenv:
Only difference is when you show the real refinementList that there are some errors in the console

mthuret:
Additional notes: I played with it a bit, it works like a charm except that it produces 3-4 query per selection. Do we have a way to optimize this @Haroenv?

Haroenv:
Hmm, should look at that issue @mthuret, thanks for noticing! I wonder what the cause would be, since it should just be another refinementList

Haroenv:
It's the `virtualRefinementList` that causes the queries, the actual widgets being there doesn't matter there has no effect

mthuret:
@Haroenv did you had the chance to understand why? Can we fix it? I'm not comfortable with shipping a guide that does all those extra queries and I'd rather try to find a solution that is cleaner. It might involved dedicated dev if nothing possible using React capabilities.

mthuret:
@Haroenv did you had the chance to understand why? Can we fix it? I'm not comfortable with shipping a guide that does all those extra queries and I'd rather try to find a solution that is cleaner. It might involved dedicated dev if nothing possible using React capabilities.

Haroenv:
No, I didn't find the actual reason yet, it doesn't seem easy to find out what the cause of the queries are. See also #355

vvo:
Quick note, is the status on this PR pending while we find reasons for https://github.com/algolia/react-instantsearch/issues/355?

Haroenv:
yep, this PR is blocked by #355

vvo:
@Haroenv Any way we can move forward with this PR? Is there anything we can do to help? Maybe the behaviour has changed now

Haroenv:
Interesting, with newest RIS and React they show up twice, still multiple queries per refinement too... 

https://codesandbox.io/s/oY1klpZYB

Haroenv:
Whoa, there's no longer a duplicate query done, will look at the comments, then it's mergeable

vvo:
> Whoa, there's no longer a duplicate query done, will look at the comments, then it's mergeable

How come? :D

Haroenv:
Nothing changed that I know of, but there's no longer a multiple query

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,552 @@
...
+  render() {
+    return (
+      <InstantSearch
+        appId="OFCNCOG2CU"
+        apiKey="f54e21fa3a2a0160595bb058179bfb1e"
+        indexName="npm-search"
+      >
+        <SearchBox />
+        <CurrentRefinements />
+        <Hits hitComponent={({hit}) => <Hit hit={hit} />}/>
```
samouss: We can't pass an anonymous function to the `hitComponent`. We can't use it as a render prop because, inside `Hits` the function will be used with `createElement`. Then since on every render the reference of the function will change React will unmount & mount the `Hit` component on every render.

I have made a little example of both approach, you can see that on each render (trigger by `forceUpdate`) the each `Hit` component is unmount / mount.

> https://jsfiddle.net/samuel_vllnt/uwpw621o/

Haroenv:
This is ready for review: https://deploy-preview-258--react-instantsearch.netlify.com/react-instantsearch/guide/refining_hits

The last section still needs to be verified based on #830

vvo:
Do we need anything else on this PR to move?

Haroenv:
needs another review, implementation of that last deduplication in the CodeSandbox and a release of that

samouss:
I will review it today or tomorrow.

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,591 @@
...
+          refinements: {
+            ...refinements,
+            [attributeName]: [...oldValue, value],
+          },
+        };
+      }
+
+      // if it's already refined, remove it from the refinements
+      const newValue = refinements[attributeName].slice();
+      newValue.splice(index, 1);
```
samouss: You can use a filter instead of a splice but it's matter of preference :)

samouss:
Note that for this kind of behaviour it might be interesting to have something like a "controlled" widgets. For example by providing the `currentRefinement` to the widget. The prop became the source of truth instead of the `searchState`. In that case it will allow us to get rid of the part with the double check on `defaultRefinement` + `currentRefinement` since the "real" value will always be the `currentRefinement`.

In React the `input` element works a bit like that:

- without a `value` the element is uncontrolled but you can still provide a `defaultValue` that will be used only for the first render (this is our behaviour)
- with a `value` the element is controlled and you are responsible for all changes

Haroenv:
Will that still allow changes from “upstream”?

samouss:
Sorry I didn’t see the question. What do you mean by that?

The version with the update on CurrentRefinements has been released so we can test the de duplication part.

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,591 @@
...
+    .map(item => {
+      delete item.__dedupe;
+      return item;
+    });
+
+class App extends Component {
+  render() {
+    return (
+      // the rest of the app
+      <CurrentRefinements transformData={deleteDuplicates} />
```
Haroenv: should be `transformItems`

+ it doesn't work with more than 1 duplicate 🤔

docgen/src/guide/Refining_hits.md
```diff
@@ -0,0 +1,591 @@
...
+}
+
+export default App;
+```
+
+We notice that in `CurrentRefinements`, each list show up twice. This is because the default current refinements widget will render each widget's refinements by default, and we have both a `RefinementList` and a `VirtualRefinementList` for the same attribute. As you might see in the console, this causes "duplicate key" errors by React. To prevent those, we can use `transformData` on `CurrentRefinements`. Each `item` of the current refinements has an `index` and `id`. These two properties explain which index the widget works on (see [multi index](guide/Multi_index.html)) and which the identifier of the widget has (this can be `'query'`, `attributeName`, `'page'` and some others). We can deduplicate the current refinements using those two properties like this:
+
+```jsx
+// ...
+const deleteDuplicates = items =>
```
Haroenv: this is wrong, should be: 

```js
export const deleteDuplicates = (items) =>
  items
    .map(item => {
      // make a temporary key for deduplication
      item.__dedupe = `${item.index}.${item.id}`;
      return item;
    })
    .filter(
      (obj, pos, arr) =>
        arr.map(item => item.__dedupe).indexOf(obj.__dedupe) === pos
    )
    .map(item => {
      // delete it afterwards
      delete item.__dedupe;
      return item;
    });
```

Haroenv:
```js
export const deleteDuplicates = (items) =>
  items
    .filter(
      (obj, pos, arr) =>
        arr.map(item => `${item.index}.${item.id}`).indexOf(`${obj.index}.${obj.id}`) === pos
    )
```

two less iterations (still O(n!))

samouss: Since the array is pretty small (less than 50 elements most of the time) I think the more understandable is the best 🙂

samouss:
@Haroenv I still have multiple queries but they won't be inevitable at some point since we play with the `defaultRefinement`. But for example when you click on the `CurrentRefinements` it triggers 4 requests. Maybe we should try remove some of them before merging this? I'm not sure that with the current state of the lib we could reduce the number of requests.

samouss:
We can't merge it right away because of the number of requests that are triggered. We keep the branch until we have a proper solution for the request issue.

PR #259, opened 2017-08-16
Fix/pagination

marielaures:
This fixes the pagesPadding offset: it used to be off-by-one (see screenshot attached where the padding is 3 by default).

screen shot 2017-08-16 at 10 22 08

**Summary**

Changed the offset + added some tests.

**Result**
The offset no longer exists and the widgets shows the same number of pages on each side of the currently selected page.
screen shot 2017-08-16 at 10 26 25


Haroenv:
Great! I think I opened an issue for that a while ago, spent some time trying to fix it, but couldn't manage to. Good job Marie-Laure 👏

Issue #102, opened 2017-05-23
MultiRange Label Object


flouc001:
# Question
Should [`MultiRange`](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/src/widgets/MultiRange.js) accept `Object` typed variables passed to it's label property to support JSX? Currently it throws a warning as it only expects type `String`.

mthuret:
Hi @flouc001 

That's a choice we made, translations such as `labels` are only accepting strings. 

If you wish to have a complete different render then you should use the `connectMultiRange` connector.

`MultiRange` also accepts a `transformItems` props if you need to add information without touching the DOM.

What are you trying to achieve in term of display?

flouc001:
Hi @mthuret 

That's unfortunate. The render is completely fine and gives me the desired look and feel however I get a PropType warning telling me I should be using a `String` when I am in fact passing a JSX `Object`.

It's just a warning so it can be ignored however it would be nice to add objects to the acceptable PropTypes. However like you say the decision was made.

Thanks for your reply!

mthuret:
I'm curious, can you share the look and feel of what you're doing?

flouc001:
So I have created a React `Component` called `Icon` which essentially takes properties and renders a FontAwesome formatted `<i>` element. I am replicating the `StarRating` widget from just using the `MultiRange` widget instead. To make it look nice I use FontAwesome [stars](http://fontawesome.io/icon/star/) for the label property. Here's an example:

    <MultiRange attributeName="rank" 
        items={[
            {start:0, label: (<span><Icon icon="star" /><span>{" & Up"}</span></span>)}
        ]}
    />

Obviously this example isn't complete but it gives an idea.

mthuret:
That's good to know. I'm closing this for now. 

I'll let you know if there's any evolution about this subject.

PR #266, opened 2017-08-18
[RFC] Adds support for any type of renderable element in List component

mschinis:
**Summary**
I have been using the `<MultiRange />` like this:

```jsx
<MultiRange
  attributeName="computed_price"
  items={[
    {
      end: 10,
      label: [
        <span key={0}>&lt; </span>,
        <Price key={1} pence={1000} displayCurrency="GBP" />
      ]
    },
    {
      start: 10,
      end: 100,
      label: [
        <Price key={0} pence={1000} displayCurrency="GBP" />,
        <span key={1}> - </span>,
        <Price key={2} pence={10000} displayCurrency="GBP" />
      ]
    },
    {
      start: 100,
      end: 500,
      label: [
        <Price key={0} pence={10000} displayCurrency="GBP" />,
        <span key={1}> - </span>,
        <Price key={2} pence={50000} displayCurrency="GBP" />
      ]
    },
    {
      start: 500,
      label: [
        <span key={0}> &gt; </span>,
        <Price key={1} pence={50000} displayCurrency="GBP" />
      ]
    }
  ]}
/>
```

According to the propTypes, defined in the MultiRange component, it's meant to accept any renderable element ([`PropTypes.node`](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/src/components/MultiRange.js#L13))

Using the component like mentioned above, would throw a proptypes validation error of:
```
Warning: Failed prop type: Invalid prop `items[0].label` of type `array` supplied to `List`, expected `string`.
    in List (created by MultiRange)
    in MultiRange (created by Translatable(MultiRange))
```

Digging further, the `MultiRange` component uses the `List` component whose propTypes [don't match the same propTypes](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/src/components/List.js#L8)

**Result**

The updated propTypes remove the warning from showing up. Accepting a `node` here makes sense, as it wouldn't start throwing any other warning in other components, whatever is passed to it, as long as it is an item that can be rendered.

**Related**

https://github.com/algolia/react-instantsearch/issues/102

### DIFF
--- a/packages/react-instantsearch/src/components/List.js
+++ b/packages/react-instantsearch/src/components/List.js
@@ -5,7 +5,7 @@ import SearchBox from '../components/SearchBox';
 const itemsPropType = PropTypes.arrayOf(
   PropTypes.shape({
     value: PropTypes.any,
-    label: PropTypes.string.isRequired,
+    label: PropTypes.node.isRequired,
     items: (...args) => itemsPropType(...args),
   })
 );


mschinis:
@mthuret any chance you can release a new version of the library containing the above fix?
I'm eager to remove the warning from my console without setting the dependency to master 😁

mthuret:
It's planned for monday next week (I'll release the 4.1.0), this will be included inside :)

mschinis:
awesome, thanks 🎉

PR #268, opened 2017-08-21
refactor(ScrollTo): remove usage of findDOMNodes to using ref callback

mthuret:
As you can see here when looking to the failed build https://github.com/algolia/react-instantsearch/pull/267 there is a rule saying `findDOMNode` should not be used because it will be removed in next version of React. 

See https://github.com/yannickcr/eslint-plugin-react/issues/678

The inconvenient with this refactoring is that it introduce a breaking change because of the need of an extra div. 

What do you think? Should we bump to the 5.0 now?


vvo:
What was the resulting DOM before and what it will be now? If we go from:
- div

to:
- div
  - div

I would say it's OK and does not need a BC.

mthuret:
Yes it just add an extra div (and therefore a new classname)

vvo:
> Yes it just add an extra div (and therefore a new classname)

I cannot think of any major outages this would cause to any website. Since there was no css class name before => nobody could style it. And styling a component that is hidden has no means (since we only use this component as a marker to scroll). What do you think?

mthuret:
Yep, right.

Issue #289, opened 2017-08-28
Pictures inside hits and slow network


mthuret:
Use case: you have a slow network and you're displaying pictures inside your hits. When searching as you type, old picture are still displayed even though the results are not the same anymore. Here is a gif explaining the issue better: 

![bug-picture](https://user-images.githubusercontent.com/6885378/29772852-d3493dd6-8bfa-11e7-81bd-cb855df04bcd.gif)

Explanation: Seems to be a behaviour of React, it doesn't remove the old image before the new one is loaded.

Proposal: the user could use a placeholder each time new image are loading. We could write a guide explaining the best practice and fix our current demos.

mthuret:
To solve this is to take advantage of the `onLoad` event of pictures to know when we can render it instead of the placeholder. The only difficulty here is that this manipulation needs to be repeated at every keystrokes/refinements and in a way that if the network is fast enough it doesn't look weird. 

For our example using this component https://github.com/FormidableLabs/react-progressive-image along with a white placeholder seems to do the trick. 

For our guide I'll see if I can wrote a code easy to grasp that does the same thing or if we should just indicate libraries that already handle those kind of use case.

PR #293, opened 2017-08-31
docs(hits): ensure key is always objectID

mthuret:
see https://github.com/algolia/react-instantsearch/issues/289


mthuret:
@Haroenv the Hits and InfiniteHits widget already set the key to `objectID`.

PR #300, opened 2017-09-01
chore(cleanup): Remove yarn-error.log file

dguo:
**Summary**
It looks like this file was added accidentally.

**Result**
N/A


mthuret:
@dguo Thank you very much for spotting and fix this!

PR #317, opened 2017-09-08
Attempting to make the virtual widget docs clearer

dericcain:
When I initially read the docs about virtual widgets, I did not understand why the virtual menu was being used and it was confusing. I have added a couple of sentences that should make understanding the use of the `VirtualMenu` more clear, which should, in turn, make the understanding of virtual widgets easier to understand.

If anything seems to be unclear or can be better, please let me know!

**Summary**

**Result**

### DIFF
--- a/docgen/src/guide/Virtual_widgets.md
+++ b/docgen/src/guide/Virtual_widgets.md
@@ -29,7 +29,7 @@ const App = () =>
     <Menu attributeName="fruits" defaultRefinement="Orange" />
   </InstantSearch>;
 ```
-In this case, we are using the `VirtualMenu` to pre refine our results (within the `clothes` search, only display `hoodies`). Think of the `VirtualMenu` as a hidden filter where we define attributes and values that will always apply the our search results.
+In this case, we are using the `VirtualMenu` to pre refine our results (within the `clothes` search, only display `hoodies`). Think of the `VirtualMenu` as a hidden filter where we define attributes and values that will always be applied to our search results
 
 **Notes**:
 * The `<Hoodies>` component is what we call a virtual widget.
--- a/docgen/src/guide/Virtual_widgets.md
+++ b/docgen/src/guide/Virtual_widgets.md
@@ -7,8 +7,9 @@ navWeight: 65
 ---
 
 A lot of websites have "category pages" where the search context is already refined without the user having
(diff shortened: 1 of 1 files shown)


docgen/src/guide/Virtual_widgets.md
```diff
@@ -28,6 +29,7 @@ const App = () =>
     <Menu attributeName="fruits" defaultRefinement="Orange" />
   </InstantSearch>;
 ```
+In this case, we are using the `VirtualMenu` to pre refine our results (within the `clothes` search, only display `hoodies`). Think of the `VirtualMenu` as a hidden filter where we define attributes and values that will always apply the our search results.
```
marielaures: I wonder if there is a typo here? Did you mean "define attributes and values that will always apply whenever the page is reloaded"?

marielaures: Suggestion: As a mean to impose a parameter chosen beforehand?

dericcain: @marielaures I believe I meant to say:
> ...define attributes and values that will always be applied to our search results.

I really wish I could use a picture here to illustrate it better so there is no confusion at all. I picture a set of checkbox filters where some selections are always selected, on every search, yet the checkbox filters are hidden from the viewer.

Anyhow, does that read better or do you have anything else better to clear it up?

dericcain: @marielaures I am sorry, but I am not sure I understand what you are trying to say. Can you elaborate?

Haroenv: Adding a picture is a great idea!

marielaures: @dericcain I was trying to rephrase the second sentence ("hidden filter where we define attributes and values..." because  "pre refine" might be a bit confusing! But I agree there it's not easy :)

dericcain:
@marielaures Hey! Something else that I ran into, if I am not mistaken, was the fact that I needed to have the attributes that I used in `attributeName` included in my facets for that particular index. When I did not have those facets, I was not getting any results.

If you confirm this, I would like to make another, separate, update to clarify that so that the docs are not making assumptions that someone would know that.

Thanks for your help with this!

mthuret:
Hey @dericcain, feel free to open another pull request if you think that the doc is still unclear and you want to improve it :)

Issue #233, opened 2017-08-03
Introduce a new API to handle conditional display use case


mthuret:
Currently, when some conditional display is needed (like a no results page), you're forced to use the `createConnector` API. 

This API is very internal to `react-instantsearch`and even if, for performing some conditional display you only touch its surface, it shouldn't be needed by our end user. We need a dedicated API. 

I can see several options: 

1. Easiest solution, you test the condition and rendered the children if it match. 

```jsx
<If condition={(props, searchState) => searchState.query}>QUERY</If>
<If condition={(props, searchState) => !searchState.query}>No Query</If>
```

2. A more complex version handling `else` use case. We can use the context to store if a condition has matched. 

```jsx
<Choose>
    <When condition={(searchState, searchResults) => searchState.query}>QUERY</When>
    <Otherwise>NO QUERY</Otherwise>
</Choose>
```

3. Another version (based on react-router API), that render the first matching condition. (Or every condition, but then don't see the point versus simple If). 

```jsx
<Switch>
   <When condition={(searchState, searchResults) => searchState.query}>QUERY</When>
   <When condition={(searchState, searchResults) => !searchState.query}>NO QUERY</When>
</Switch>
```

`condition` is a function that takes `searchState` and `searchResults` to still give the maximum of flexibility. 

For the record what we use to have before the first release was something like: 

`<NoResults>`, `<Results>`, `<Query>`, `<NoQuery/>`... this wasn't very extensible. 

WDYT?

vvo:
I very like the first approach, what does props contains?

mthuret:
Indeed `props` is not needed.

marielaures:
I really think it is a great addition, saving people from having to use the connector if all they need is to display a "no results" page.
I think I would prefer the second one for the sake of clarity: the condition is written only once, and it is made very clear what is going to happen if the condition is met, and what is going to happen if it isn't, all condensed in one place (the `<Choose />` component).

Haroenv:
We should take inspiration from `<Switch />` in React Router

vvo:
> We should take inspiration from <Switch /> in React Router

Can you provide more details? And maybe enhance the proposal.

Haroenv:
It just made me think of that feature, I haven't used it in-depth though, that's why I didn't go more into detail

mthuret:
That's another possible variation. Instead of having an "If, elseIf, else" pattern we can use a switch and render the first valid condition.

3. 
```
<Switch>
[...]

PR #357, opened 2017-09-20
feat(Conditional): add connectStateResults connector

mthuret:
Add a new connector that will remove the need of using the `createConnector` API to perform some conditional rendering. Could be used for any needs requiring access to the `searchState` or the `searchResults` . 

See: https://github.com/algolia/react-instantsearch/issues/233

- [x] Add a warning for people using `createConnector`


Haroenv:
Loving this change @mthuret, how would it look like with multi-query, is that with searchState?

docgen/src/guide/Conditional_display.md
```diff
@@ -6,89 +6,65 @@ category: guide
...
-      return {query: searchState.query};
-    },
- })(({query}) => {
-    const content = query
-      ? <div>The query {query} exists</div>
-      : <div>No query</div>;
-    return <div>{content}</div>;
- });
+const Content = connectStateResults(
+  ({ searchState }) =>
```
Haroenv: I'm wondering if we can't call this: `({state, results})`

mthuret: It's called `searchState` and `searchResults` everywhere especially in the documentation when looking for the structure so I'd rather keep the same name. If a user wants to call them state or results, he can definitely do it.

mthuret:
@Haroenv, I'm not sure to understand your question regarde multi query? (multi indices?) 
The `searchState` contains all the data, you can see in the doc how it's formatted in that case.

packages/react-instantsearch/src/connectors/connectStateResults.js
```diff
@@ -0,0 +1,62 @@
...
+ * @providedPropType {string} error - If the search failed, the error will be logged here. 
+ * @providedPropType {boolean} searching - If there is a search in progress. 
+ * @providedPropType {boolean} searchingForFacetValues - If there is a search in a list in progress. 
+ * @example
+ * import React from 'react';
+ *
+ * import { InstantSearch, Hits } from 'react-instantsearch/dom';
+ * import { connectStateResults } from 'react-instantsearch/connectors';
+ *
+ * const Content = connectStateResults(
```
Haroenv: I'm having a little doubts about the name here. I think we can find something that describes intent rather than what it is. 

`connectConditional` describes the use case better IMO

mthuret: This could be use in the future for future use case such as retrieving the `userInfo` field you can get with query rules (and probably a usage with analytics also). Then we need a name that could represent this. Cf discussion in the issue thread. If you have a better name you can still propose it.

vvo: There's no API feature in this connector that has any conditional mean (no if/else) so naming it conditional would be not the best to me too.

docgen/src/guide/Conditional_display.md
```diff
@@ -6,89 +6,65 @@ category: guide
...
+    searchState && searchState.query
+      ? <div>
+          The query {searchState.query} exists
+        </div>
+      : <div>No query</div>
+);
 ```
 
 ## Displaying content when there's no results
```
vvo: Should we have an example where we use data from multiple indices? I think we said something like "searchResults, allSearchResults".?

packages/react-instantsearch/src/connectors/connectHits.js
```diff
@@ -19,8 +19,9 @@ import { getResults } from '../core/indexUtils';
  * @example
  * import React from 'react';
  *
- * import { connectHits, Highlight, InstantSearch } from 'react-instantsearch/dom';
```
vvo: Was this a mistake previously?

stories/Conditional.stories.js
```diff
@@ -1,48 +1,46 @@
...
-      const content = noResults ? (
-        <div>No results has been found for {query}</div>
-      ) : (
-        <div>Some results</div>
-      );
-      return <div>{content}</div>;
-    });
+    const Content = connectStateResults(
+      ({ searchState, searchResults }) =>
+        searchResults && searchResults.nbHits !== 0 ? (
```
vvo: N.I.C.E

packages/react-instantsearch/src/connectors/connectHits.js
Haroenv: connectHits is from /connectors, so yes

stories/Conditional.stories.js
Haroenv: looks even nicer in future JS: 

```js
searchResults?.nbHits !== 0 ? 
```

docgen/src/guide/Conditional_display.md
mthuret: Yes I can add the code of the story for multi indices here 👍

mthuret:
warning added cc @vvo @Haroenv

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -272,6 +272,32 @@ export default function createConnector(connectorDesc) {
...
+                'refine',
+                'cleanUp',
+              ].indexOf(key) > -1
+          );
+
+          if (
+            onlyGetProvidedPropsUsage &&
+            !connectorDesc.displayName.startsWith('Algolia')
+          ) {
+            // eslint-disable-next-line no-console
```
Haroenv: I _LOVE_ this change!

PR #403, opened 2017-09-28
docs(createConnector): add links for more info to deprecation

Haroenv:
**Summary**

It's always useful to immediatly learn more and learn the solution of some problem that has been warned to you if available. React itself has short links for warnings (fb.me, we could also do that for alg.li, but maybe that's for later)

**Result**

adds two links to the createConnector deprecation


mthuret:
Thanks @Haroenv!!

PR #425, opened 2017-10-04
Fix connectRange ignores boundaries on first call

### LINKED ISSUE #384: connectorRange ignores min / max boundaries on first call
If you set boundaries on a widget related to `connectRange` (`RangeSlider`, `RangeInput`), there are ignored on the first call. In the given example the boundaries are set to `30` and `100`. The UI is align with the options but in the results we still have the products with prices out of bounds.

When you change the values of the slider, then reset it to the initial position the results is correct.

https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RangeSlider&selectedStory=custom%20min%2Fmax%20bounds&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs

samouss:
**Summary**

Fix #384 

This PR only fix the problem related to missing boundaries on first call.

**Result**

You can play with the `RangeInput` & `RangeSlider` widget on Storybook.

* set boundaries on the widget
* the results should normally be filter with the provided boundaries

**Before**:

screen shot 2017-10-04 at 18 36 21

**After**:

screen shot 2017-10-04 at 18 36 30

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectRange.js
+++ b/packages/react-instantsearch/src/connectors/connectRange.js
@@ -123,7 +123,7 @@ export default createConnector({
   getProvidedProps(props, searchState, searchResults) {
     const { attributeName, min: minBound, max: maxBound } = props;
     const results = getResults(searchResults, this.context);
-    const stats = results ? results.getFacetStats(attributeName) : {};
+    const stats = results ? results.getFacetStats(attributeName) || {} : {};
     const count = results
       ? results.getFacetValues(attributeName).map(v => ({
           value: v.name,
--- a/packages/react-instantsearch/src/connectors/connectRange.js
+++ b/packages/react-instantsearch/src/connectors/connectRange.js
@@ -32,6 +32,31 @@ function getId(props) {
 
 const namespace = 'range';
 
+function getCurrentRange(boundaries, stats) {
+  let min;
+  if (_isFinite(boundaries.min)) {
+    min = boundaries.min;
+  } else if (_isFinite(stats.min)) {
+    min = stats.min;
+  } else {
+    min = -Infinity;
+  }
+
+  let max;
+  if (_isFinite(boundaries.max)) {
+    max = boundaries.max;
+  } else if (_isFinite(stats.max)) {
+    max = stats.max;
+  } else {
+    max = Infinity;
+  }
+
+  return {
+    min,
+    max,
+  };
+}
+
 function getCurrentRefinement(props, searchState, context) {
   const refinement = getCurrentRefinementValue(
     props,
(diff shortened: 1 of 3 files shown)


mthuret:
I didn't take a close look at the PR yet, but it seems that this had an impact on the currentRefinements connector (see argos).

samouss:
Yes you are right. The thing is that when you set `max` and / or `min` on the `connectRange` connector your search will have by default the refinement set to the given values.

That's why the `currentRefinement` widget is active. That's maybe a wrong behaviour but it's the logical one. Any ideas on how to handle that?

packages/react-instantsearch/src/connectors/connectRange.js
```diff
@@ -49,6 +49,19 @@ function getCurrentRefinement(props, searchState, context) {
       return { min, max };
     }
   );
+
+  const hasMin = props.min !== undefined;
+  const hasMax = props.max !== undefined;
```
Haroenv: these can just be inlined I think

mthuret:
@samouss Maybe there's something to change on the StarRating implementation then because it's using the max/min attributes to setup the number of stars rows?

I think that we don't want to have an active `clearFilters` button that does nothing in case your using the StarRating.

packages/react-instantsearch/src/connectors/connectRange.test.js
```diff
@@ -140,11 +140,23 @@ describe('connectRange', () => {
     });
 
     it('refines the corresponding numeric facet', () => {
-      params = getSP(
+      params = getSearchParameters(
```
vvo: 👍

samouss:
After discuss with @mthuret, we found a solution for enable the `clearFilters` button only when the range & values are different.

mthuret:
@samouss for an unknown reason as for now, argos is blocked on this story if you run  `yarn run test` locally.

https://deploy-preview-425--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=StarRating&selectedStory=with%20filter%20on%20rating&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs

samouss:
@mthuret It's fixed! It was a bug related to how we compute the range. At first we used `Infinity` & `-Infinity` as boundaries. But the current implementation of `StarRating`, loop on the range for build the stars components. On the first render the `min` range was `-Infinity` and we end up with an infinite loop. So now the default value for the range are `undefined` to be backward compatible with the current widgets.

IMO it makes more sense to have a default range of `Infinity`. It's a better representation of missing values.

packages/react-instantsearch/src/connectors/connectRange.js
```diff
@@ -1,4 +1,5 @@
 import PropTypes from 'prop-types';
+import { isFinite as _isFinite } from 'lodash';
```
mthuret: By curiosity, why using lodash here instead of the native one?

mthuret:
Apart from the example, LGTM, thanks a lot for this! Really great stuff here :)

packages/react-instantsearch/src/connectors/connectRange.js
samouss: The global `isFinite` (see [reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite)) function is bit broken. For example:

```js
isFinite(null) // true
```

We should use `Number.isFinite` (see [reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite)), but the support is not so great! The `isFinite` function from `lodash` is based on `Number.isFinite` (see [docs](https://lodash.com/docs/4.17.4#isFinite)).

samouss:
The `RangeInput` problem doesn't seems to come from my code. Because other PR have the same problem (see #469).

PR #431, opened 2017-10-06
Migrate to React 16

mthuret:
Just to test with CI 

Remaining known issues: 

- [x] recipes test passes but there seems to have a "timeout" issue that make the script failed. For instance, when removing the content of the autocomplete tests, `npm run test:recipes` pass. 

- [x] storybook is failing like crazy. See: https://github.com/storybooks/storybook/issues/1965 (workaround found by deactivating animations on the sidebar).

- [ ] warning in test about Error Boundaries. See: https://github.com/facebook/react/issues/11098 (not mandatory)

### DIFF
--- a/packages/react-instantsearch/examples/autocomplete/package.json
+++ b/packages/react-instantsearch/examples/autocomplete/package.json
@@ -2,18 +2,19 @@
   "private": true,
   "devDependencies": {
     "identity-obj-proxy": "3.0.0",
+    "jest-environment-jsdom-11.0.0": "^20.0.9",
     "react-scripts": "1.0.14",
-    "react-test-renderer": "15.6.2"
+    "react-test-renderer": "16.0.0"
   },
   "dependencies": {
-    "antd": "^2.12.1",
-    "lodash": "^4.17.4",
-    "prop-types": "^15.5.10",
-    "react": "^15.6.1",
-    "react-autosuggest": "^9.3.1",
-    "react-dom": "^15.6.1",
-    "react-instantsearch": "^4.1.2",
-    "react-instantsearch-theme-algolia": "^4.1.2"
+    "antd": "2.12.1",
+    "lodash": "4.17.4",
+    "prop-types": "15.5.10",
+    "react": "16.0.0",
+    "react-autosuggest": "9.3.1",
+    "react-dom": "16.0.0",
+    "react-instantsearch": "4.1.3",
+    "react-instantsearch-theme-algolia": "4.1.3"
   },
   "scripts": {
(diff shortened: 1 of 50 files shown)


packages/react-instantsearch/examples/autocomplete/package.json
```diff
@@ -2,18 +2,19 @@
   "private": true,
   "devDependencies": {
     "identity-obj-proxy": "3.0.0",
+    "jest-environment-jsdom-11.0.0": "^20.0.9",
```
vvo: this is now an external package is that it?

packages/react-instantsearch/examples/autocomplete/package.json
```diff
@@ -23,7 +24,9 @@
     "notify": false,
     "moduleNameMapper": {
       "\\.(css)$": "identity-obj-proxy"
-    }
+    },
+    "setupFiles": ["../../../../shim.js"],
```
vvo: Maybe you can use <ROOT_DIR> smg here?

shim.js
```diff
@@ -0,0 +1,4 @@
+// see: https://github.com/facebookincubator/create-react-app/issues/3199
```
vvo: 👍  Love when we do that :)

vvo:
That's some serious chore/build stuff, much needed to ensure we are always up with the latest versions! GG

packages/react-instantsearch/examples/autocomplete/package.json
mthuret: http://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string

packages/react-instantsearch/examples/autocomplete/package.json
mthuret: Could be better indeed :)

mthuret: @vvo is <rootDir> here referring to the directory where the **example** package.json is, and therefore, can't work if shim.js is only at the very root of the project?

vvo: Indeed root_dir would be the example/package.json.

Also this seems temporary given comments in https://github.com/facebookincubator/create-react-app/issues/3199, just watch the issue and remove it afterwards

vvo:
LGTM

shim.js
Haroenv: It's already fixed in Jest, just not released yet; see https://github.com/facebook/jest/issues/4545#issuecomment-332762365

Issue #417, opened 2017-10-03
connectStateResults should give access to props


denkristoffer:
**Do you want to request a *feature* or report a *bug*?**

Bug

**Bug: What is the current behavior?**

I am using `createConnector` to access props, searchState and searchResults and do some calculations in the `getProvidedProps` function. After version 4.1.2 I get the warning that I should use the new `connectStateResults` instead. However that doesn't let me access props the way `createConnector` does.

**Bug: What is the expected behavior?**

Either `connectStateResults` should give access to props, or the warning shouldn't be thrown when `createConnector` is used to access props.

mthuret:
Hi @denkristoffer,

Thanks for your feedback, I actually wondered myself if `props` was something useful to access in this context. I'm curious to know what's your use case with them :)

denkristoffer:
I'm only using it to get the `attributeName`. The idea was to make my connector generic enough that it could be used for different attributes in the future.

```js
const rawData = searchResults.results.disjunctiveFacets.filter(
        facet => facet.name === props.attributeName,
      )
```

mthuret:
I see, I'll add it back soon :)

Issue #404, opened 2017-09-28
only display createConnector warning once


Haroenv:
**Do you want to request a *feature* or report a *bug*?**

bug(ish)

**Bug: What is the current behavior?**

the createConnector warning is fired on every search

**Bug: What is the expected behavior?**

the warning is fired only once

**Bug: What browsers are impacted? Which versions?**

any

**What is the version you are using? Always use the latest one before opening a bug issue.**
4.1.2

PR #434, opened 2017-10-06
feat(connectStateResults): add component props

mthuret:
See https://github.com/algolia/react-instantsearch/issues/417 and https://github.com/algolia/react-instantsearch/issues/404


mthuret:
@Haroenv instead of firing at every render, it'll fired only when mounting. If you meant having only one warning for the whole app, we can go there, but I'm not sure that's really needed. What do you think?

Haroenv:
@mthuret, perfect, I think I reviewed it and didn't see that change

Issue #215, opened 2017-07-27
Start testing React 16


vvo:
It's out! https://github.com/facebook/react/issues/10294

There will be soon a way to activate async rendering too (No idea what that means).

The sooner we try it the better usually for this kind of dependency.

Last time for React 15 we did not do it while we discovered breaking changes in React 15 (once released). See https://github.com/algolia/react-element-to-jsx-string/issues/56#issuecomment-261113118

Haroenv:
see #212 and #211

mthuret:
I'm currently in the process to write tests for all our recipes. We'll merge those PR afterwards (only concerns react-native example). 

And yes, we should tried the beta on all the rest of the code.

mthuret:
https://github.com/storybooks/storybook/issues/1868

mthuret:
Step to get there: 

- [ ] Ensure that storybook is working with 16

- [ ] Fix our Enzyme test using the new adapter strategy

boaticus:
As mentioned, defaultProps not being static on List is still throwing this warning:

`Setting defaultProps as an instance property on List is not supported and will be ignored. Instead, define defaultProps as a static property on List.`

mthuret:
@boaticus I'm in the middle of cleaning and upgrading everything :)

PR #442, opened 2017-10-09
fix(List): remove React16 warning

mthuret:
As I don't know how many times the switch to React16 for our tooling will last, let's release a new version of RIS that remove the List warning. 

See https://github.com/algolia/react-instantsearch/issues/215


vvo:
> As I don't know how many times the switch to React16 for our tooling will last

Do you mean that this warning is triggered by one of our tooling not yet being compatible with React 16? Which one?

mthuret:
@vvo sorry, it's the opposite. This warning is fired by our lib when used by our user. There's another PR taking care of migrating everything on React 16 (storybook, recipes, tests...) but because of some issues I can't merge it yet, that's why I'm fixing this warning separately.

vvo:
Small note, we never discussed this, but do you think our codebase is still React 15 compatible?

mthuret:
Actually this fix a bug in React 15 (`defaultProps` were not taken into account if not static).

PR #447, opened 2017-10-10
docs(connectStateResults): fix code sample

mthuret:
See: https://discourse.algolia.com/t/react-instantsearch-v4-1-3-is-out-tada/3167/2?u=marielaure.thuret

Code sample has a mistake.


mthuret:
@vvo I change the travis config because two lines seemed not to be taken into account. (the build was failing because of the release of yarn 1.2.0). What would be your strategy here to still have a reliable build but keeps updating this version from time to time?

PR #474, opened 2017-10-12
Feat/menu select

iam4x:
**Summary**

Displays a menu widget as a <select> DOM Element

**Result**

menu select

### DIFF
--- a/packages/react-instantsearch/src/components/MenuSelect.js
+++ b/packages/react-instantsearch/src/components/MenuSelect.js
@@ -28,7 +28,7 @@ class MenuSelect extends Component {
 
   get selectedValue() {
     const { value } = find(this.props.items, { isRefined: true }) || {
-      value: '',
+      value: 'ais__see__all__option',
     };
     return value;
   }
@@ -42,7 +42,7 @@ class MenuSelect extends Component {
   }
 
   handleSelectChange = ({ target: { value } }) => {
-    this.props.refine(value);
+    this.props.refine(value === 'ais__see__all__option' ? '' : value);
   };
 
   render() {
@@ -54,7 +54,7 @@ class MenuSelect extends Component {
         onChange={this.handleSelectChange}
         {...cx('select')}
       >
-        <option value="" {...cx('option')}>
+        <option value="ais__see__all__option" {...cx('option')}>
           {translate('seeAllOption')}
         </option>
 
--- a/packages/react-instantsearch/src/components/MenuSelect.js
+++ b/packages/react-instantsearch/src/components/MenuSelect.js
@@ -1,0 +1,73 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { find } from 'lodash';
+
+import classNames from './classNames.js';
+import translatable from '../core/translatable';
+
+const cx = classNames('MenuSelect');
+
+class MenuSelect extends Component {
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch/src/widgets/MenuSelect.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * The MenuSelect component displays a select that lets the user choose a single value for a specific attribute.
+ * @name MenuSelect
+ * @kind widget
+ * @requirements The attribute passed to the `attributeName` prop must be present in "attributes for faceting"
+ * on the Algolia dashboard or configured as `attributesForFaceting` via a set settings call to the Algolia API.
+ * @propType {string} attributeName - the name of the attribute in the record
+ * @propType {string} [defaultRefinement] - the value of the item selected by default
+ * @propType {function} [transformItems] - Function to modify the items being displayed, e.g. for filtering or sorting them. Takes an items as parameter and expects it back in return.
+ * @themeKey ais-MenuSelect__select - the select container of all items in the
```
mthuret: missing the end?

packages/react-instantsearch/src/components/MenuSelect.js
```diff
@@ -0,0 +1,73 @@
...
+        onChange={this.handleSelectChange}
+        {...cx('select')}
+      >
+        <option value="ais__see__all__option" {...cx('option')}>
+          {translate('seeAllOption')}
+        </option>
+
+        {items.map(item => (
+          <option key={item.value} value={item.value} {...cx('option')}>
+            {item.label} ({item.count})
```
mthuret: I'm wondering if there were a solution to allow our user removing count if they want to?

packages/react-instantsearch/src/components/MenuSelect.js
```diff
@@ -0,0 +1,73 @@
...
+      })
+    ),
+  };
+
+  static contextTypes = {
+    canRefine: PropTypes.func,
+  };
+
+  get selectedValue() {
+    const { value } = find(this.props.items, { isRefined: true }) || {
```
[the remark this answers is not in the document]
iam4x: I like the syntax and it's already used into `connectMultiRanges` so it does not increase the size of the build.

packages/react-instantsearch/src/components/MenuSelect.js
iam4x: Good idea, we can add a prop for it 👍

iam4x:
Should be good now, but I'm not sure about the naming of the prop to customize the label output: `formatItemLabel` what do you think @mthuret?

mthuret:
@iam4x is the feature allowed by `formatItemLabel` available on other menuSelect implementation? 

I'm wondering if this props will not be confused with `transformItems` (which is used to modify items data for instance).

What do you think @vvo @iam4x? To remove the count is this worth a prop or should we default on 'you should use a connector'.

iam4x:
This feature is not available in other `menuSelect` implementation, to keep consistency we should remove it or add it everywhere else.

I think it will confuse, I wanted to call it `transformItems` first.

mthuret:
Ok, so let's remove this option @iam4x. Sorry you developed it :(

iam4x:
@mthuret Option removed, rebased on master should be good 👍

Issue #405, opened 2017-09-28
connectStateResults, use undefined instead of null


Haroenv:
if undefined would be used for `searchState` and `searchResults`, it can be given a default parameter and destructured, but now it's null

mthuret:
@Haroenv, can you gave an example of what you're trying to write?

Haroenv:
I did an initial try of connectStateResult in Yarn: https://github.com/yarnpkg/website/commit/de24e47767ac9d7a9a174678f4fe4c28f9891bea

I'd like to write it like this, but it's telling me that `null` is not an object, and thus can't be destructured:

```jsx
const Results = connectStateResults(
  ({
    searchState: { query } = {},
    searchResults: { nbHits, nbPages } = {},
    onTagClick,
  }) => {
    if (isEmpty(query)) {
      body.classList.remove('searching');
      return <span />;
    } else if (nbHits === 0) {
      body.classList.add('searching');
      const docMessage = window.i18n.no_results_docsearch.split(/[{}]+/);
      docMessage[docMessage.indexOf('documentation_link')] = (
        <a href={`${window.i18n.url_base}/docs`}>{window.i18n.documentation}</a>
      );

      return (
        <div className="container text-center mt-5">
          <p>{window.i18n.no_package_found.replace('{name}', query)}</p>
          <p>
            {docMessage.map((val, index) => <span key={index}>{val}</span>)}
          </p>
        </div>
      );
    } else {
      body.classList.add('searching');
      return <ResultsFound pagination={nbPages > 1} onTagClick={onTagClick} />;
    }
  }
);
```

mthuret:
Makes sense :)

sarahdayan:
@Haroenv I guess we won't do it now that React InstantSearch Hooks is here and will soon replace React InstantSearch v6. Let's close?

PR #475, opened 2017-10-13
refactor(connectStateResults): remove usage of createConnector for conditional display

mthuret:
We were sill using `createConnector` on our examples.


docgen/src/examples/e-commerce-infinite/App.js
```diff
@@ -204,21 +204,13 @@ const Hit = ({ item }) => {
...
-  getProvidedProps(props, searchState, searchResults) {
-    const noResults = searchResults.results
-      ? searchResults.results.nbHits === 0
-      : false;
-    return { query: searchState.query, noResults };
-  },
-})(({ noResults, query }) => {
-  if (noResults) {
+const CustomResults = connectStateResults(({ searchState, searchResult }) => {
+  if (searchResult && searchResult.nbHits === 0) {
```
Haroenv: here I think you'd also have advantage of the destructuring as described in #405

PR #478, opened 2017-10-13
Fix RangeInput react@16

samouss:
**Summary**

Since we upgraded to `react@16` the `RangeInput` with a `Panel` has stopped working. The problem is that: for an strange reason when we call `setState` for update the current value inside the inputs, the lifecycle hook `componentWillReceiveProps` is called. Since we don't check that the previous values have **really** changed the state is always override with the original values from set on the `searchState`.

This is only a workaround the real problem is that the hook is called and it should not be called. I tried to dig a little and actually for `react` the `context` really changed before & after the setState that's why the hook is called. But I don't figure out why it changed... I tried to always return the same object from the `Panel::getChildContext`, it actually works but nobody does that and nobody have a problem with that ;) As I said in the code comment I don't think that's a bug related to `react` because I failed to reproduce the same behaviour with a simple hierarchy of components.

Actually inside the documentation of `react` they don't categorically say that `setState` never run the hook `componentWillReceiveProps`. But they say to **always** verify that real changes occur inside this hook.

[See documentation](https://reactjs.org/docs/react-component.html#componentwillreceiveprops)

**Result**

You can play with the `RangeInput` with `Panel` on Storybook. And you can also try the ecommerce example inside the documentation.

**Before**:

before

**After**:

after

### DIFF
--- a/packages/react-instantsearch/src/components/Panel.js
+++ b/packages/react-instantsearch/src/components/Panel.js
@@ -20,11 +20,16 @@ class Panel extends Component {
 
   constructor(props) {
     super(props);
-    this.state = { canRefine: true };
-    this.canRefine = canRefine => {
-      this.setState({ canRefine });
+
+    this.state = {
+      canRefine: true,
     };
   }
+
+  canRefine = canRefine => {
+    this.setState({ canRefine });
+  };
+
   render() {
     return (
       <div {...cx('root', !this.state.canRefine && 'noRefinement')}>
--- a/packages/react-instantsearch/src/components/RangeInput.js
+++ b/packages/react-instantsearch/src/components/RangeInput.js
@@ -50,9 +50,11 @@ export class RawRangeInput extends Component {
     // to avoid to override current state when values are not yet refined. In the react documentation,
     // they DON'T categorically say that setState never run componentWillReceiveProps.
     // see: https://reactjs.org/docs/react-component.html#componentwillreceiveprops
+
     if (
       nextProps.canRefine &&
-      (this.props.currentRefinement.min !== nextProps.currentRefinement.min ||
+      (this.props.canRefine !== nextProps.canRefine ||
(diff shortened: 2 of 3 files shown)


Haroenv:
@samouss Argos seems to show no more placeholder / default value, is that expected?

samouss:
@Haroenv Nope, it should be fixed now.

packages/react-instantsearch/src/components/Panel.js
```diff
@@ -20,11 +20,16 @@ class Panel extends Component {
...
-    this.canRefine = canRefine => {
-      this.setState({ canRefine });
+
+    this.state = {
+      canRefine: true,
     };
   }
+
+  canRefine = canRefine => {
+    this.setState({ canRefine });
```
Haroenv: maybe inline instead of a block? no big opinion on it

samouss: On a `class` I prefer to use block. It's a bit more readable, not a big opinion neither.

packages/react-instantsearch/src/components/RangeInput.js
```diff
@@ -6,8 +6,9 @@ import classNames from './classNames.js';
 
 const cx = classNames('RangeInput');
 
-class RangeInput extends Component {
+export class RawRangeInput extends Component {
```
mthuret: Is this just for testing purpose? What was the issue you have?

samouss: Yes, just for testing purpose. I want to shallow render the component in order to test some lifecycle hook where the problem come from.

mthuret: Is using `mount` would have change anything here?

samouss: I don't think so. Inside the test I use `setProps` to trigger the hook `willReceiveProps` and the method is only available on the root wrapper.

Issue #430, opened 2017-10-04
With `RangeInput` impossible to reset the refinement


samouss:
When you set a range with `RangeInput`, if you want to reset the refinement with the inputs it's impossible. The component disable the possibility to send empty value.

**Step to reproduce with `RangeInput`:**

* submit the form a first time
* set the inputs to an empty value
* submit the form a second time
* look at the query parameters the refinement is still set `"price>=0.39","price<=799"`

Is it the intentional behaviour?

[Storybook example]()

Issue #429, opened 2017-10-04
connectRange don't handle floating values


samouss:
When we set the input with a floating value it will be parse as an integer.

**Step to reproduce with `RangeInput`:**

* increase the min value to `3,45`
* submit your range
* the next value render in the min input will be `3`
* look at the query parameters you will have `"price>=3"` instead of `price>=3.45`

**Possible solution:**

We need to avoid to consider value as integer in `connectRange`. We also need to handle the floating number inside the inputs of `RangeInput`. At this time only a step of `1` is possible. We should provide an option for manage the [`step` attribute on inputs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Allowing_decimal_values).

[Storybook example](https://community.algolia.com/react-instantsearch/storybook/?knob-max=300&knob-translations=%7B%22submit%22%3A%22%20go%22%2C%22separator%22%3A%22to%22%7D&selectedKind=RangeInput&selectedStory=default&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs)

PR #491, opened 2017-10-16
Fix connectRange & RangeInput

### LINKED ISSUE #428: connectRange ignore boundaries when refine
When boundaries are set the user is still able to select a range outside of it.

**Step to reproduce with `RangeInput`:**

* set the max boundary to `300`
* set the max input to `450`
* look at the query parameters you will have `"price<=450"` instead of `price<=300`

**Possible solution:**

We have to handle the logic in `connectRange` to avoid input values to go above the limit (for user that use the connector API). But we also have to handle the limit in the `RangeInput` component to avoid end user to go above the limit in the input.

[Storybook example](https://community.algolia.com/react-instantsearch/storybook/?knob-max=300&knob-translations=%7B%22submit%22%3A%22%20go%22%2C%22separator%22%3A%22to%22%7D&selectedKind=RangeInput&selectedStory=playground&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs)

samouss:
**Summary**

*Sorry for the pretty long PR, I merged another fix into it and I shouldn't have.*

Fix #428 #429 #430

This PR aims to fix some problems related to `connectRange` & `RangeInput`. 

For `connectRange` improvements are:

- you can now reset the refinement
- you cannot refine with values outside of the range
- you can now specify the number of digits for decimal values with `precision` option (same as IS)

For `RangeInput` improvements are:

- the range is displayed with placeholders (same as IS)
- you can now reset the refinement (same as IS)
- you can refine with decimal value with the `precision` option (same as IS)
- you cannot refine out of bounds, thanks to the HTML5 validation (same as IS)

**Result**

You can play with the `RangeInput` on Storybook. You can also play with the IS version of it for compare the behaviour of both, you can access to it [here](https://deploy-preview-2440--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=RangeInput.default).


stories/RangeInput.stories.js
```diff
@@ -18,6 +18,26 @@ stories
...
+      <div style={{ display: 'none' }}>
+        <SearchBox defaultRefinement="ds" />
+      </div>
+    </WrapWithHits>
+  ))
+  .add('with precision of 2', () => (
+    <WrapWithHits linkedStoryGroup="RangeInput">
+      <RangeInput attributeName="price" precision={2} />
+    </WrapWithHits>
+  ))
```
mthuret: You can add `precision` as a field of the `playground` story

packages/react-instantsearch/src/connectors/connectRange.js
```diff
@@ -76,29 +77,86 @@ function getCurrentRefinement(props, searchState, context) {
...
-function refine(props, searchState, nextRefinement, context) {
-  if (
-    !_isFinite(parseFloat(nextRefinement.min)) ||
-    !_isFinite(parseFloat(nextRefinement.max))
-  ) {
-    throw new Error(
-      "You can't provide non finite values to the range connector"
-    );
+function refine(props, searchState, nextRefinement, currentRange, context) {
+  const { min: nextMin, max: nextMax } = nextRefinement;
```
mthuret: maybe we could create one or several intermediary functions?

package.json
```diff
@@ -131,7 +132,8 @@
     ],
     "setupFiles": [
       "./shim.js"
-    ]
+    ],
+    "snapshotSerializers": ["enzyme-to-json/serializer"]
```
mthuret: what does that bring?

mthuret:
I forget what `reset` the refinement means, could explain? :)

package.json
Haroenv: It should make sure that the snapshots made with enzyme methods like shallow are also looking JSX, like the raw jsx would be I think

see [yarn.pm/jest-serializer-enzyme](https://yarn.pm/jest-serializer-enzyme)

samouss: @Haroenv says right (the original package is [this one](https://github.com/adriantoine/enzyme-to-json))

It allows you to do:

```js
import { shallow } from 'enzyme';

const wrapper = shallow(MyComponent);

```

We don't rely on `react-test-renderer` inside the test anymore. You can handle all your tests on components with enzyme (Snapshot + assertions).

samouss:
The `reset` refinement means that you can undo your refinement. In the current implementation if you refine with `10` you can't reset the refinement via the input. You need to pass via the `Clear` refinements button. Now you can just reset the input.

**Before:**

before

**After:**

after

mthuret:
Hey @samouss could you check the argos differences? 

I spotted two I have questions about: 

- we now display 0 instead 0.39, is it expected? 
- there's also a difference with the slider, is that due to the precision modification?

samouss:
@mthuret Yep the difference on Argos are due to the precision modification. The default precision is `0` for share the same behaviour as `InstantSearch`. That's why the input is round to `0` instead to `0.39`. Same for the slider the changes is due to range that is bit longer than before. If we want to share the same behaviour as before, we just need to set the default precision to `2`.

mthuret:
@samouss I would go for keeping the same behaviour as before for now as I don't want to bump the version yet. 

However we can open an issue and handle this later for a v5. What do you think?

mthuret:
Some precision: when I say later, it'll be indeed part of the InstantSearch.css compliance version :)

samouss:
@mthuret Should be good now! The differences on Argos are related to the changes on the range. We don't display the `min` & `max` as value now but as `placeholder`. The behaviour for the query is the same.

PR #502, opened 2017-10-18
Remove "Unhandled promise rejection" from test

samouss:
**Summary**

The test on the CI throw and ignore an error because we access an `undefined` object. Locally `Jest` show some tests always running but don't throw an error.

**Result**

You can inspect the Travis output.

**Before:**

screen shot 2017-10-18 at 14 56 06

**After:**

*Pending Travis*


vvo:
Weirdly this should make the test fails no?

samouss:
Yep you are right @vvo! It should fail, but the test is async and `done` wasn't used so the test case only check the first assertion and finished.

Now it's fixed.

vvo:
Great, and from the Node.js message those misses will soon fail the process completely. Which is good. If something fails unexpectedly then we want to know it.

Issue #60, opened 2017-04-25
Add storybook-addon-jsx to display each widgets usage


mthuret:
With this storybook addon: https://github.com/Kilix/storybook-addon-jsx, we have the opportunity to display examples just near their story. It could also be an opportunity to remove all our static examples (with a link to the right story...).

PR #533, opened 2017-10-27
Docs/storybookjsx

marielaures:
**Summary**

Add a new Storybook addon to display the example usage of every widget (see #60).
 
**Result**

Our users can now have a better idea of how the widgets are used. 
They can see the widget in action and get the same configuration as the one displayed in the story.
image

It also works for the 'playground' story where the props can be modified.
addonjsx

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "rheostat": "2.1.1",
     "sass-loader": "6.0.6",
     "stat-mode": "0.2.2",
-    "storybook-addon-jsx": "^5.0.0",
+    "storybook-addon-jsx": "5.0.0",
     "string": "3.3.3",
     "style-loader": "0.19.0",
     "uglify-js": "3.0.25",
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "rheostat": "2.1.1",
     "sass-loader": "6.0.6",
     "stat-mode": "0.2.2",
+    "storybook-addon-jsx": "^5.0.0",
     "string": "3.3.3",
     "style-loader": "0.19.0",
     "uglify-js": "3.0.25",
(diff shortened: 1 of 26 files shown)


stories/Breadcrumb.stories.js
```diff
@@ -1,51 +1,66 @@
...
-    </WrapWithHits>
-  ))
+    ),
+    { displayName: changeDisplayName, filterProps: filteredProps }
+  );
+
+stories
   .add('with custom component', () => (
     <WrapWithHits hasPlayground={true} linkedStoryGroup="Breadcrumb">
       <Breadcrumb
```
Haroenv: These last two don't have jsx

stories/SearchBox.stories.js
```diff
@@ -1,65 +1,83 @@
...
+  .addWithJSX(
+    'with submit and reset components',
+    () => (
+      <WrapWithHits
+        searchBox={false}
+        hasPlayground={true}
+        linkedStoryGroup="SearchBox"
+      >
+        <SearchBox
+          submitComponent={<span>🔍</span>}
```
Haroenv: syntax highlighting was *off* in the live version here

Haroenv:
The `jsx` tab is also visible on those that don't have it, a little confusing, because I happened to click on Conditional and Configure first 

- Conditional
- Configure
- Snippet

marielaures:
Concerning the fact that we're displaying an empty JSX tab, I completely agree with you @Haroenv  but in some cases it can be even more confusing to output something that is either not relevant or messy (see example below):

image

It seems it is not possible to hide/display the panel conditionally. Happy to discuss it if you think of a better alternative!

package.json
```diff
@@ -113,6 +113,7 @@
     "rheostat": "2.1.1",
     "sass-loader": "6.0.6",
     "stat-mode": "0.2.2",
+    "storybook-addon-jsx": "^5.0.0",
```
samouss: The dependency should be pin like the others.

Haroenv:
Awesome 🎉 🌮

PR #541, opened 2017-10-31
Update Toggle.js type

davidfurlong:
From `connectToggle`:

```js
  propTypes: {
    label: PropTypes.string,
    filter: PropTypes.func,
    attributeName: PropTypes.string,
    value: PropTypes.any,
...
```

Additionally this is really confusing if it is supposed to be a string, how do you toggle on a boolean value?

Also I asked this here: https://discourse.algolia.com/t/how-to-create-a-toggle-for-the-absence-of-a-string-attribute/2460
but without response.

### DIFF
--- a/packages/react-instantsearch/src/widgets/Toggle.js
+++ b/packages/react-instantsearch/src/widgets/Toggle.js
@@ -12,7 +12,7 @@ import ToggleComponent from '../components/Toggle.js';
  *
  * @propType {string} attributeName - Name of the attribute on which to apply the `value` refinement. Required when `value` is present.
  * @propType {string} label - Label for the toggle.
- * @propType {string} value - Value of the refinement to apply on `attributeName` when checked.
+ * @propType {any} value - Value of the refinement to apply on `attributeName` when checked.
  * @propType {boolean} [defaultRefinement=false] - Default state of the widget. Should the toggle be checked by default?
  * @themeKey ais-Toggle__root - the root of the component
  * @themeKey ais-Toggle__checkbox - the toggle checkbox


mthuret:
Sorry I didn't see your question on discourse. Indeed this can work with a boolean. 

Thanks @davidfurlong! Did you manage to add it to your website since your question on discourse?

Haroenv:
I think @mthuret asked if you got the functionality working, this is only a warning, and should be okay to ignore until it’s released @davidfurlong

davidfurlong:
Yeah I found it wasn't working as expected. Additionally this string type
business has got me really confused, as I have experienced this issue with
the `Menu` widget as well.
I have a `Menu` widget as so:
```
<ExpiredMenu
                          attributeName="expired"
                          defaultRefinement="false"
                        />
```

and my `expired` attribute is a **boolean** in our index, with every index
item having a value `true` or `false` for it. The code above works, with
`false` as a string, but when I make it a boolean it shows the same issue:
```
Warning: Failed prop type: Invalid prop `defaultRefinement` of type
`boolean` supplied to `AlgoliaMenu(LocalTranslationProvider)`, expected
`string`.
```

davidfurlong:
I thought it didn't work, but it actually does, it seems that the `Menu`
widget has the same string type mistake. Might be worth going through all
the widgets Haroen and checking the types are correct, as its pretty
confusing when you read the docs and it tells you that the type is a
string: makes you think that you can only use this widget on certain types
of data (which isnt the case). Additionally I would like to suggest
feedback that its not clear how widgets will behave on attributes that are
arrays of values

PR #542, opened 2017-10-31
fix(searchbox): add maxlength 512

### LINKED ISSUE #510: Searchbox: limit to 512 characters

bit of both

Search interface accepts input over 512 character, but the engine does not

The interface to limit the search query

The workaround for now is to use the connectSearchBox and manually trimming the input or not allowing longer input

no option to add, just limit the input box

4.1.3

Haroenv:
**Summary**

The API can only handle queries up to 512 characters and would throw an error in this case (copy pasted addresses for example). 

**Result**

This prevents users from typing more than possible, and would simply discard later characters

fixes #510 (haha would be funny is that was issue 512)

supported in all browsers, but would fall back nicely anyway: http://caniuse.com/#feat=maxlength

cc @67ideas


mthuret:
Apparently you should use `maxLength` @Haroenv :)

Haroenv:
updated @mthuret

PR #544, opened 2017-10-31
feat: loading indicator

bobylito:
This PR adds a loading indicator based on the [proposal made here](https://github.com/algolia/instantsearch.js/issues/2315). However this version has changed a little:
 - the mechanics that determine if the search is stalled is implemented in the core of the library
 - there are options both on `<InstantSearch>` and `<SearchBox>`. The first one is for configuring the timer and the other is for displaying a loading indicator.

This is a POC implementation and the there is no design for the indicator (it's a red background) on purpose. Happy to discuss the API and the implementation.


mthuret:
@bobylito since you updated the yarn version under package.json you'll need to update also the travis script :)

bobylito:
> @bobylito since you updated the yarn version under package.json you'll need to update also the travis script :)

Sure will do. 👍

stories/SearchBox.stories.js
```diff
@@ -44,6 +44,11 @@ stories
       />
     </WrapWithHits>
   ))
+  .add('Display feedback when search is stalled', () => (
+    <WrapWithHits searchBox={false} linkedStoryGroup="SearchBox">
+      <SearchBox showLoading={true} />
+    </WrapWithHits>
+  ))
   .add('playground', () => (
     <WrapWithHits searchBox={false} linkedStoryGroup="SearchBox">
```
mthuret: Could you add the showLoading props as knos on playground story?

bobylito:
> If I'm not mistaken there's only the spinner missing here. Do you need action/help of someone here @bobylito?

No I'm good for the implementation. I'm not quite settled on the implementation for all InstantSearch. Once I'm sure of the consistency then I'll finish the feature :)

I'm wondering how to properly test that though... 🤔

anonrig:
We should totally add this to autoComplete connector.

bobylito:
I think that now it should be good to go :) cc @samouss

.travis.yml
```diff
@@ -9,7 +9,7 @@ addons:
 before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
-  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.2.0
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.2.1
```
samouss: The version should match the one in the `package.json` => `1.3.2`

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -172,6 +176,10 @@ class SearchBox extends Component {
       </svg>
     );
 
+    const loadingIndicatorComponent = this.props.loadingIndicatorComponent ? (
+      this.props.loadingIndicatorComponent
+    ) : <DefaultLoadingIndicator />;
```
samouss: We could provide a default props for `loadingIndicatorComponent` with the value `DefaultLoadingIndicator`. It will avoid the need to do the condition.

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -216,7 +224,15 @@ class SearchBox extends Component {
             />
           </symbol>
         </svg>
-        <div role="search" {...cx('wrapper')}>
+        <div
+          role="search"
+          {...cx(
+            'wrapper',
+            this.props.showLoadingIndicator &&
```
samouss: We can probably create a variable for ` this.props.showLoadingIndicator && this.props.isSearchStalled` since we are using this expression three times. It will be a bit less verbose in the code.

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -55,12 +56,19 @@ class InstantSearch extends Component {
     const initialState = this.isControlled ? props.searchState : {};
     this.isUnmounting = false;
 
+    const stalledSearchDelay =
+      props.stalledSearchDelay === undefined ||
+      props.stalledSearchDelay === null
+        ? 200
+        : props.stalledSearchDelay;
```
samouss: We could provide a default props for `stalledSearchDelay ` with the value `200`. It will avoid the need to do the condition.

packages/react-instantsearch/src/core/createInstantSearchManager.js
```diff
@@ -19,16 +20,21 @@ export default function createInstantSearchManager({
...
+  stalledSearchDelay,
 }) {
   const baseSP = new SearchParameters({
     ...searchParameters,
     index: indexName,
     ...highlightTags,
   });
 
+  let stalledSearchTimer = null;
+  let isSearchStalled = false;
```
samouss: You could store the variable `isSearchStalled` directly into the store. We avoid to keep track of the same value twice.

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -25,6 +26,9 @@ class SearchBox extends Component {
     onReset: PropTypes.func,
     onChange: PropTypes.func,
 
+    isSearchStalled: PropTypes.bool,
```
samouss: We could set this value required since `isSearchStalled` will be always pass to the component.

packages/react-instantsearch/src/components/List.js
```diff
@@ -134,6 +134,7 @@ class List extends Component {
               selectItem(items[0], this.resetQuery);
             }
           }}
+          showLoadingIndicator={false}
```
samouss: We could provide a default props for `showLoadingIndicator` with the value `false`. It will avoid the need to force the value to `false` when we don't want the indicator to display.

packages/react-instantsearch/src/components/SearchBox.js
bobylito: From a usage perspective, I would argue that this value is only useful if `showLoadingIndicator` is true and the only value that we really about is `true`. Considering that, do you think it should still be marked as required?

packages/react-instantsearch/src/components/SearchBox.js
bobylito: And do you have an idea for a name for this variable?

bobylito: Nevermind, it's gonna be isSearchStalled :)

packages/react-instantsearch/src/components/SearchBox.js
samouss: We can leave it as it it then, but when we define an optional props it's recommended to always give them a default value. So we can put a default value to `false` on the `isSearchStalled`. WDYT?

bobylito: Ok WFM 👍

bobylito:
I took care of your comments. Feel free to have another look ;)

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -227,12 +227,14 @@ export default function createConnector(connectorDesc) {
           metadata,
           resultsFacetValues,
           searchingForFacetValues,
+          isSearchStalled,
         } = store.getState();
         const searchState = {
```
samouss: We can take advantage of your changes for rename the variable to `searchResults`. The third parameter pass to `getProvidedProps` is `searchResults`. The second one is `searchState`. So it's bit confusing to have `searchResults` called `searchState` 🙂

> https://community.algolia.com/react-instantsearch/guide/Custom_connectors.html#getprovidedpropsprops-searchstate-searchresults-meta-searchforfacetvaluesresults

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -233,9 +274,22 @@ class SearchBox extends Component {
...
+          {this.props.showLoadingIndicator && (
+            <div
+              style={{
+                display: isSearchStalled ? 'block' : 'none',
+              }}
+              {...cx('loading-indicator')}
+            >
+              {loadingIndicatorComponent}
+            </div>
+          )}
```
samouss: We could add a test for this logic. Verify depending on the given props that the loading component is displayed or not (with a Snapshot).

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -156,21 +203,9 @@ class SearchBox extends Component {
...
-    const resetComponent = this.props.resetComponent ? (
-      this.props.resetComponent
-    ) : (
-      <svg role="img" width="1em" height="1em">
-        <use xlinkHref="#sbx-icon-clear-3" />
-      </svg>
-    );
+    const submitComponent = this.props.submitComponent;
+    const resetComponent = this.props.resetComponent;
+    const loadingIndicatorComponent = this.props.loadingIndicatorComponent;
```
samouss: We could use destructuring assignment for those variables.

bobylito:
Happy to oblige :) Here you are :)

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -203,9 +203,11 @@ class SearchBox extends Component {
...
     const query = this.getQuery();
 
-    const submitComponent = this.props.submitComponent;
-    const resetComponent = this.props.resetComponent;
-    const loadingIndicatorComponent = this.props.loadingIndicatorComponent;
+    const {
+      loadingIndicatorComponent,
+      resetComponent,
+      submitComponent,
+    } = this.props;
```
samouss: We could merge the destructuring props with the one above. Because we destruct the same attributes two times (see line `203`).

PR #547, opened 2017-11-01
fix(babelrc): add a key for each env development, production, es

### LINKED ISSUE #499: /es/ folder doesn't contain es2015 code, but commonjs

bug I guess

the es folder contains the same as the main folder

see https://unpkg.com/react-instantsearch@4.1.3/es/

the es folder to contain es2015 modules

4.1.3

samouss:
**Summary**

Fix #499 

Since `babel@7` is not yet release, I just updated the `.babelrc` to correctly output ES modules.
The merge of the presets with the `env` seems to not work, so I duplicated each configuration. As soon as `babel@7` is out we should migrate to `.babelrc.js` as @Haroenv suggest.


Haroenv:
Good idea, it breaks the tests though, I know that Jest can only have the `test` env, and not the fallback `development`, might be related

samouss:
Yep, you are right I forgot that... Should be good now.

PR #575, opened 2017-11-05
Fix server side rendering example

samouss:
**Summary**

The `server-side-rendering` example is broken. The build fail due to an incompatibility between `babel-core@7.x` & `babel-preset-react@6.x`.

- Fix the build with a rollback on `babel@6.26.0`
- Remove all the dependencies related to styling, because we don't use any style in the example
- Remove the React warning in the console, because we use `render` instead of `hydrate`

**Before:**

screen shot 2017-11-05 at 16 02 39

**After:**

screen shot 2017-11-05 at 16 05 45


Haroenv:
seems good!

mthuret:
@samouss did you see it with a renovate PR failing? If not, do we have a way to test what was failing?

samouss:
Nope it wasn't with Renovate, I tried to run the example. One thing we can do is to run the `build` task on each recipe when we trigger the `test:recipes`. It will verify that one the build is correctly generate and then that the tests pass. The only downside of that, it's that will increase the CI build time.

WDYT?

mthuret:
Personally I prefer to have a CI a little longer but that covers more failures :)

samouss:
@mthuret I'm okay with that 👍

PR #578, opened 2017-11-06
Update happo-* to v5

samouss:
**Summary**

Update all the `happo` dependencies at the same time in order to avoid the build to fail. I also bump the version of FF to `56.0` since the test suite of `happo` is running on this version.

https://github.com/Galooshi/happo/blob/master/.travis.yml#L10


samouss:
Will see 🤞

samouss:
Actually it works. The visual differences are due to the changes in FF. I checked a bit the different pictures, the major differences are in padding, input & typography.

Haroenv:
I manually approved on the Argos site, you can do that too if you log in with GitHub @samouss

samouss:
Cool! Thanks for the information.

PR #565, opened 2017-11-03
docs(storybook): add a11y addon

marielaures:
**Summary**

Add a new Storybook addon to monitor our components for being accessible.
See #221.

**Result**

There is a new "Accessibility" tab displayed on the right sidebar that shows both the a11y "violations" and the "passes". 

image

PR #579, opened 2017-11-06
docs(storybook): refactor Storybook addon addition

marielaures:
**Summary**

Following up on #565 and @samouss's comment, this PR aims at simplifying the process of adding a new Storybook addon.

**Result**

The addons are added at the global level in the Storybook config file, using the built-in feature rather than adding them on every story or using a custom function. Adding a new addon will be easier (if it is meant to be applied to all the stories): it will only consist in a single line in the config file!


stories/Breadcrumb.stories.js
```diff
@@ -1,24 +1,27 @@
...
+import { object, text } from '@storybook/addon-knobs';
+import {
+  createStoryWithAddons,
+  displayName,
+  filterProps,
+  WrapWithHits,
+} from './util';
 import JSXAddon from 'storybook-addon-jsx';
 
 setAddon(JSXAddon);
```
Haroenv: Could this also be done in the `createStoryWithAddons`?

samouss:
Actually we can set decorators globally & addons globally, don't know if it's a best practice or not! @mthuret @Haroenv any opinion / informations about that? 

https://github.com/airbnb/react-dates/blob/master/.storybook/config.js

```js
import { configure, setAddon, addDecorator } from '@storybook/react';
import { withKnobs } from '@storybook/addon-knobs';
import { checkA11y } from 'storybook-addon-a11y';
import JSXAddon from 'storybook-addon-jsx';

addDecorator(withKnobs);
addDecorator(checkA11y);

setAddon(JSXAddon);

configure(() => {}, module);
```

Haroenv:
I think I prefer the global method, since there's no way to enable or disable the display of addons per story.

samouss:
I also prefer because we can remove our custom wrapping function and use the built in feature.

mthuret:
Yep let's add them globally :)

PR #603, opened 2017-11-13
Chore: Generate UMD build with Rollup

### LINKED ISSUE #593: Migration to Rollup
Today we use Webpack for bundle the UMD build & a custom UglifyJS script to minify and generate the sourcemap. We could probably use Rollup instead for three main reasons:

- remove the hard dependency on Webpack & UglifyJS script for the build
- probably save a bit of Kb on the payload of the library
- simplify the build process

**Metrics with Webpack:**

| File            | Normal   | Minify + Gzip |
| :--------------- | :---------| :------------- |
| `Core.js`       | 115Kb    | 8.52 kB       |
| `Dom.js`        | 979Kb    | 69.7 kB       |
| `connectors.js` | 626kB    | 42.9 kB       |

samouss:
**Summary**

Closes #593 

Use Rollup only for generate the UMD build, previously Webpack. The gain is more in term of maintainability rather than performance. We gain some Kb but not so much:

**Before:**

| File            | Normal   | Minify + Gzip |
| :--------------- | :---------| :------------- |
| `Core.js`       | 115Kb    | 8.52 kB       |
| `Dom.js`        | 979Kb    | 69.7 kB       |
| `connectors.js` | 626kB    | 42.9 kB       |

**After:**

| File            | Normal   | Minify + Gzip |
| :--------------- | :---------| :------------- |
| `Core.js`       | 89Kb **(-22%)**   | 7.28 kB **(-14%)**      |
| `Dom.js`        | 789Kb **(-19%)**   | 59.83 kB **(-14%)**       |
| `connectors.js` | 536kB **(-15%)**    | 37.22 kB **(-13%)**      |


packages/react-instantsearch/rollup.config.js
```diff
@@ -0,0 +1,112 @@
...
+        'AlgoliaSearchHelper',
+        'SearchParameters',
+        'SearchResults',
+        'url',
+      ],
+    },
+  }),
+  // Useful for define process.env variable, inside the AlgoliaSearchClient we
+  // have a dependency on it for RESET_APP_DATA_TIMER.
+  // see: http://bit.ly/2zGLrsu (Permalink was too long)
```
vvo: Would still put `https://github.com/algolia/algoliasearch-client-javascript/blob/bb5d2ce16cf4b5a6e6de8ebae2de2880f783d967/src/AlgoliaSearchCore.js#L13`. eslint is fine with long comments of urls.

vvo: Also that process.env in the client should already be resolved/set in the build of the js client no?

vvo:
> The gain is more in term of maintainability rather than performance.

In term of file size reduction that's still impressive for just a bundler change really, -10% on gzip size is big to me. Would love to see that on instantsearch.js too!

In term of maintainability, can you precise why it's better?

package.json
```diff
@@ -109,13 +110,20 @@
...
+    "rollup-plugin-uglify": "2.0.1",
     "sass-loader": "6.0.6",
     "stat-mode": "0.2.2",
     "storybook-addon-a11y": "3.1.9",
     "storybook-addon-jsx": "5.0.0",
     "string": "3.3.3",
     "style-loader": "0.19.0",
-    "uglify-js": "2.7.5",
     "webpack": "3.8.1",
     "webpack-bundle-analyzer": "2.9.1",
```
Haroenv: same for those I guess, cc https://github.com/algolia/react-instantsearch/pull/603/files#r151405932

packages/react-instantsearch/package.json
```diff
@@ -1,39 +1,39 @@
 {
   "name": "react-instantsearch",
+  "version": "4.2.0",
```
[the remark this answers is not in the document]
Haroenv: seems so

packages/react-instantsearch/rollup.config.js
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import resolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
+import babel from 'rollup-plugin-babel';
+import globals from 'rollup-plugin-node-globals';
+import replace from 'rollup-plugin-replace';
+import uglify from 'rollup-plugin-uglify';
+import filesize from 'rollup-plugin-filesize';
+/* eslint-enable import/no-extraneous-dependencies */
```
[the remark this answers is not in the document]
Haroenv: I think those comments can simply be removed, since you use them later 🤔

packages/react-instantsearch/package.json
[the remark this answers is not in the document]
samouss: Yes it is!

package.json
```diff
@@ -109,13 +110,20 @@
...
+    "rollup-plugin-replace": "2.0.0",
+    "rollup-plugin-uglify": "2.0.1",
     "sass-loader": "6.0.6",
     "stat-mode": "0.2.2",
     "storybook-addon-a11y": "3.1.9",
     "storybook-addon-jsx": "5.0.0",
     "string": "3.3.3",
     "style-loader": "0.19.0",
-    "uglify-js": "2.7.5",
     "webpack": "3.8.1",
```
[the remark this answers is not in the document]
samouss: We still need Webpack because of the documentation part.

package.json
samouss: Actually it's referenced inside the documentation pipeline so I don't touch it.

https://github.com/algolia/react-instantsearch/blob/master/docgen/webpack.config.build.babel.js#L5

packages/react-instantsearch/rollup.config.js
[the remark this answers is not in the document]
samouss: > Why not add this to our global eslint rules? https://github.com/algolia/eslint-config-algolia/

Actually the rule is pretty useful most of the time. We will see if the resolution problem is still present with the workspaces.

> I think those comments can simply be removed, since you use them later 🤔

I don't understand your remark. If I remove the comment ESLint will trigger warning about external dependencies import. It's because modules are installed in the root folder instead of this one.

packages/react-instantsearch/rollup.config.js
samouss: No really because Rollup will resolve the file [algoliasearchLite](https://github.com/algolia/algoliasearch-client-javascript/blob/cd22e01862f2612b22b595b1815968ad1daec148/src/browser/builds/algoliasearchLite.js) according to the [import](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/dom.js#L3) & the [`package.json`](https://github.com/algolia/algoliasearch-client-javascript/blob/master/package.json#L9). This entry point require `AlgoliaSearchCore` and so the process.env is still present when Rollup do the build.

What we could do also if you are sure to not use Node globals elsewhere, is to replace **all** `process.env` calls with their corresponding values at build time (like we do with `process.env.NODE_ENV`). We can gain maybe 1-2k of payload size without using the Node globals plugin.

samouss:
> In term of maintainability, can you precise why it's better?

IMO it's easier because we now have only one tools for generate the UMD build which is Rollup. Before we relied on three different tools with they own configurations & complexities (Webpack + UglifyJS + bash script). 

The configuration is centralised in one file, and this file is responsible for the bundling + magnification. Nothing is spread across different place.

packages/react-instantsearch/rollup.config.js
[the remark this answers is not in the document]
Haroenv: misunderstood which rule was disabled; lgtm

packages/react-instantsearch/rollup.config.js
Haroenv: I'll replace that in the client by an option instead.

vvo: Ok so we just ensure process.env is defined and then we use the default value in the client. WFM.

PR #610, opened 2017-11-15
fix dom

JonasBa:
pug is auto closing the head tag inside the file, thus they have to be escaped and set properly in the html otherwise services like DocSearch break because they can't parse HTML.

cc @maxiloc


Haroenv:
@JonasBa, can you make the same edit as the subsequent PR in vue-instantsearch?

maxiloc:
Jonas is a bit under water right now. Feel free to do it.

PR #618, opened 2017-11-17
Chore: update Expo & React Native

samouss:


packages/react-instantsearch/examples/react-native/tests/__snapshots__/App.test.js.snap
```diff
@@ -278,7 +277,7 @@ exports[`renders correctly 1`] = `
                   <Text
                     accessible={true}
                     allowFontScaling={true}
-                    disabled={false}
+                    disabled={undefined}
```
Haroenv: why not get rid of the prop then?

samouss: We do a full rendering for taking the Snapshot, so it's not generated by us. It's the default value on the `Button` component in React Native. You can check on the given file, there is not default `props` for `disabled` but it's always pass to the child component.

https://github.com/facebook/react-native/blob/master/Libraries/Components/Button.js

samouss:
Any blockers @Haroenv?

PR #641, opened 2017-11-22
Use state updater pattern

jaredpalmer:
Whenever you need to adjust state based on previous state, better to use the updater/reducer pattern of setState.

### DIFF
--- a/packages/react-instantsearch/examples/geo-search/src/App.js
+++ b/packages/react-instantsearch/examples/geo-search/src/App.js
@@ -33,7 +33,6 @@ class App extends Component {
   // when a click on the map is performed, the query is reset and the aroundLatLng property updated.
   onLatLngChange({ lat, lng }) {
     this.setState(prevState => ({
-      ...prevState,
       searchState: {
         ...prevState.searchState,
         query: '',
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch/examples/geo-search/src/App.js
```diff
@@ -32,10 +32,14 @@ class App extends Component {
 
   // when a click on the map is performed, the query is reset and the aroundLatLng property updated.
   onLatLngChange({ lat, lng }) {
-    this.setState({
-      searchState: { ...this.state.searchState, query: '' },
+    this.setState(prevState => ({
+      ...prevState,
```
Haroenv: I thought the updater did a shallow merge, so there’s no need to spread

packages/react-instantsearch/examples/geo-search/src/App.js
```diff
@@ -32,10 +32,14 @@ class App extends Component {
...
+      searchState: {
+        ...prevState.searchState, 
+        query: ''
```
Haroenv: Can you add a trailing comma so prettier is happy?

vvo:
I wonder if:
1. this can trigger different behaviour for our users
2. it can actually fix some bad behaviours we had recently (infinite loops, double requests)

cc @samouss

Haroenv:
I think this will fix 2 indeed (no out-of-order updates), but this is just the example, we need it in some code too I think @vvo

jaredpalmer:
😎

samouss:
> 1. this can trigger different behaviour for our users

Even it's only in the example, the impact on the users should be null.

> 2. it can actually fix some bad behaviours we had recently (infinite loops, double requests)

I don't think so, not yet investigate this behaviours.

PR #654, opened 2017-11-26
Refactor Hits component

samouss:
Small refactoring around the `Hits` component:

- remove useless ESLint comment
- rename defaultProps parameter it can be confusing to see `hit` instead of `props`
- use functional component instead of class

### DIFF
--- a/packages/react-instantsearch/src/components/Hits.js
+++ b/packages/react-instantsearch/src/components/Hits.js
@@ -20,7 +20,6 @@ Hits.propTypes = {
   hitComponent: PropTypes.func.isRequired,
 };
 
-/* eslint-disable react/display-name */
 Hits.defaultProps = {
   hitComponent: hit => (
     <div
@@ -34,6 +33,5 @@ Hits.defaultProps = {
     </div>
   ),
 };
-/* eslint-enable react/display-name */
 
 export default Hits;
--- a/packages/react-instantsearch/src/components/Hits.js
+++ b/packages/react-instantsearch/src/components/Hits.js
@@ -21,7 +21,7 @@ Hits.propTypes = {
 };
 
 Hits.defaultProps = {
-  hitComponent: hit => (
+  hitComponent: props => (
     <div
       style={{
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch/src/components/Hits.js
```diff
@@ -1,39 +1,32 @@
...
 
 Hits.propTypes = {
   hits: PropTypes.array,
   hitComponent: PropTypes.func.isRequired,
 };
 
-/* eslint-disable react/display-name */
 Hits.defaultProps = {
-  hitComponent: hit => (
+  hitComponent: props => (
```
Haroenv: It's more the `hit` which is passed into the single hit, not one of the props of Hits, right?

samouss: Actually `hit` is one of the props pass to the `HitComponent`. Maybe destruct the props will be easier to understand:

```js
Hits.defaultProps = {
  hitComponent: ({ hit }) => (
    ...
  ),
}
```

Haroenv: yes, what we really want to display is the destructured, so we should stringify that 👍

samouss: At the end maybe it's better to stringify all the props pass to the `HitComponent`. When you render the `Hits` component without the custom component we want to see all the props passed to it. Can be confusing to stringify only the `hit`, the user may think that the object displayed are the props. In fact it's only the `hit` property on the props. WDYT?

> https://www.argos-ci.com/algolia/react-instantsearch/builds/15116

Haroenv: sure, makes sense. Personally I'd like it more if we just passed the attributes as props instead of the hit, but that would be a breaking change. Thanks for taking a look at this. Seems like keeping it non-destructured is the good call here indeed

samouss: Completely agree! I will create an issue as a reminder for a later release with BC. Thanks!

PR #660, opened 2017-11-27
fix: track all index in the manager

### LINKED ISSUE #636: In a multi-index configuration, SortBy doesn't sort the main index once inside an Index component

a bug

In a multi-index configuration, you can't sort the main index (the one specified in the ``InstantSearch`` component) once inside an ``Index`` component specifying the same index name. In this fiddle, you can sort the second index but not the first one : https://jsfiddle.net/ar80rgpc/31/

The ``SortBy`` component should sort the first index too.

all

npm@latest : 4.2.0

samouss:
**Summary**

Fix #636 

We already track the index for the `shared` parameters, but not for the `main` parameters. This PR aims to fix this issue by tracking all the indices.

**Result**

You can play with the example on [Storybook](https://deploy-preview-660--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=%3CIndex%3E&selectedStory=with%20SortBy%20nested%20in%20same%20Index%20as%20Root&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs).

### DIFF
--- a/packages/react-instantsearch/src/core/createInstantSearchManager.derived.test.js
+++ b/packages/react-instantsearch/src/core/createInstantSearchManager.derived.test.js
@@ -120,7 +120,82 @@ describe('createInstantSearchManager', () => {
           });
         });
       });
+
+      it('updates the store and searches with duplicate Index & SortBy', () => {
+        // <InstantSearch indexName="first">
+        //   <SearchBox defaultRefinement="query" />
+        //
+        //   <Index indexName="first">
+        //     <SortBy defaultRefinement="third" />
+        //   </Index>
+        //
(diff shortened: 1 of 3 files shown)


mthuret:
LGTM! 
I'm wondering if we shouldn't introduce sublevel into the storybook. It's convenient to add stories to demonstrate those tricky use cases, but imo they do not have the same importance as stories such as "Autocomplete" here. Wdyt?

samouss:
Yep why not! Because this story is more related to a test case rather than documentation. It makes sense to me.

PR #665, opened 2017-11-29
fix: reset page with multi index

### LINKED ISSUE #637: With multiple indices, Configure prevents InfiniteHits to load more results on secondary indices

a bug

With multiple indices, Configure prevents InfiniteHits to load more results on secondary indices. On the given fiddle, you can load more results for the main index but not for the one in an ``Index`` component :
 https://jsfiddle.net/4w04hrud/1/

The bug exists only since version 4.1.1, 4.1.0 works fine.

Be able to load more results on all indices.

all

npm@latest : 4.2.0

samouss:
**Summary**

Fix #637 

With the multi indices we always reset the page when a shared widget is updated even when `resetPage` property is not passed. This PR aims to fix this, now we reset the page only the property is provided.

**Result**

We can play the example on [Storybook](https://deploy-preview-665--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=%3CIndex%3E&selectedStory=with%20Hits%20%26%20Configure&full=0&down=1&left=1&panelRight=1&downPanel=kadira%2Fjsx%2Fpanel).

### DIFF
--- a/packages/react-instantsearch/src/core/indexUtils.js
+++ b/packages/react-instantsearch/src/core/indexUtils.js
@@ -39,11 +39,12 @@ export function refineValue(
         )
       : refineMultiIndex(searchState, nextRefinement, context, resetPage);
   } else {
-    /* 
-      If we have a multi index page with shared widgets we should also reset their page to 1
-      see: https://github.com/algolia/react-instantsearch/issues/310
-    */
-    if (searchState.indices) {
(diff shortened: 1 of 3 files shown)


Haroenv:
Looks good

samouss:
@mthuret any blockers about it?

PR #672, opened 2017-12-01
Feature/instantsearch.css

ronanlevesque:
This PR will include filenames or docs changes at the very end for the sake of keeping it readable. Those changes will come once the PR has been validated.

### Most noticeable changes:

- Added a `BaseWidget` widget which acts like a wrapper for most widgets. It will add header, body and footer to them.
- The `classNames` helper has been rewritten. Instead of returning an objet, it returns a string
- ~~The way `CurrentRefinements` works has been changed to match the behaviour of the one in IS.js. Instead of displaying `category: One ✕ Two ✕` it will display `One ✕` `Two ✕`.~~

### Roadblocks/questions:

- HierarchicalMenu has a few issues:
  - it doesn't give the depth of nested lists
  - nested items won't be present inside the DOM until clicking their parent, thus making it impossible to correctly style an element which has children VS an element which doesn't.
- RangeSlider is different from our other widgets because it relies on [rheostat](https://github.com/airbnb/rheostat). It would be great to make it a fully cross-libs widgets which will offer the same view and CSS classes. That would require to create `vue-rheostat` and `angular-rheostat` though.
- Is it ok to use [Fragments](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html)?

### TODO list

- [x] Update widgets DOM
- [x] Test widgets rendering
- [x] Write/update Jest tests
- [x] Remove Panel
- [x] Update widgets names everywhere (filenames, etc.)
- [x] Update docs:
  - [x] classNames
  - [x] migration guide
  - [x] styling guide


Haroenv:
We support older versions of React, so probably shouldn't be using fragments just yet IMO

ronanlevesque:
@Haroenv thanks, I guess I can't use arrays to return JSX elements either then?

Haroenv:
Best not to be safe. By the way, you should update the snapshots: `yarn jest -u`

packages/react-instantsearch/src/components/CurrentRefinements.js
```diff
@@ -32,35 +33,36 @@ class CurrentRefinements extends Component {
...
-                    <button
-                      {...cx('itemClear')}
-                      onClick={() => refine(nestedItem.value)}
-                    >
-                      {translate('clearFilter', nestedItem)}
-                    </button>
-                  </div>
-                ))
-              ) : (
-                <button {...cx('itemClear')} onClick={() => refine(item.value)}>
```
samouss: We forgot this part in the refactoring.

packages/react-instantsearch/src/components/Hits.js
```diff
@@ -1,33 +1,40 @@
...
-const Hits = ({ hits, hitComponent: HitComponent }) => (
+const Hits = ({ hits, hitComponent: HitComponent, header, footer }) => (
   // Spread the hit on HitComponent instead of passing the full object. BC.
   // ex: <HitComponent {...hit} key={hit.objectID} />
-  <div {...cx('root')}>
-    {hits.map(hit => <HitComponent key={hit.objectID} hit={hit} />)}
-  </div>
+  <BaseWidget widgetClassName={widgetClassName} header={header} footer={footer}>
+    <ul {...cx(['list'])}>
+      {hits.map(hit => <HitComponent key={hit.objectID} hit={hit} />)}
```
samouss: Maybe we can wrap the `HitComponent` in a `li` because otherwise the user will always need to render his component in a `li` in order to have a valid DOM. We will also be able to add the `className` item on it.

packages/react-instantsearch/src/components/InfiniteHits.js
```diff
@@ -14,25 +16,31 @@ class InfiniteHits extends Component {
...
       translate,
+      header,
+      footer,
     } = this.props;
     const renderedHits = hits.map(hit => (
       <ItemComponent key={hit.objectID} hit={hit} />
     ));
     const loadMoreButton = hasMore ? (
-      <button {...cx('loadMore')} onClick={() => refine()}>
+      <button {...cx(['loadMore'])} onClick={() => refine()}>
```
samouss: We can avoid the condition, when a `button` is disabled you can't click on it. We can also inline the `button` inside the `return`.

packages/react-instantsearch/src/components/List.js
```diff
@@ -145,7 +144,9 @@ class List extends Component {
     const searchBox = withSearchBox ? this.renderSearchBox() : null;
     if (items.length === 0) {
       return (
-        <div {...cx('root', !canRefine && 'noRefinement')}>{searchBox}</div>
+        <ul {...cx(['list', !canRefine && 'list--noRefinement'])}>
+          {searchBox}
```
samouss: The wrapping element of `searchBox` is `div` instead of `li`.

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -186,70 +204,51 @@ class SearchBox extends Component {
...
+        >
           <input
             ref={this.onInputMount}
-            type="search"
+            type="text"
             placeholder={translate('placeholder')}
             autoFocus={autoFocus}
             autoComplete="off"
-            autoCorrect="off"
-            autoCapitalize="off"
```
samouss: Sure about those changes?

packages/react-instantsearch/src/components/StarRating.js
```diff
@@ -161,7 +168,24 @@ class StarRating extends Component {
...
     }
-    return <div {...cx('root', !canRefine && 'noRefinement')}>{items}</div>;
+    return (
+      <BaseWidget
+        widgetClassName={widgetClassName}
+        otherWidgetClassNames={[!canRefine && `-noRefinement`]}
+        header={header}
+        footer={footer}
+      >
+        <svg xmlns="http://www.w3.org/2000/svg" style={{ display: 'none' }}>
```
samouss: Maybe we can simply put this markup into a constant variable, then using it for pushing the stars in the array. Rather than use the `use` syntax. Don't have strong opinions about that, WDYT?

packages/react-instantsearch/src/components/classNames.js
```diff
@@ -2,11 +2,19 @@ import cx from 'classnames';
 const prefix = 'ais';
 
 export default function classNames(block) {
-  return (...elements) => ({
-    className: cx(
-      elements
-        .filter(element => element !== undefined && element !== false)
-        .map(element => `${prefix}-${block}__${element}`)
-    ),
+  return (bemElements, otherElements) => ({
```
samouss: I am not a huge fan of the fact that `classNames` function return an object. IMO it will be much easier to return a string and apply this string to the className property. With this kind of return type we can't use it with outside of the `jsx` scope. We can't easily do something like:

```js
const myClassName = cx(['myClass', isTrue && 'myOtherClass']); // ais-Block-myClass ais-Block-myOtherClass
```

vvo:
> RangeSlider is different from our other widgets because it relies on rheostat. It would be great to make it a fully cross-libs widgets which will offer the same view and CSS classes. That would require to create vue-rheostat and angular-rheostat though.

I think the community would benefit from that yes. I am "tired" of having to choose the right slider and find a different API/features. Rheostat is great for us for our own usage, would be nice to have those packages. And it would be a great JavaScript community addition. It would allow us to easily have the slider styled with InstantSearch.css. Along with avoiding to have custom styling inlined in the library like we have with Angular InstantSearch.

Angular InstantSearch uses IS.css but has a different slider. Thus we bundle IS.css and add more styles to it.

Ultimately we should never bundle IS.css inside JS libs.

> HierarchicalMenu has a few issues:
> it doesn't give the depth of nested lists
> nested items won't be present inside the DOM until clicking their parent, thus making it impossible to correctly style an element which has children VS an element which doesn't.

Just to summarize it a bit. We want to style hierarchical menu in a way that displays an arrow that is different if the element has or has not sub elements. But we only know about those subelements when we click on the facet. This would require changes in the helper.

> Is it ok to use Fragments?

We discussed it briefly with @samouss, a good requirement for ris should be to be compatible with the two last versions of React, always. But when one version last update (15.x.x) was more than one year old then we can drop its support if there's a very good feature.

packages/react-instantsearch/src/components/StarRating.js
vvo: Also possible to use https://github.com/smooth-code/svgr and have real svg images (need to build the rollup plugin)

vvo:
> I have a comment on the usage of BaseWidget. I'm not a huge fan of using it inside the components file.

Indeed, component should stay as simple as possible and "plumbing" can be done inside widgets/.

But maybe that's not doable, you can dig that while reviewing together.

vvo:
Impressed by the work you put in here @ronanlevesque! Keep the momentum and review with @samouss side by side to iterate!

packages/react-instantsearch/src/components/SearchBox.js
ronanlevesque: Yup. Those attributes are not w3c valid.

packages/react-instantsearch/src/components/StarRating.js
ronanlevesque: I'd rather stick to `use` as it is the recommended way to deal with repeating SVGs.

packages/react-instantsearch/src/components/CurrentRefinements.js
ronanlevesque: This part wouldn't be necessary according to the changes I've suggested on the PR desc. Happy to discuss it if needed!

samouss: Wasn't sure about that, works for me.

packages/react-instantsearch/src/components/SearchBox.js
ronanlevesque: As discussed with @Haroenv we will keep it because it's still convenient for our users. (eg avoids annoying behaviours on Safari)

packages/react-instantsearch/src/components/Panel.js
```diff
@@ -32,8 +32,8 @@ class Panel extends Component {
 
   render() {
     return (
-      <div {...cx('root', !this.state.canRefine && 'noRefinement')}>
-        <h5 {...cx('title')}>{this.props.title}</h5>
+      <div {...cx(['', !this.state.canRefine && '-noRefinement'])}>
+        <h5 {...cx(['title'])}>{this.props.title}</h5>
```
samouss: You can use a list of parameters instead of an array.

packages/react-instantsearch/src/components/ScrollTo.js
```diff
@@ -20,7 +20,7 @@ class ScrollTo extends Component {
 
   render() {
     return (
-      <div ref={ref => (this.el = ref)} {...cx('root')}>
+      <div ref={ref => (this.el = ref)} {...cx([''])}>
```
samouss: You can use an empty string.

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -201,14 +203,43 @@ class SearchBox extends Component {
...
+        className={cx('resetIcon')}
+        xmlns="http://www.w3.org/2000/svg"
+        viewBox="0 0 20 20"
+        width="10"
+        height="10"
+      >
+        <path d="M8.114 10L.944 2.83 0 1.885 1.886 0l.943.943L10 8.113l7.17-7.17.944-.943L20 1.886l-.943.943-7.17 7.17 7.17 7.17.943.944L18.114 20l-.943-.943-7.17-7.17-7.17 7.17-.944.943L0 18.114l.943-.943L8.113 10z" />
+      </svg>
+    );
+
```
samouss: We should move the default markup on the default components that are at the top of the file. Then we can reuse the default props and avoid the useless condition.

packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -230,81 +261,55 @@ class SearchBox extends Component {
...
+          }}
+        >
+          {submitComponent}
+        </button>
+        <button
+          type="reset"
+          title={translate('resetTitle')}
+          className={cx('reset')}
+          onClick={this.onReset}
+          style={query ? { display: 'block' } : { display: 'none' }}
```
samouss: Prefer the ternary inside the value instead on the all object.

packages/react-instantsearch/src/widgets/BaseWidget.js
```diff
@@ -0,0 +1,26 @@
...
+
+    return (
+      <div className={cx('', cantRefine && `-noRefinement`)}>
+        {header && <div className={`${cx('header')} ais-header`}>{header}</div>}
+        <div className={`${cx('body')} ais-body`}>{children}</div>
+        {footer && <div className={`${cx('footer')} ais-footer`}>{footer}</div>}
+      </div>
+    );
+  }
+}
```
samouss: We don't need a `class` we can move for a functional component.

packages/react-instantsearch/src/widgets/Breadcrumb.js
```diff
@@ -82,4 +87,16 @@ import BreadcrumbComponent from '../components/Breadcrumb.js';
  *   );
  * }
  */
-export default connectBreadcrumb(BreadcrumbComponent);
+
+const Widget = props => (
```
samouss: We should destruct the `props` in order to extract `header`, `footer` to avoid to pass them down the three. We should also add `PropTypes` on the `Widget` component.

packages/react-instantsearch/src/components/Highlight.js
```diff
@@ -3,7 +3,13 @@ import React from 'react';
 import Highlighter from './Highlighter';
 
 export default function Highlight(props) {
-  return <Highlighter highlightProperty="_highlightResult" {...props} />;
+  return (
+    <Highlighter
+      highlightProperty="_highlightResult"
+      widgetClassName="Highlight"
```
samouss: We can adopt the same strategy as on the other widgets. We could bind the `classNames` from the wdiget and then pass `cx` down the tree (`Widget` > `Hightlight` > `Hightlighter`). We can also do it for the `Snippet` widget.

packages/react-instantsearch/src/components/classNames.js
```diff
@@ -2,11 +2,18 @@ import cx from 'classnames';
...
+          element =>
+            element !== undefined && element !== false && element !== null
+        )
+        .map(
+          element =>
+            element === ''
+              ? `${prefix}-${block}`
+              : `${prefix}-${block}-${element}`
+        )
+    );
```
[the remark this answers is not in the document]
ronanlevesque: Not sure about what you mean in your example. We need to be able to either return `ais-Widget` or `ais-Widget-element`. With your example it would just return `-element` or an empty string?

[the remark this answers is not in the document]
samouss: Nope it will have the same behaviour, if the element is defined then it will return `ais-Widget-element` otherwise it returns `ais-Widget`. Keep as it is if you think it's simpler.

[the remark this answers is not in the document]
ronanlevesque: Slightly different behaviour (the class is doubled for some reason). Let's keep it this way for now if you don't mind.

vvo:
Do you need another review @samouss?

samouss:
For now, it should be good. I make PR on every new changes that a made so normally it should be good 🙂

samouss:
Close in favour of #924

PR #683, opened 2017-12-04
docs(snippet): import correctly

Haroenv:
**Summary**

imports from the wrong package for custom hits

**Result**

import from connectors, then refactored not to use connectors

### DIFF
--- a/packages/react-instantsearch/src/widgets/Snippet.js
+++ b/packages/react-instantsearch/src/widgets/Snippet.js
@@ -16,7 +16,8 @@ import SnippetComponent from '../components/Snippet.js';
  * @example
  * import React from 'react';
  *
- * import { connectHits, Snippet, InstantSearch } from 'react-instantsearch/dom';
+ * import { Snippet, InstantSearch } from 'react-instantsearch/dom';
+ * import { connectHits } from 'react-instantsearch/connectors';
  *
  * const CustomHits = connectHits(({ hits }) =>
  * <div>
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch/src/widgets/Snippet.js
```diff
@@ -15,29 +15,24 @@ import SnippetComponent from '../components/Snippet.js';
...
- *    <InstantSearch
+ *   return (
+ *     <InstantSearch
  *       appId="latency"
  *       apiKey="6be0576ff61c053d5f9a3225e2a90f76"
  *       indexName="ikea"
  *     >
- *       <CustomHits />
+ *       <Hits
+ *         hitComponent={hit => (
```
samouss: The parameter hit are the props of the component, so you can't access it directly. You should use destructuring:

```jsx
<Hits
  hitComponent={({ hit }) => (
    ...
  )}
/>
```

packages/react-instantsearch/src/widgets/Snippet.js
```diff
@@ -15,29 +15,24 @@ import SnippetComponent from '../components/Snippet.js';
...
- * <div>
- *   {hits.map(hit =>
- *     <p key={hit.objectID}>
- *       <Snippet attributeName="description" hit={hit} />
- *     </p>
- *   )}
- * </div>
- * );
- *
+ * import { Snippet, InstantSearch, Hits } from "react-instantsearch/dom";
```
samouss: Use single quote as the reset of the example.

Haroenv: my prettier settings were on default

samouss:
Prettier is still not happy... See the CI log.

PR #714, opened 2017-12-12
fix(SearchBox): use hidden over style to hide loader

bobylito:
This way the browser is better able to report the state of the component for accessibility.


samouss:
@bobylito You forgot to update the Snapshot 🙂

bobylito:
Good point thanks 👍

PR #715, opened 2017-12-12
feat(Highlight): support array of strings

### LINKED ISSUE #22: Highlight: support arrays
# Use case

Your records contain a field that is an array of strings and you want to highlight them. 

# Need

The `<Highlight/>` and `<Snippet/>` widgets are not compatible with an array of strings. 

# Proposal

TBD

See https://github.com/algolia/instantsearch.js/issues/1932

marielaures:
**Summary**

This PR fixes #22.
Before this, the Highlight widget was only compatible with **strings**. 
Now we've added support for **arrays of strings**.

**Result**

The `Highlight` widget is used in the same way for attribute names that point to either a string or an array of strings. 

In the following example we have two Highlight widgets, the first one for `attributeName="name"` where `hit.name` is a string, and the second one for `attributeName="tags"` where `hit.tags` is an array of strings (`tags: ["Ancient Egypt", "Online Course", "History"]` ):

```
<Highlight attributeName="name" hit={hit} />
<Highlight attributeName="tags" hit={hit} />
```

highlightarray


docgen/plugins/rev-assets.js
```diff
@@ -57,10 +57,9 @@ function renameReferences(
   return replace({ from, to, files: inFiles });
 }
 
-function computeFiles([filesGlob, replaceFilesGlob]: Array<
-  string,
-  string
->): Promise<*> {
+function computeFiles(
+  [filesGlob, replaceFilesGlob]: Array<string, string>
```
Haroenv: seems like a different version of prettier was run over the code base

packages/react-instantsearch/src/components/Highlighter.js
```diff
@@ -38,4 +78,12 @@ Highlighter.propTypes = {
...
   highlightProperty: PropTypes.string.isRequired,
   tagName: PropTypes.string,
+  nonHighlightedTagName: PropTypes.string,
+  separator: PropTypes.node,
+};
+
+Highlighter.defaultProps = {
+  tagName: 'em',
+  nonHighlightedTagName: 'span',
+  separator: ',',
```
Haroenv: default should be `, `

packages/react-instantsearch/src/components/Highlighter.js
```diff
@@ -1,35 +1,75 @@
...
+  return (
+    <span className="ais-Highlight">
+      {parsedHighlightedValue.map((item, i) => {
+        if (Array.isArray(item)) {
+          const isLast = i === parsedHighlightedValue.length - 1;
+          return (
+            <span key={`split-${i}-${hit[attributeName][i]}`}>
+              {item.map((element, index) => (
+                <Highlight
+                  key={`split-${index}-${element.value}`}
```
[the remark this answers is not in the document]
marielaures: The key generation is slightly different, since in one case we're accessing an element of an array (`hit[attributeName][i]`) and for the case where `attributeName` is a string, we do `element.value`. Do I do a function a maybe pass an attribute which will be a Boolean to specify whether it is an array and return either the first option or the second or leave it as is?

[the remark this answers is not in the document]
samouss: Nope, you can simply create function that takes `(index, value)` as parameters. Then you called it with the according values (ex: `element.value`, `hit[attributeName][i]`).

packages/react-instantsearch/src/components/Highlighter.js
```diff
@@ -1,35 +1,79 @@
...
 import React from 'react';
+import classNames from './classNames';
+
+const cx = classNames('Highlight');
+
+function generateKey(i, value) {
+  return `split-${i}-${value}`;
+}
+
+export const Highlight = ({
```
vvo: Should we export this as a component and then test the rendering out of it? So we freeze the implementation. Right now we are only testing that Highlight is called with some props, not the actual rendering of it (because it's shallow rendered in Highlighter.test.js)

samouss: What do you mean by freezing the implementation? `Highlight` is tested separately so we don't need to do a full DOM rendering in `Highlighter`. We can just test that the the correct props are applied on the component. I think it's easier to reason about unit, rather than the all implementation.

vvo: What I meant is that right no there's no test ensuring that the output of this single Highlight component is some specific DOM.

Which means that if we change the output without even touching the props being passed, nothing will break, while the public API contract will break (DOM will change).

If that's not clear still let me know

samouss: Nope perfectly clear! But we already test the output of the `Highlight` component (see [#L11-39](https://github.com/algolia/react-instantsearch/blob/f3522ade97b94a2ee0c82590c02fc6c648be267d/packages/react-instantsearch/src/components/Highlighter.test.js#L11-L39))

vvo: 👍

PR #619, opened 2017-11-17
feat(refreshcache): add prop refresh to InstantSearch instance

marielaures:
**Summary**

The goal of this PR is to make it possible to refresh an InstantSearch instance.
It is allows the user to make a new request to Algolia in case a change has happened on the index for instance instead of getting the results from the cache.

Fixes #252. 

**Result**

Adding a prop `refresh` to the `InstantSearch` component will trigger the refresh. 
It takes a Boolean as an argument.

```
  <InstantSearch
        appId="xyz"
        apiKey="xyz"
        indexName="indexName"
        refresh={this.state.refresh}
  >
```

You can see how it works on the refreshCache story in Storybook. 

demo

PR #732, opened 2017-12-18
docs(refreshcache): add guide

marielaures:
**Summary**

In the wake of the refresh cache feature (#619), this is a guide on how to refresh the cache with the description of the use case and an explanation of the two different approaches that can be used.

### DIFF
--- a/docgen/src/guide/Handling_cache.md
+++ b/docgen/src/guide/Handling_cache.md
@@ -1,0 +1,113 @@
+---
+title: Handling cache
+mainTitle: Guides
+layout: main.pug
+category: guide
+navWeight: 15
+---
+
+
+By default, Algolia caches the search results of the queries, storing them locally in the cache. If the user ends up entering a search (or part of it) that has already been entered previously, the results will be retrieved from the cache, instead of requesting them from Algolia, making the application much faster.
+
+While it is a very convenient feature, in some cases, it is useful to have the ability to clear the cache and make a new request to Algolia. For instance, when changes are made on some records on your index, you might want the frontend side of your application to be updated to reflect that change (in order to avoid displaying stale results retrieved from the cache).
+
+To do so, there is a prop on the `<InstantSearch>` component called `refresh` that, when set to true, clears the cache and triggers a new search.
+
+There are two different approaches to handle the cache:
+ - have the refresh of the cache triggered by an action from the user 
+ - refresh the cache periodically
+ 
+ ## Have the refresh of the cache triggered by an action from the user
+If you know that the cache needs to be refreshed conditionally of a specific event, then you can trigger the refresh based on an user action (adding a new product, clicking on a button for instance).
+
+```jsx
+import React, { Component } from 'react';
+import { InstantSearch, SearchBox } from '../packages/react-instantsearch/dom';
+
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      refresh: false,
+    };
+  }
+
+  refresh = () => {
+    this.setState(prevState => ({
+      refresh: !prevState.refresh,
+    }));
+  };
+
+  onSearchStateChange = () => {
+    this.setState({ refresh: false });
+  };
+
+  render() {
+    return (
+      <InstantSearch
+        appId="yourAppId"
+        apiKey="yourApiKey"
+        indexName="yourIndexName"
+        refresh={this.state.refresh}
+        onSearchStateChange={this.onSearchStateChange}
+      >
+        <SearchBox />
+        <button
+          onClick={this.refresh}
+        >
+          Refresh cache
+        </button>
+        <CustomHits />
+      </InstantSearch>
+    );
+  }
+}
+```
+
+See the example in this [story](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RefreshCache&selectedStory=with%20a%20refresh%20button&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs). 
+
+ ## Refresh the cache periodically
+You also have the option to setup an given period of time that will determine how often the cache will be cleared. 
+This method will ensure that the cache is cleared on a regular basis.
+You should use this approach if you cannot use a user action as a specific event to trigger the clearing of the cache.
+
+```jsx
+import React, { Component } from 'react';
+import { InstantSearch, SearchBox } from '../packages/react-instantsearch/dom';
+
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      refresh: false,
+    };
+  }
+
+  componentDidMount = () => {
+    setInterval(
+      () =>
+        this.setState({ refresh: true }, () =>
+          this.setState({ refresh: false })
+        ),
+      5000
+    );
+  };
+
+  render() {
+    return (
+      <InstantSearch
+        appId="yourAppId"
+        apiKey="yourApiKey"
+        indexName="yourIndexName"
+        refresh={this.state.refresh}
+        onSearchStateChange={this.onSearchStateChange}
+      >
+        <SearchBox />
+        <CustomHits />
+      </InstantSearch>
+    );
+  }
+}
+```
+
+See the example in this [story](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RefreshCache&selectedStory=with%20setInterval&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs).
--- a/docgen/src/guide/Refreshing_cache.md
+++ b/docgen/src/guide/Refreshing_cache.md
@@ -1,5 +1,5 @@
 ---
-title: Handling cache
+title: Refreshing the cache
 mainTitle: Guides
 layout: main.pug
 category: guide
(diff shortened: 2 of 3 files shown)


docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+There are two different approaches to handle the cache:
+ - have the refresh of the cache triggered by an action from the user 
+ - refresh the cache periodically
+ 
+ ## Have the refresh of the cache triggered by an action from the user
+If you know that the cache needs to be refreshed conditionally of a specific event, then you can trigger the refresh based on an user action (adding a new product, clicking on a button for instance).
+
+```jsx
+import React, { Component } from 'react';
+import { InstantSearch, SearchBox } from '../packages/react-instantsearch/dom';
```
Haroenv: `... from 'react-instantsearch/dom'`

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+See the example in this [story](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RefreshCache&selectedStory=with%20a%20refresh%20button&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs). 
+
+ ## Refresh the cache periodically
+You also have the option to setup an given period of time that will determine how often the cache will be cleared. 
+This method will ensure that the cache is cleared on a regular basis.
+You should use this approach if you cannot use a user action as a specific event to trigger the clearing of the cache.
+
+```jsx
+import React, { Component } from 'react';
+import { InstantSearch, SearchBox } from '../packages/react-instantsearch/dom';
```
Haroenv: `react-instantsearch/dom`

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+        <CustomHits />
+      </InstantSearch>
+    );
+  }
+}
+```
+
+See the example in this [story](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RefreshCache&selectedStory=with%20a%20refresh%20button&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs). 
+
+ ## Refresh the cache periodically
```
Haroenv: This has a leading space

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+      >
+        <SearchBox />
+        <CustomHits />
+      </InstantSearch>
+    );
+  }
+}
+```
+
+See the example in this [story](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RefreshCache&selectedStory=with%20setInterval&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs).
```
Haroenv: Would be nice to have an example that explains

1. you do some action
2. index backend
3. `waitTask`
4. send some event to the frontend (maybe web sockets)
5. refresh

but it might be too long, because it has quite some setup with the sockets

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+category: guide
+navWeight: 15
+---
+
+
+By default, Algolia caches the search results of the queries, storing them locally in the cache. If the user ends up entering a search (or part of it) that has already been entered previously, the results will be retrieved from the cache, instead of requesting them from Algolia, making the application much faster.
+
+While it is a very convenient feature, in some cases, it is useful to have the ability to clear the cache and make a new request to Algolia. For instance, when changes are made on some records on your index, you might want the frontend side of your application to be updated to reflect that change (in order to avoid displaying stale results retrieved from the cache).
+
+To do so, there is a prop on the `<InstantSearch>` component called `refresh` that, when set to true, clears the cache and triggers a new search.
```
vvo: proposal: `true` instead of true

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+
+There are two different approaches to handle the cache:
```
vvo: Proposal:

> There are two different use cases where you would want to discard the cache:
> - your application data is being updated by another process that you don't manager, for example a CRON job updates users inside Algolia. For this you might want to periodically refresh the cache of your app
> - you application data is being updated by the users of your app directly (for example, in a dashboard). In this use case you would want to refresh the cache base on some application state like the last modification of the user

wdyt?

I put this to better explains the scenarios.

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+          Refresh cache
+        </button>
+        <CustomHits />
+      </InstantSearch>
+    );
+  }
+}
+```
+
+See the example in this [story](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RefreshCache&selectedStory=with%20a%20refresh%20button&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs).
```
vvo: > See a live example in storybook

docgen/src/guide/Handling_cache.md
vvo: > See a live example in storybook

docgen/src/guide/Handling_cache.md
Haroenv: > that you don’t manage

docgen/src/guide/Handling_cache.md
samouss: Yes you are right it would be nice, but it's a bit harder to summarise. We point users to the right direction, they have the basics for using the update cache feature. IMO, how to receive the event for trigger the refresh is an implementation detail.

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
...
+    this.state = {
+      refresh: false,
+    };
+  }
+
+  componentDidMount = () => {
+    setInterval(
+      () =>
+        this.setState({ refresh: true }, () =>
+          this.setState({ refresh: false })
```
samouss: It can be confusing for the user to use two different approach for reset the `refresh` property.

stories/RefreshCache.stories.js
```diff
@@ -128,3 +129,74 @@ class AppWithRefresh extends Component {
...
+  };
+
+  componentDidMount = () => {
+    setInterval(
+      () =>
+        this.setState(prevState => ({
+          count: prevState.count + 1,
+        })),
+      1000
+    );
```
samouss: We don't need to set two intervals, only one should be sufficient. Set an interval of 1000 and reset it when it reach the desired value. Then you can get rid of this component I think. Try to also use the same reset strategy as explained above.

stories/RefreshCache.stories.js
```diff
@@ -128,3 +129,74 @@ class AppWithRefresh extends Component {
...
+
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      refresh: false,
+    };
+  }
+
+  componentDidMount = () => {
```
samouss: Don't need to use a class property, the regular method should work `componentDidMount() { ... }`

docgen/src/guide/Handling_cache.md
marielaures: Yep, that space is here on purpose to get the heading format (see other guides)

docgen/src/guide/Handling_cache.md
Haroenv: It might still be nice to mention that if they are waiting for an Algolia action, they'll need to use `waitTask` and send _some_ event, otherwise the refreshing will be too early

docgen/src/guide/Handling_cache.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Handling cache
```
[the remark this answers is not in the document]
Haroenv: >Refreshing the cache

also sounds nice to me

docgen/src/guide/Handling_cache.md
marielaures: What would you suggest here? Did you mean just to change the call to `refresh` on the previous example to:
```
this.setState({ refresh: true });
```
?

docgen/src/guide/Refreshing_cache.md
```diff
@@ -0,0 +1,115 @@
...
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      refresh: false,
+    };
+  }
+
+  refresh = () => {
+    this.setState({refresh: true});
```
samouss: Be careful with the code style. Prettier seems to nut run on the markdown in your editor. But you can still copy / paste the snippet inside the [Prettier playground](https://prettier.io/playground). Should be:

```js
({ refresh: true })
```

docgen/src/guide/Refreshing_cache.md
```diff
@@ -0,0 +1,115 @@
...
+        refresh={this.state.refresh}
+        onSearchStateChange={this.onSearchStateChange}
+      >
+        <SearchBox />
+        <button
+          onClick={this.refresh}
+        >
+          Refresh cache
+        </button>
+        <CustomHits />
```
samouss: This variable doesn't exist in the snippet, use `Hits` instead of `CustomHits`. Don't forget to add the import.

docgen/src/guide/Refreshing_cache.md
```diff
@@ -0,0 +1,115 @@
...
+
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      refresh: false,
+    };
+  }
+
+  componentDidMount = () => {
```
samouss: Prefer the usage of method instead of class property.

```js
componentDidMount() { ... }
```

docgen/src/guide/Refreshing_cache.md
```diff
@@ -0,0 +1,115 @@
...
+    setInterval(
+      () =>
+        this.setState({ refresh: true }), 5000 );
+  };
+
+   onSearchStateChange = () => {
+    this.setState({ refresh: false });
+  };
```
samouss: Add a space after this declaration.

docgen/src/guide/Refreshing_cache.md
```diff
@@ -0,0 +1,115 @@
...
+    return (
+      <InstantSearch
+        appId="yourAppId"
+        apiKey="yourApiKey"
+        indexName="yourIndexName"
+        refresh={this.state.refresh}
+        onSearchStateChange={this.onSearchStateChange}
+      >
+        <SearchBox />
+        <CustomHits />
```
samouss: This variable doesn't exist in the snippet, use `Hits` instead of `CustomHits`. Don't forget to add the import.

docgen/src/guide/Refreshing_cache.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+See a live example on our [Storybook](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=RefreshCache&selectedStory=with%20a%20refresh%20button&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs). 
+
+ ## Refresh the cache periodically
+You also have the option to setup an given period of time that will determine how often the cache will be cleared. 
+This method will ensure that the cache is cleared on a regular basis.
+You should use this approach if you cannot use a user action as a specific event to trigger the clearing of the cache.
+
+```jsx
```
samouss: Same comment as above, use [Prettier](https://prettier.io/playground) for the all the snippet.

docgen/src/guide/Refreshing_cache.md
Haroenv: You should be able to put the whole markdown in the playground, but we probably should have a lint-staged command that runs prettier on all staged files, WDYT?

PR #739, opened 2017-12-19
chore(size): export <Index /> from its own widget

Haroenv:
This makes /dom, /react-native "pure modules", which _might_ have advantages in webpack 4 cc @thelarkinn

We'll also need `side-effects: false` in our package.json, see [example](https://github.com/webpack/webpack/tree/next/examples/side-effects), but I'm not sure if we're _actually_ side-effects RN, could you confirm Sean?


Haroenv:
this is actually .2kb smaller gzipped because the code is more similar

samouss:
Not sure about the timings for those changes, the all repo / file structure will be updated in the coming weeks. Do you mind if we postpone those changes at that time?

Haroenv:
Sure, let’s do it later

PR #749, opened 2017-12-20
fix(createInstantSearch): remove the client from the Snapshot

samouss:
**Summary**

With `jest@22` the mock function are now print with the snapshot. 

So when we call `createInstantSearch` from the test, we provide a mock client with the current version of the lib. We already test that the instance is called with the correct client (see below). So we don't need to print the client mock in the snapshot. It will avoid that every time we release a new version we have to fix the snapshot, just because the version bump.

> https://github.com/algolia/react-instantsearch/pull/749/files#diff-89fda296b36d39c13ca0e69ebdb33163R34

### DIFF
--- a/packages/react-instantsearch/src/core/__snapshots__/createInstantSearch.test.js.snap
+++ b/packages/react-instantsearch/src/core/__snapshots__/createInstantSearch.test.js.snap
@@ -2,15 +2,6 @@
 
 exports[`createInstantSearch wraps InstantSearch 1`] = `
 Object {
-  "algoliaClient": Object {
-    "addAlgoliaAgent": [MockFunction] {
-      "calls": Array [
-        Array [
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch/src/core/__snapshots__/createInstantSearch.test.js.snap
```diff
@@ -2,15 +2,6 @@
 
 exports[`createInstantSearch wraps InstantSearch 1`] = `
 Object {
-  "algoliaClient": Object {
-    "addAlgoliaAgent": [MockFunction] {
-      "calls": Array [
-        Array [
-          "react-instantsearch 4.3.0-beta.0",
```
Haroenv: yep, makes sense to not have the version number in the tests

PR #751, opened 2017-12-20
fix(recipes): update Snapshot

samouss:


samouss:
Close in favour of #749.

PR #770, opened 2017-12-27
feat: InstantSearch root props

### LINKED ISSUE #285: {flex: 1} doesn't work inside <InstantSearch> component
Is there a way to pass `{flex: 1}` to <InstantSearch> component? The only way I found is to change the component code inside the library:

```
{Root: View, props: {flex: 1}}
```

samouss:
**Summary**

Fix #285 

Enable the possibility to pass custom `root` props to InstantSearch. Don't introduce any breaking change, since we use the root of `createInstantSearch` as `defaultProps`.

**Result**

The following example is with React Native, when we don't pass `flex: 1` to InstantSearch the list is crop before the end.

**Before**

simulator screen shot - iphone x - 2017-12-27 at 12 18 14

**After**

simulator screen shot - iphone x - 2017-12-27 at 12 18 35


mike623:
Did it push to NPM? It seems not don't have effect

samouss:
Nope it’s not released yet, it will be published today or tomorrow.

PR #784, opened 2018-01-02
fix: refresh cache memory leak example

samouss:
**Summary**

Clear the interval on component `willUnmount`.

### DIFF
--- a/docgen/src/guide/Refreshing_cache.md
+++ b/docgen/src/guide/Refreshing_cache.md
@@ -82,7 +82,11 @@ class App extends Component {
   }
 
   componentDidMount() {
-    setInterval(() => this.setState({ refresh: true }), 5000);
+    this.interval = setInterval(() => this.setState({ refresh: true }), 5000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.interval);
   }
 
   onSearchStateChange = () => {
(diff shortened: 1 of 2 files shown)


docgen/src/guide/Refreshing_cache.md
```diff
@@ -82,7 +82,11 @@ class App extends Component {
   }
 
   componentDidMount() {
-    setInterval(() => this.setState({ refresh: true }), 5000);
+    this.interval = setInterval(() => this.setState({ refresh: true }), 5000);
```
Haroenv: WDYT of setting line length in markdown files to something like 60? I prefer having shorter lines when reading something in a post, rather than while writing code

samouss: Agree, I'm looking for something like that yesterday :)

Haroenv: I think we'll need two prettier configs for that, not sure if it's possible already 🤔

samouss: I think we can, `printWidth` also handle markdown by default.

> https://prettier.io/docs/en/options.html#print-width

Haroenv: I meant that we can't have `js: { printWidth: 80 }, md: { printWidth: 55 }` IIRC, but would be good to add

samouss: Ah ok. Yep would be great to support this kind of configuration!

PR #786, opened 2018-01-02
docs(react-native): query suggestions demo

mthuret:
Hi :)

I worked on an example demonstrating query suggestions implem with React Native to have some materials to tweet when we'll launch it. 

You can play with it on your phone: https://expo.io/@mlthuret/query-suggestions-rn

I add some trouble only having one InfiniteHits widgets and hiding/showing the suggestions that's why I ended up with the dual Hits/InfiniteHits, but if someone wants to review the code to see if I'm missing something that'd be great :) 

Tell me what you think of it. 

It misses the right metadata before being publish and the screenshot to be on the documentation. Also it needs to be tested on Android (didn't add time). 

- [x]  Icon for the suggestions
- [x]  Check if tap can be done only one in the suggestion list instead of a long press or double tap
- [x]  Close keyboard when the suggestion is selected
- [x]  Readme with instructions
- [x]  When user type something, scrolls, and type again without losing the searchbox focus, the suggestions doesn't appear. 

Some improvement I can think of: 

* Recent search 
* Animation when hiding the suggestions 
* Display categories and pre-filter with them


Haroenv:
Your expo app is private I think, I don't see it on your account 😄

mthuret:
Here the QR code:

image

packages/react-instantsearch/examples/query-suggestions-rn/App.js
```diff
@@ -0,0 +1,251 @@
...
+    const onEndReached = function() {
+      if (hasMore) {
+        refine();
+      }
+    };
+    return (
+      <HitsList
+        removeSuggestions={removeSuggestions}
+        hits={hits}
+        onEndReached={onEndReached}
```
Haroenv: we should probably revert the addition of `react/jsx-no-bind`, since this is clearly correct, even though eslint doesn't like it

cc @vvo @samouss

packages/react-instantsearch/examples/query-suggestions-rn/App.js
```diff
@@ -0,0 +1,251 @@
...
+  refine: PropTypes.func,
+};
+
+const HitsList = ({ hits, removeSuggestions, onEndReached }) => (
+  <FlatList
+    renderItem={({ item }) => (
+      <View style={styles.hitsContainer}>
+        <Image
+          source={{
+            uri: `https://res.cloudinary.com/hilnmyskv/image/fetch/h_300,q_100,f_auto/${item.image}`,
```
Haroenv: Travis tells that you should run `prettier` again

packages/react-instantsearch/examples/query-suggestions-rn/App.js
samouss: If you convert the function to `() => {}` it woks I think (not sure). Or we can simply allow functions in the rule.

> https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#allowfunctions

samouss:
Thanks for the demo @mthuret! Some comments on the UI rather than the code. 

IMO we should avoid to move the results when suggestions appear. We could instead display on overlay when the focus is on the SearchBox and then display the suggestions on top of the results when they are available. Youtube does this kind of behaviour (see screenshot below). Amazon does a similar thing, they have a dedicated page for the suggestions and when you click on a suggestion or when you trigger the search with the keyboard you move to the results page. 

There is a little bug also the list is crop before the end when few results are displayed. You can try to search for "iottie" for example. When you reach the end you still have results but you can't scroll more.

img_0236

img_0235

mthuret:
Here I tried to mimic the Inbox UI that mix suggestions on top and best results after the first keystroke to demonstrate the most complete use case. The only difference is that I choosed to keep displaying the best results of the empty query. 

Thanks for spotting the bug, will try to fix it.

packages/react-instantsearch/examples/query-suggestions-rn/Readme.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Clone the example
+
+```
+curl https://codeload.github.com/algolia/react-instantsearch/tar.gz/master | tar -xz --strip=2 react-instantsearch-master/packages/react-instantsearch/examples/query-suggestions-rn
+```
+
+## Start the example
+
+You need an expo account. [Create an account or signup with github](https://expo.io/signup).
```
samouss: Maybe we should move for `create-react-native-app`, it's a pain to create an account just for run the App locally? WDYT?

mthuret:
@samouss I updated to CRNA and updated the RIS version. I think I'm done in term of fix/new features, feel free to check :)

samouss:
Do you tried to remove the dual `Hits` & `InfiniteHIts`? Seems a bit confusing from the reader perspective.

mthuret:
I added that because having only `connectInfinitHits` was not working. I suspect something going wrong due to the accumulation of hit. Didn't retry lately and I will not have time to do so, but you can investigate if you want.

samouss:
Ok will I will merge it and give it a try in the coming days.

PR #792, opened 2018-01-03
Fix/custom root index

samouss:
**Summary**

In a previous [PR #770](https://github.com/algolia/react-instantsearch/pull/770) we have introduce the `root` prop for the `InstantSearch` component. But we forgot to handle the case for the `Index` component. The test of `createInstantSearch` has also be updated because the previous test didn't tested the `root` prop.

Enable the possibility to pass custom `root` prop to Index. Don't introduce any breaking change, since we use the root of `createIndex` as `defaultProps`.

**Result**

*Waiting the preview...*


Haroenv:
looks good when skimming

PR #797, opened 2018-01-04
fix(connectRange): check if facet exist before access

### LINKED ISSUE #630: Changing the attributeName prop of a ConnectedRange results in failed request
**Do you want to request a *feature* or report a *bug*?** BUG

When changing the value of the `attributeName` prop passed to a mounted component wrapped with `connectRange`, the new request fails with `ATTRIBUTE_NAME_2 is not present in 'facets' or 'disjunctiveFacets'`

Here is a simple reproduction example: https://codesandbox.io/s/ojxzzoky55

To generate the issue, switch the select tag to "Movies & TV shows"

The request doesn't fail and the ConnectedRange uses the correct `attributeName` passed down as a prop.

**What is the version you are using? Always use the latest one before opening a bug issue.** npm@latest

samouss:
**Summary**

Fix #630 

In the `connectRange` we need to access the stats & the values of the given facet in order to compute the range. But it may happen that the facet is not yet present on the results so we need to check that the facet is present on results before accessing it.

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectRange.js
+++ b/packages/react-instantsearch/src/connectors/connectRange.js
@@ -190,8 +190,13 @@ export default createConnector({
   getProvidedProps(props, searchState, searchResults) {
     const { attributeName, precision, min: minBound, max: maxBound } = props;
     const results = getResults(searchResults, this.context);
-    const stats = results ? results.getFacetStats(attributeName) || {} : {};
-    const count = results
+    const isFacetExist = results && results.getFacetByName(attributeName);
+
+    const stats = isFacetExist
+      ? results.getFacetStats(attributeName) || {}
+      : {};
+
+    const count = isFacetExist
       ? results.getFacetValues(attributeName).map(v => ({
           value: v.name,
           count: v.count,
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch/src/connectors/connectRange.js
```diff
@@ -190,8 +190,13 @@ export default createConnector({
   getProvidedProps(props, searchState, searchResults) {
     const { attributeName, precision, min: minBound, max: maxBound } = props;
     const results = getResults(searchResults, this.context);
-    const stats = results ? results.getFacetStats(attributeName) || {} : {};
-    const count = results
+    const isFacetExist = results && results.getFacetByName(attributeName);
```
Haroenv: `facetExists` seems like a better name

samouss: I prefer to prefix all the boolean value like one with `is` or `has` it make them explicit. Would you agree for `isFacetExists` ?

Haroenv: makes sense, but grammatically it reads weird. What about `hasFacet` as something which matches the pattern but also grammar?

samouss:
@Haroenv should be good now!

PR #804, opened 2018-01-05
fix(stories): avoid to use linear-background it breaks Argos every time

samouss:
**Summary**

Use an other style attribute for the custom root prop, because the background with the `linear-gradient` breaks Argos almost every time.

### DIFF
--- a/stories/InstantSearch.stories.js
+++ b/stories/InstantSearch.stories.js
@@ -41,7 +41,7 @@ stories
           Root: 'div',
           props: {
             style: {
-              background: 'linear-gradient(80deg, #00D8FF, #00A7FF)',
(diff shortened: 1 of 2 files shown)


stories/InstantSearch.stories.js
```diff
@@ -41,7 +41,7 @@ stories
           Root: 'div',
           props: {
             style: {
-              background: 'linear-gradient(80deg, #00D8FF, #00A7FF)',
+              padding: '30px',
```
Haroenv: what about: `border: 1em solid red`? that will still be very easy to see, but will also show up in Argos

samouss: Great idea!

Issue #803, opened 2018-01-05
Docs for Highlighter: `separatorComponent` should be `separator`


Greenie10:
In the documentation for [Highlighter](https://github.com/algolia/react-instantsearch/blob/dc4d702600482cd759a8c0d53778ed702cee8964/packages/react-instantsearch/src/components/Highlighter.js), the Prop `separatorComponent` should be `separator` as declared in the [Highlighter.js component in GitHub](https://github.com/algolia/react-instantsearch/blob/dc4d702600482cd759a8c0d53778ed702cee8964/packages/react-instantsearch/src/components/Highlighter.js) - or perhaps the Highlighter.js file should be edited to say `separatorComponent`, which I think I prefer :)

Greenie10:
Have also updated the PropTypes in Highlight and Snippet components, as per @samouss request.

PR #809, opened 2018-01-06
Update prop name

Greenie10:
**Summary**

Amending the Highlight.js widget documentation to correct prop name for `separator`. The rationale for changing documentation rather than change to actual prop is because the single word `separator` is more in keeping with the other props.  Issue from  https://github.com/algolia/react-instantsearch/issues/803


samouss:
Close in favour of #810

samouss:
Hi @Greenie10, thanks for the contribution! 

We should also update the `PropTypes` on the `Highlight` component.

https://github.com/algolia/react-instantsearch/blob/170c3f1122dc56d6e011b292251f75c03250f76b/packages/react-instantsearch/src/components/Highlight.js#L15

PR #817, opened 2018-01-08
Escape HTML in changelog

hkdobrev:
It seems the changelog generation tool you've used doesn't take
into account that HTML-looking code is good to be escaped when
converted to Markdown. This is something which could be looked into separately.

Also, I'd prefer to go with fixed-width inline code elements for things like that: `InstantSearch` and `<InstantSearch>`.

**Summary**

**Result**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 
 * **createInstantSearch:** remove the client from the Snapshot ([#749](https://github.com/algolia/react-instantsearch/issues/749)) ([700d8f4](https://github.com/algolia/react-instantsearch/commit/700d8f4))
 * refresh cache memory leak example ([#784](https://github.com/algolia/react-instantsearch/issues/784)) ([cf228ac](https://github.com/algolia/react-instantsearch/commit/cf228ac))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 * **createInstantSearch:** remove the client from the Snapshot ([#749](https://github.com/algolia/react-instantsearch/issues/749)) ([700d8f4](https://github.com/algolia/react-instantsearch/commit/700d8f4))
 * refresh cache memory leak example ([#784](https://github.com/algolia/react-instantsearch/issues/784)) ([cf228ac](https://github.com/algolia/react-instantsearch/commit/cf228ac))
-* **stories:** rename InstantSearch to <InstantSearch> ([#789](https://github.com/algolia/react-instantsearch/issues/789)) ([05efda5](https://github.com/algolia/react-instantsearch/commit/05efda5))
+* **stories:** rename InstantSearch to \<InstantSearch\> ([#789](https://github.com/algolia/react-instantsearch/issues/789)) ([05efda5](https://github.com/algolia/react-instantsearch/commit/05efda5))
```
Haroenv: What about `<InstantSearch>` (simpler to read)

Thanks for the catch!

PR #829, opened 2018-01-10
chore: add .prettierrc

Haroenv:
I couldn't find a way to have a sharable config, but my default configuration was overriding the one for this repository, so this is useful.


samouss:
We run `prettier` through `eslint` for the project. So the configuration is taken from `eslint` rather than `prettier` I think.

Haroenv:
Yes, this is for prettier in markdown mostly

PR #830, opened 2018-01-10
fix(currentRefinements): give access to id and index from transformItems for deduplication

Haroenv:
**Summary**

The currentRefinemements widget (and connector) will give exactly the same data as the widgets which are present in the layout. As seen in #258, when there are two refinements with the same attribute, this will give those items the same `key` and render with warnings. 

**Result**

1. Give access to `index` and `id` (`${attributeName}`, `query`, ...) to `transformItems` so users can deduplicate themselves
2. test how the deduplication can be done


Haroenv:
There's one problem I can think of by doing that. Two _separate_ widgets with the same attribute should be shown twice (menu / refinementList), but they both use `attributeName` as `id`. Should that be this? 

```js
function getId(props) {
  return `${namespace}.${props.attributeName}`;
}
```

PR #839, opened 2018-01-12
feat: add className on root element for all widgets

samouss:
**Summary**

This PR enable the possibility to pass a `className` that will be applied **only** on the root element of the widgets. It's available on all the widgets that output DOM. It can be easier to pass a custom `className` rather than treat with the built-in classNames when we do styling.

**Result**

```jsx
<Menu 
  className="MyCustomMenu"
  attributeName="category"
/>

// Will output

<div className="ais-Menu MyCustomMenu">
  Rest of the Menu DOM
</div>

### CODE REVIEW

**Comment by Haroenv** [comment]
How will you pass classnames to inner components? Or just not a feature and needs to be solved by more selectors?

**Comment by samouss** [comment]
To inner DOM elements you mean?

**Comment by Haroenv** [comment]
yep

File: packages/react-instantsearch/src/components/Highlight.js [hunk 1]
```diff
@@ -9,13 +8,4 @@ const Highlight = props => (
   <Highlighter {...props} highlightProperty="_highlightResult" cx={cx} />
 );
 
-Highlight.propTypes = {
```
Haroenv: Why is this removed?

packages/react-instantsearch/src/components/RefinementList.js
```diff
@@ -34,8 +29,19 @@ class RefinementList extends Component {
...
   };
 
+  static defaultProps = {
+    className: '',
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = { query: '' };
```
Haroenv: can't we use the `state = { query: '' };` pattern, since class fields are transpiled anyway

packages/react-instantsearch/src/components/RefinementList.test.js
```diff
@@ -2,10 +2,10 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 import Enzyme, { mount } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
-Enzyme.configure({ adapter: new Adapter() });
-
 import RefinementList from './RefinementList';
 
+Enzyme.configure({ adapter: new Adapter() });
```
Haroenv: not for this PR, but maybe it should be done globally and overridden if we test another version?

packages/react-instantsearch/src/components/Snippet.js
```diff
@@ -9,13 +8,4 @@ const Snippet = props => (
   <Highlighter {...props} highlightProperty="_snippetResult" cx={cx} />
 );
 
-Snippet.propTypes = {
```
Haroenv: same Q as earlier. Is it because `<Snippet />` already has PropTypes?

packages/react-instantsearch/src/components/Highlight.js
samouss: Don't need to duplicate the props since we don't use it at that stage, we just forward them. All the props are still typed on `Highlighter`.

vvo:
For future ideas (we discussed that too): why not just accept every prop on the root component and then isolate ours (or pass a extraProps={} like other libs are doing).

Haroenv:
`rootProps` seems also useful

PR #845, opened 2018-01-13
refactor(RatingMenu): always use a anchor element

samouss:


Haroenv:
Seems like the error was because of an outage on npm??

packages/react-instantsearch/src/components/RatingMenu.js
```diff
@@ -109,17 +108,17 @@ class RatingMenu extends Component {
           disabled && 'item--disabled'
         )}
       >
-        <StarsWrapper
-          aria-label={`${rating}${translate('ratingLabel')}`}
+        <a
           className={cx('link')}
+          aria-label={`${rating}${translate('ratingLabel')}`}
```
Haroenv: Shouldn’t this have a space in-between those? and maybe ratingLabel should be a function as translate key instead?

samouss: The space is in the default translated value (see below). IMO the `translate` function should accept parameter in order to enable placeholder in the translation. But it's not for this PR.

```js
// Usage
translate('rating', { value: 5 });

// Translations
{
  rating: '{ value } & up'
}
```

https://github.com/algolia/react-instantsearch/blob/cb991194724bb8700489bfdbcee58899e820707c/packages/react-instantsearch/src/components/RatingMenu.js#L182-L184

Haroenv: all right, that sounds reasonable. We should definitely go for more functions in the future

PR #850, opened 2018-01-15
refactor: rename xxxComponent props

samouss:
**Summary**

This PR rename some props from `xxxComponent` to `xxx` because they don't really accept a component. They just accept anything that can be render by React (aka: [`node`](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes)). Only `Hits` & `InifiniteHits` accept a real component. The principal changes are on the `SearchBox`, otherwise it's mainly docs changes.

I also moved the props related to the rendering from `connectBreadcrumb` to the `Breadcrumb` component since they are unused in the connector.

**SearchBox**:

- `submitComponent` -> `submit`
- `resetComponent` -> `reset`
- `loadingIndicatorComponent` -> `loadingIndicator`

### DIFF
--- a/packages/react-instantsearch/src/components/Breadcrumb.js
+++ b/packages/react-instantsearch/src/components/Breadcrumb.js
@@ -21,12 +21,14 @@ class Breadcrumb extends Component {
     items: itemsPropType,
     refine: PropTypes.func.isRequired,
     rootURL: PropTypes.string,
-    separator: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+    separator: PropTypes.node,
     translate: PropTypes.func.isRequired,
     className: PropTypes.string,
   };
 
   static defaultProps = {
+    rootURL: null,
+    separator: '>',
     className: '',
   };
 
--- a/packages/react-instantsearch/src/components/Breadcrumb.js
+++ b/packages/react-instantsearch/src/components/Breadcrumb.js
@@ -28,7 +28,7 @@ class Breadcrumb extends Component {
 
   static defaultProps = {
     rootURL: null,
-    separator: '>',
+    separator: ' > ',
     className: '',
   };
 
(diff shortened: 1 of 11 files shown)


packages/react-instantsearch/src/components/__snapshots__/Breadcrumb.test.js.snap
```diff
@@ -284,7 +306,9 @@ exports[`Breadcrumb outputs the default breadcrumb with no refinement 1`] = `
     >
       <span
         className="ais-Breadcrumb-separator"
-      />
+      >
+        &gt;
```
Haroenv: is this change on purpose? I don't see the change in the test

samouss: Yes since I move the `defaultProps` from the connector to the component it print now the default separator in the snapshot.

Haroenv: Ah okay makes sense

packages/react-instantsearch/src/components/Breadcrumb.js
```diff
@@ -21,12 +21,14 @@ class Breadcrumb extends Component {
...
     rootURL: PropTypes.string,
-    separator: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+    separator: PropTypes.node,
     translate: PropTypes.func.isRequired,
     className: PropTypes.string,
   };
 
   static defaultProps = {
+    rootURL: null,
+    separator: '>',
```
Haroenv: Should it have spaces around?

samouss: Don't think so, the default theme adds margin around the separator.

Haroenv: Since it's `span` it will get read as `level>sublevel` and might not be what we want

samouss: For screen reader you mean?

Haroenv: yes

PR #853, opened 2018-01-15
chore(theme): use jsDelivr by default and set alias

vvo:
We recommended unpkg at the begining, it was the only one serving other
files than JS from npm.

Now jsDelivr supports it by default since a long time, and we can inform
jsDelivr whats the default file to serve (to avoid style.min.css).

This will only work on the new release of the lib.

Also I updated the versions of the theme in the docs, might be worth
automating it or you'll have to do it next time too.

### DIFF
--- a/docgen/src/Getting_started.md
+++ b/docgen/src/Getting_started.md
@@ -105,7 +105,7 @@ We provide an Algolia theme that should be a good start.
 Include it in your webpage with this CDN link or copy paste the raw content:
 
 ```html
-<link rel="stylesheet" href="https://unpkg.com/react-instantsearch-theme-algolia@4.0.0/style.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/react-instantsearch-theme-algolia@4.4.1">
 ```
 
 Read the [styling](guide/Styling_widgets.html) guide for more information.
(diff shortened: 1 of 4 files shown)


samouss:
Maybe it's not worth it to automatise it, because the theme package will be removed in the next major release in favour of IS.css.

vvo:
> Maybe it's not worth it to automatise it, because the theme package will be removed in the next major release in favour of IS.css.

Yep indeed, I thought we already had this automation in the doc somewhere but if not OK.

(Ultimately we will want the version of the lib in the doc, to be able to show "last version: x.x.x", let's see that once we revisit the websites)

vvo:
Relacunhed netlify again, got timeout too, strannnge

samouss:
The `yarn` version used by Netlify is really old (`v0.18.1`) maybe it's related? Don't know.

vvo:
> The yarn version used by Netlify is really old (v0.18.1) maybe it's related? Don't know.

Oh yes, we must use the same as in dev (1.3.2). It can be done in two ways:
- via env variables inside netlify build configuration, define YARN_VERSION to 1.3.2
- via netlify.toml (see https://github.com/algolia/community-project-boilerplate/blob/master/netlify.toml)

vvo:
We're lucky it worked until now :D

PR #855, opened 2018-01-16
Feature/instantsearchcss/panel

samouss:
**Summary**

Restore the `<Panel>` on the docs example.


docgen/src/examples/e-commerce-infinite/App.js
```diff
@@ -68,43 +68,39 @@ const Facets = () => (
 
     <section className="facet-wrapper">
       <div className="facet-category-title facet">Show results for</div>
+
```
Haroenv: nit: whitespace is not necessary here

samouss: My bad, I don't like when the code is too compressed 🙂

PR #856, opened 2018-01-16
chore: Netlify configuration

samouss:
Add a Netlify configuration file for update the `yarn` version. (Not sure about the format)

cc @vvo


samouss:
It seems to work at least for the preview!

vvo:
> at least for the preview!

We use netlify for now only for the preview mode, so all good. I like those configuration files because then the way to deploy is explicit into the code rather than hidden in a dashboard.

PR #866, opened 2018-01-17
Feature/instantsearchcss/remove theme

samouss:
**Summary**

This PR removes the `react-instantsearch-theme-algolia` package since we now rely on IS.css.


scripts/test:build.sh
```diff
@@ -4,5 +4,4 @@
 # to ensure we will be able to release new versions
 
 (cd packages/react-instantsearch && yarn build) &&
```
Haroenv: stray &&

samouss: I think `&&` is required otherwise the script will continue even if the previous command has failed. Not sure.

Haroenv: that's what `set -e` is for. `&&` is implied on every newline then IIRC

samouss: Ok, I will add it on this file and also in the `boot.sh` then.

PR #874, opened 2018-01-18
Use substr instead of startsWith

### LINKED ISSUE #869: IE 11 doesn't support String.startsWith()
`react-instantsearch` 4.4.1 breaks in IE 11 because of the use of [`startsWith()` in `createConnector`](https://github.com/algolia/react-instantsearch/blob/9c5106de70fb08de851e3aaca0ab7ce20d5eac04/packages/react-instantsearch/src/core/createConnector.js#L97) — I'm unsure if this is on purpose or not, since I can't find a list of supported browsers. Are users expected to polyfill this themselves or should it be a part of the library?

denkristoffer:
Necessary for IE 11 compatibility and avoids having to use a polyfill. Fixes #869

### DIFF
--- a/packages/react-instantsearch/src/core/createConnector.js
+++ b/packages/react-instantsearch/src/core/createConnector.js
@@ -94,7 +94,7 @@ export default function createConnector(connectorDesc) {
 
           if (
             onlyGetProvidedPropsUsage &&
-            !connectorDesc.displayName.startsWith('Algolia')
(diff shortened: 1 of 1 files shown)


Haroenv:
Did you see any other incompatibilities?

denkristoffer:
Nope, this was the only one IE 11 was complaining about in our app.

PR #884, opened 2018-01-22
docs: add missing providedProps for connectRange 

### LINKED ISSUE #882: How to get auto computed Min and Max for connectRange
Hi,

I'm working on a RangeSlider component and would like to know how to get the auto computed Min and Max values as mentioned in the docs.

>name | type | description
>---|---|---
> min | type: number |  Minimum value. When this isn’t set, the minimum value will be automatically computed by Algolia using the data in the index.
> max | type: number |  Maximum value. When this isn’t set, the maximum value will be automatically computed by Algolia using the data in the index.

I've tried currentRefinement but that shows undefined until you actually start moving the slider, then it has the proper min and max. Also this doesn't seem correct as it will change as a refinement is applied.

Thanks,

samouss:
**Summary**

Fixes #882


samouss:
@Haroenv right, the documentation is build from those comments.

PR #906, opened 2018-01-26
refactor(CurrentRefinements): revert behaviour changes

samouss:
**Summary**

This PR restore the previous behaviour of `CurrentRefinements`. The component also move for a SFC since we didn't use state & hooks.

**Result**

You can play with the widget on [Storybook](https://deploy-preview-906--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=CurrentRefinements&selectedStory=with%20RefinementList&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs).

### DIFF
--- a/docgen/src/guide/Migration_guide_v5.md
+++ b/docgen/src/guide/Migration_guide_v5.md
@@ -216,7 +216,7 @@ No change.
 
 #### Behaviour
 
-Instead of displaying the widget as `category: One ✕ Two ✕`, it now displays like this: `One ✕` `Two ✕`.
+No change.
 
 #### CSS classes equivalency table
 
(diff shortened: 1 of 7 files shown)


docgen/src/guide/Migration_guide_v5.md
```diff
@@ -187,7 +187,7 @@ No change.
 
 #### Behaviour
 
-Instead of displaying the widget as `category: One ✕ Two ✕`, it now displays like this: `One ✕` `Two ✕`.
+No change.
```
Haroenv: aren't the classes still different?

samouss: Right 😅

samouss: Actually in the migration guide there is nothing to add, since we list only the changes of the classnames. But I updated the widget file with the missing classnames.

Issue #891, opened 2018-01-23
connectToggle.js properties docs


catamphetamine:
```
@providedPropType {boolean} currentRefinement - the refinement currently applied
```

should be replaced with something like

```
@providedPropType {boolean} currentRefinement - current checkbox value
```

because it is not "the refinement currently applied"

catamphetamine:
Also

```js

  getMetadata(props, searchState) {
    const id = getId(props);
    const checked = getCurrentRefinement(props, searchState, this.context);
    const items = [];
    const index = getIndex(this.context);
    if (checked) {
      items.push({
        label: props.label,
        currentRefinement: props.label,
        attributeName: props.attributeName,
        value: nextState => refine(props, nextState, false, this.context),
      });
    }
    return { id, index, items };
  },
```

Most likely `currentRefinement: props.label,` is a bug.

PR #908, opened 2018-01-26
docs(connectToggle): currentRefinement description

samouss:
**Summary**

Close partially #891.

Also remove the part on the `off` option, since we don't support this option.

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectToggle.js
+++ b/packages/react-instantsearch/src/connectors/connectToggle.js
@@ -42,7 +42,7 @@ function cleanUp(props, searchState, context) {
 
 /**
  * connectToggle connector provides the logic to build a widget that will
- *  provides an on/off filtering feature based on an attribute value. Note that if you provide an “off” option, it will be refined at initialization.
+ * provides an on/off filtering feature based on an attribute value.
  * @name connectToggle
  * @kind connector
  * @requirements To use this widget, you'll need an attribute to toggle on.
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch/src/connectors/connectToggle.js
```diff
@@ -56,7 +56,7 @@ function cleanUp(props, searchState, context) {
  * @propType {boolean} [defaultRefinement=false] - Default searchState of the widget. Should the toggle be checked by default?
  * @providedPropType {function} refine - a function to toggle a refinement
  * @providedPropType {function} createURL - a function to generate a URL for the corresponding search state
- * @providedPropType {boolean} currentRefinement - the refinement currently applied
+ * @providedPropType {boolean} currentRefinement - `true` when the refinement is applied, `false` otherwise
```
Haroenv: I think the currentRefinement is `[]` or [{...stuff, currentRefinement: true}]` makes more sense

samouss: I don't understand? The `currentRefinement` provided by the connector is a boolean value. The shape that you describe is the one provided by the `getMetadata` function I think.

https://github.com/algolia/react-instantsearch/blob/9a8a154ca937bf7fa9df9b41b3a79b86a41613ed/packages/react-instantsearch/src/components/Toggle.js#L19-L31

Haroenv: oh yes that's true, never mind then 😂

PR #909, opened 2018-01-26
fix(connectToggle): use currentRefinement in metadata instead of the label

### LINKED ISSUE #891: connectToggle.js properties docs

samouss:
**Summary**

Close partially #891

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectToggle.js
+++ b/packages/react-instantsearch/src/connectors/connectToggle.js
@@ -109,14 +109,16 @@ export default createConnector({
     const checked = getCurrentRefinement(props, searchState, this.context);
     const items = [];
     const index = getIndex(this.context);
+
     if (checked) {
       items.push({
         label: props.label,
-        currentRefinement: props.label,
+        currentRefinement: checked,
         attributeName: props.attributeName,
         value: nextState => refine(props, nextState, false, this.context),
       });
     }
+
     return { id, index, items };
   },
 });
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch/src/connectors/connectToggle.js
```diff
@@ -109,14 +109,16 @@ export default createConnector({
     const checked = getCurrentRefinement(props, searchState, this.context);
     const items = [];
     const index = getIndex(this.context);
+
     if (checked) {
       items.push({
```
Haroenv: really weird conditional pushing IMO

packages/react-instantsearch/src/connectors/connectToggle.js
```diff
@@ -109,14 +109,16 @@ export default createConnector({
...
         label: props.label,
-        currentRefinement: props.label,
+        currentRefinement: checked,
```
Haroenv: makes sense wrt that issue

packages/react-instantsearch/src/connectors/connectToggle.js
samouss: Which behaviour you will expect instead?

Haroenv:
```js
const items = checked ? [{..stuff}] : []
```

I'm not sure if it's that much better though

samouss: Yep not sure, because the ternary will be split on several line.

PR #913, opened 2018-01-26
docs(react): fix react-instantsearch SSR example and improve SSR docs

flybayer:
**Summary**

I spent all day trying to get SSR working. I finally realized the `this.state` conditional was preventing the SSR search query from being passed to `<InstantSearch>` on the server. 

This PR improves the example to make it more clear what is happening (everything I had to learn the hard way 😄 

It also adds more details to the SSR documentation.

**Result**

I manually verified SSR is working with this code both in my project and in the example project. All tests are passing.

### DIFF
--- a/docgen/src/guide/Server-side_rendering.md
+++ b/docgen/src/guide/Server-side_rendering.md
@@ -25,7 +25,7 @@ When called, `createInstantSearch` returns:
 * A dedicated [`<InstantSearch>`](widgets/<InstantSearch>.html) component accepting a `resultsState` prop containing the Algolia results.
 * A `findResultsState` function to retrieve a `resultsState`.
 
-The server-side reference is available in the [API docs](server-side-rendering/).
+More details are available in the [server-side API docs](server-side-rendering/).
 
 We split this guide in three parts:
 - App.js is the server and browser shared main React component from your application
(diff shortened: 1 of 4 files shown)


Haroenv:
Thank you!

flybayer:
@samouss, I think `initialSearchState` is superior for the following reasons:

1. Example code should err on the side of being extra explicit and clear because the whole point is to teach others about how to use it and how it works.
2. Someone that's not an expert in React can easily get confused when the same variable name is used twice in different places for different purposes.
3. Only negatives are stylistic and a few more characters in the code.

Let me know what you think :)

samouss:
Thanks for the explanation @flybayer!

When we create the initial `searchState` we already are in the scope of a function called `getInitialProps` the intent is pretty clear. It also means that you know how Next works since this function is used internally by the Framework. By design this function has been created for retrieve initial data from async actions.

For the `resultsState` it doesn't add much sense neither since we only use the variable for initialise the `<App>`. We don't use the variable elsewhere. 

I also prefer to keep the actual wording for a better coherence across the docs and the examples. Inside the documentation we talk about `searchState` & `searchResults` find the same references in the example help to the comprehension.

@Haroenv @vvo thoughts on the subjects?

PR #923, opened 2018-01-29
fix(connectRange): use the same behaviour for currentRefinement in getMetadata

### LINKED ISSUE #920: connectRange "currentRefinements" from getProvidedProps is not aligned with getMetadata
I'm using a Multi Slider in connectRange with auto computed max and min range. When I adjust the slider markers it is reflected in my currentRefinement widget. The issue is that if I leave either slider marker at the autoComputed max or min, that value shows up as undefined. Otherwise values show up as they should. I've traced it through console log and the refine() function is passed the correct max or min value, but when it shows up in currentRefinements its undefined, but only if its the max or min value. Any ideas?

Console log:

max min undefined

Here's my code:

```
export const CurrentRefinements = connectCurrentRefinements(({ refine, items }) => {
  console.log('CurrentRefinements items', items);
  return(
    <View style={items.length === 0 ? null : {padding:8}}>
      <FlatList
        data={items}
        horizontal
        keyExtractor={(item, index) => item.objectID}
        renderItem={({ item }) => {
          const RenderRefinement = (() => {
              var renderString = ''
              switch(item.attributeName) {
                case 'categories.lv0':
                  renderString = `Category: ${item.currentRefinement}`              
                  break;
                case 'companyName':
                  renderString = `Company: ${item.currentRefinement}`
                  break;
                case 'price':
                  renderString = `Price: ${item.currentRefinement.min} - ${item.currentRefinement.max}`
                  break;
              return(
                renderString
              ) 
          })
          return (
            <View style={{padding: 8}}>
              <TouchableOpacity
                style={{backgroundColor: Colors.secondaryColor, padding: 6}}
                onPress={() => {
                  refine(item.value);
                }}
              >
                <View style={{alignItems: 'center', justifyContent: 'center', flexDirection: 'row'}}>
                  <Text style={{color: Colors.primaryGrayColor}}>
                    <RenderRefinement/>
                  </Text>
                  <Icon style={{paddingLeft: 2}} name={'close'} size={20} color={Colors.primaryGrayColor}/>
                </View>
              </TouchableOpacity>
            </View>
          );
        }}
      />
    </View>
  )
})


export const DrawerRangeMultiSlider = connectRange(({refine, currentRefinement, header, min, max}) => {
  return(
    <RangeMultiSlider refine={refine} currentRefinement={currentRefinement} header={header} min={min} max={max}/>
  )
})

export class RangeMultiSlider extends React.Component{
  constructor(props){
    super(props);
    this.state = {
      sliderValues: {
        min: this.props.min, 
        max: this.props.max
      },
      sliderRange: {
        min: this.props.min,
        max: this.props.max
      },
      markersEnabled: true,
    }; 
  }

  round(value, decimals) {
    return Number(Math.round(value+'e'+decimals)+'e-'+decimals);
  }

  onChangeValues = (values) => {
    this.setState({
      sliderValues: {
        min: this.round(values[0],2),
        max: this.round(values[1],2)
      }
    }, console.log(this.props.header,' this.state.sliderValues', this.state.sliderValues))
  }

  componentWillReceiveProps(nextProps){

    if(nextProps.min === nextProps.max){
      this.setState({
        markersEnabled: false,
      })
    } else if (nextProps.min !== nextProps.max && this.state.markersEnabled === false){
      this.setState({
        markersEnabled: true,
      })      
    }

    if(this.state.sliderValues.min !== nextProps.currentRefinement.min || this.state.sliderValues.max !== nextProps.currentRefinement.max){
      this.setState({
        sliderValues: {
          min: nextProps.currentRefinement.min, 
          max: nextProps.currentRefinement.max
        },
      })
    }
    if(this.state.sliderRange.min !== nextProps.min || this.state.sliderRange.max !== nextProps.max){
      this.setState({
        sliderRange: {
          min: nextProps.min,
          max: nextProps.max
        }
      })
    }
  }

  onChangeValuesFinish() {
    if (this.props.currentRefinement.min !== this.state.sliderValues.min || this.props.currentRefinement.max !== this.state.sliderValues.max) {
      console.log(this.props.header,' onChangeValuesFinish this.state.sliderValues',this.state.sliderValues)
      this.props.refine({
        min: this.state.sliderValues.min, 
        max: this.state.sliderValues.max
      });
    }    
  }

  render(){
    return(
    <View style={{padding:8}}>
      <Text style={{paddingLeft: 10, fontSize: 14, color: Colors.primaryTextColor, alignSelf: 'flex-start' , textAlign: 'left'}}>
        {`${this.props.header}:`}
      </Text>
      <View style={{paddingTop: 20, paddingLeft: 8}}>
        <MultiSlider
          sliderLength={150}
          containerStyle={{height:10}}
          onValuesChange={(sliderState) => this.onChangeValues(sliderState)}
          min={this.state.sliderRange.min}
          max={this.state.sliderRange.max}
          onValuesChangeFinish={() => this.onChangeValuesFinish()}
          values={[this.state.sliderValues.min, this.state.sliderValues.max]}
          step={1}/>
      </View>
      <View style={{flexDirection: 'row', justifyContent: 'space-between'}}>
        <Text style={{fontSize: 14, color: Colors.primaryTextColor, textAlign: 'left'}}>
          {this.state.sliderValues.min}
        </Text>
        <Text style={{fontSize: 14, color: Colors.primaryTextColor , textAlign: 'right'}}>
          {this.state.sliderValues.max}
        </Text>
      </View>
    </View>
    )
  }
}
```

samouss:
**Summary**

Close #920 

We provide two different values for the `currentRefinement` in `getProvidedProps` & `getMetadata`. In the first case we "normalize" the value with the range, in the latter we pass the raw values. It can lead to some issues when the user use the `currentRefinement` from the `connectCurrentRefinement` since the values aren't aligned.

**Edit:** At some point we should avoid to "normalize" the `currentRefinement` with the range from the connector. Passing the raw values to the view and let it decide which value to render is more flexible. It's a bit more work to do in the view but the connector will be more reusable (ex: We can implement a "correct" behaviour for `RatingMenu`).

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectRange.test.js
+++ b/packages/react-instantsearch/src/connectors/connectRange.test.js
@@ -470,7 +470,7 @@ describe('connectRange', () => {
           {
             label: '5 <= wot',
             attributeName: 'wot',
-            currentRefinement: { min: 5, max: undefined },
+            currentRefinement: { min: 5, max: 100 },
             // Ignore clear, we test it later
             value: metadata.items[0].value,
           },
@@ -505,7 +505,7 @@ describe('connectRange', () => {
           {
             label: 'wot <= 10',
             attributeName: 'wot',
-            currentRefinement: { min: undefined, max: 10 },
+            currentRefinement: { min: 0, max: 10 },
             value: metadata.items[0].value,
           },
         ],
@@ -753,7 +753,7 @@ describe('connectRange', () => {
           {
             label: '5 <= wot',
             attributeName: 'wot',
-            currentRefinement: { min: 5, max: undefined },
+            currentRefinement: { min: 5, max: 100 },
             // Ignore clear, we test it later
             value: metadata.items[0].value,
           },
@@ -792,7 +792,7 @@ describe('connectRange', () => {
           {
             label: 'wot <= 10',
             attributeName: 'wot',
-            currentRefinement: { min: undefined, max: 10 },
+            currentRefinement: { min: 0, max: 10 },
             value: metadata.items[0].value,
           },
         ],
--- a/packages/react-instantsearch/src/connectors/connectRange.js
+++ b/packages/react-instantsearch/src/connectors/connectRange.js
@@ -107,6 +107,13 @@ function getCurrentRefinement(props, searchState, currentRange, context) {
   return refinement;
 }
 
+function getCurrentRefinementWithRange(refinement, range) {
+  return {
+    min: refinement.min !== undefined ? refinement.min : range.min,
+    max: refinement.max !== undefined ? refinement.max : range.max,
+  };
+}
+
 function nextValueForRefinement(hasBound, isReset, range, value) {
   let next;
   if (!hasBound && range === value) {
@@ -217,7 +224,7 @@ export default createConnector({
       max: rangeMax,
     };
 
-    const { min: valueMin, max: valueMax } = getCurrentRefinement(
+    const currentRefinement = getCurrentRefinement(
       props,
       searchState,
       this._currentRange,
@@ -228,10 +235,10 @@ export default createConnector({
       min: rangeMin,
       max: rangeMax,
       canRefine: count.length > 0,
-      currentRefinement: {
-        min: valueMin === undefined ? rangeMin : valueMin,
-        max: valueMax === undefined ? rangeMax : valueMax,
-      },
+      currentRefinement: getCurrentRefinementWithRange(
+        currentRefinement,
+        this._currentRange
+      ),
       count,
       precision,
     };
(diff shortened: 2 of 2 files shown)


packages/react-instantsearch/src/connectors/connectRange.test.js
```diff
@@ -470,7 +470,7 @@ describe('connectRange', () => {
           {
             label: '5 <= wot',
             attributeName: 'wot',
-            currentRefinement: { min: 5, max: undefined },
```
Haroenv: so does undefined not work anymore to refine with only min?

samouss: Nope I didn't update the `refine` behaviour. I only have updated the `currentRefinements` value provided by the `getMetadata` function.

samouss:
@Haroenv @vvo Any blockers?

PR #925, opened 2018-01-30
docs: add deprecation warning

samouss:
**Summary**

Add a deprecation banner for the current documentation with a link on the next version documentation website.

**Result**

screen shot 2018-01-31 at 09 28 36

### DIFF
--- a/docgen/layouts/common/header.pug
+++ b/docgen/layouts/common/header.pug
@@ -3,4 +3,4 @@ div !{header}
     .banner__description
       a(href='http://react-instantsearch-beta.netlify.com')
         p
-          strong Checkout React InstantSearch V5! Our new and improved React InstantSearch version →
+          strong Check out React InstantSearch V5! Our new and improved React InstantSearch version →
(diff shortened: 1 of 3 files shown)


samouss:
I think we should link directly on the migration guide instead of the homepage. WDYT?

Haroenv:
The migration guide could be very prominent on the site, or in another banner?

samouss:
What do you mean? On which website?

docgen/layouts/common/header.pug
```diff
@@ -1 +1,6 @@
 div !{header}
+  .marketing-banner
+    .banner__description
+      a(href='http://react-instantsearch-beta.netlify.com')
+        p
+          strong Checkout React InstantSearch V5! Our new and improved React InstantSearch version →
```
Haroenv: Check out

samouss:
@Haroenv @vvo What about adding a yellow banner like the one on [InstantSearch.js]() for the v4 website. And for the v5 website we use the blue banner for notice the user that he is on the next version of the documentation. We also add a link to the migration guide inside it. WDYT?

Haroenv:
still necessary to add the banner to the beta site

PR #929, opened 2018-01-31
fix(StarRatings): always show the stars below

### LINKED ISSUE #674: StarRating '& up' only shows exact matches

1. [Visit this demo](https://internal-preview.algolia.com/movie-voice-demo/)
2. Filter by genre:drama & actor:alex rocco
3. Observe that that Rating widget has no value for 4* & up (It should show 1)

When showing the ratings using standard 'and up' behaviour. There should be no blank values when there are results for a higher rating. For example, if I have results for 1, 3 and 5 i'd expect 

5* & up (1)
4* & up (1)
3* & up (2)
2* & up (2)
1*  & up (3)

samouss:
**Summary**

Fix #674 

In the current implementation we can have sparse rows when the results only return one facet value or if some numeric filters (ex: rating > 3) are applied. Now all the rows under the highest value are filled. Since the v5 is in `beta` I didn't update the tests for avoid to much conflict at the merge. But we should get rid of the `react-test-renderer` for `enzyme`. Same for the component we could move to a SFC after the merge.

**Before:**

screen shot 2018-01-31 at 16 24 48

**After:**

screen shot 2018-01-31 at 16 24 37

You can play with the widget on [Storybook](https://deploy-preview-929--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=StarRating&selectedStory=default&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs).

### DIFF
--- a/packages/react-instantsearch/src/components/StarRating.js
+++ b/packages/react-instantsearch/src/components/StarRating.js
@@ -131,15 +131,16 @@ class StarRating extends Component {
 
     // min & max are always set when there is a results, otherwise it means
     // that we don't want to render anything since we don't have any values.
-    const limitMin = min !== undefined ? min : 1;
-    const limitMax = max !== undefined ? max : limitMin - 1;
-    const inclusivePlaceholderLength = limitMax - (limitMin - 1);
+    const limitMin = min !== undefined && min >= 0 ? min : 0;
+    const limitMax = max !== undefined && max >= 0 ? max : -1;
+    const inclusiveLength = limitMax - limitMin + 1;
+    const safeInclusiveLength = Math.max(inclusiveLength, 0);
 
     const values = count
       .map(item => ({ ...item, value: parseFloat(item.value) }))
       .filter(item => item.value >= limitMin && item.value <= limitMax);
 
-    const range = new Array(inclusivePlaceholderLength)
+    const range = new Array(safeInclusiveLength)
       .fill(null)
       .map((_, index) => {
         const element = values.find(item => item.value === limitMax - index);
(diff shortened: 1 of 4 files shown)


Haroenv:
https://deploy-preview-929--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=StarRating&selectedStory=default&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs

Haroenv:
@samouss nothing seems to happen in the playground now 🤔

samouss:
@Haroenv @vvo Any blockers?

packages/react-instantsearch/src/components/StarRating.js
```diff
@@ -128,39 +127,47 @@ class StarRating extends Component {
...
-    }
+
+    const items = range.map((item, index) =>
+      this.buildItem({
+        lowerBound: item.value,
+        count: item.total,
+        isLastSelectableItem: range.length - 1 === index,
+        max: limitMax,
+        translate,
+        createURL,
```
Haroenv: why is `refine` not in this object anymore?

samouss: Take a look at the signature of `buildItem` refine is not used. We access the function through the `props`.

https://github.com/algolia/react-instantsearch/blob/4064e0cac7dc96a23565a7f33d11b4de0e253434/packages/react-instantsearch/src/components/StarRating.js#L52-L59

PR #932, opened 2018-02-01
fix: update CSS import

samouss:
**Summary**

Remove the import of the `reset` from the documentation & examples, because it's already loaded with the `algolia` theme.

### DIFF
--- a/docgen/src/Getting_started.md
+++ b/docgen/src/Getting_started.md
@@ -107,10 +107,6 @@ We provide an Algolia theme that should be a good start.
 Include it in your webpage with this CDN link or copy paste the raw content:
 
 ```html
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.0.0/themes/reset-min.css">
-
-<!-- or -->
-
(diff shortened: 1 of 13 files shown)


docgen/src/Getting_started.md
```diff
@@ -108,6 +108,9 @@ Include it in your webpage with this CDN link or copy paste the raw content:
 
 ```html
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.0.0/themes/reset-min.css">
+
+<!-- or -->
+
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.0.0/themes/algolia-min.css">
```
ronanlevesque: I would also specify that `reset.css` is already included inside `algolia.css` to make things even clearer.

samouss:
I remove the `or` inside the "Getting Started" since it's a short guide I took the more straightforward approach which is: link only the `algolia` theme. For the other guide I added some precision.

cc @ronanlevesque @Haroenv

PR #949, opened 2018-02-06
fix(StarRating): move to 1 based instead of 0

samouss:
**Summary**

With the update of the `StarRating` widget I move for a zero based approach. But the most common approach is to have a 1 based menu. It could have some weird effect when you only set the `max` value on the first load.

**Before**

before

**After**

after


Haroenv:
This is a breaking change, did you list it in the migration guide?

samouss:
Nope it’s not it was the original behavior.

PR #1049, opened 2018-03-06
docs: update banner

samouss:
Update the banner text & color to match the release.

**Before**

screen shot 2018-03-06 at 16 41 32

**After**

screen shot 2018-03-06 at 16 41 23


samouss:
The link is not working, should be relative not absolute... Will change it.

PR #1050, opened 2018-03-06
fix(recipes): update Snaphot

samouss:


Haroenv:
when was this changed?

samouss:
With the current release flow we update the version of React InstantSearch in the recipes after the release of the package. Before pushing this update to master we should update the Snapshot but I forgot to amend the commit.

PR #1071, opened 2018-03-12
chore: remove requestAnimationFrame shim for Jest

samouss:
**Summary**

React 16.x requires an environment where `requestAnimationFrame` is available. Jest was not shipped with this function until the version 22.0.0 (see [CHANGELOG](https://github.com/facebook/jest/blob/master/CHANGELOG.md#features-11)). Now we can drop the custom shim that we have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,9 +138,6 @@
     "testPathIgnorePatterns": [
       "/packages/react-instantsearch/examples"
     ],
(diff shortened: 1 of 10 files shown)


samouss:
Yep, see [Jest docs](https://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string).

PR #1073, opened 2018-03-12
poc: connectGeoSeach with createConnector

samouss:
### **This PR is a POC, it's not intended to be merged**.

I played a bit around the connector API in order to handle a "'stateful" instance. This lets us the possibility to implement the UI logic directly in the connector.

- **`initialUiState`**: You can define a new property on the connector declaration. This property takes a function that returns the initial UI state for the connector. The function takes `props` as argument.

- **`setUiState`**: You can access this function in every method of the connector. It's wrapper around the `setState` function that lets you update only the `uiState`. It means that now we can trigger new render on the fly to respond to users interactions.

It's definitely not a final API, but it can fit pretty easily with the current implementation without rewriting everything. Happy to have your feedback!

You can play with the GeoSearch example on [Storybook](https://deploy-preview-1073--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=GeoSearch&selectedStory=default&full=0&addons=0&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs).

**Note**: the test are failing it's completely normal.

### DIFF
--- a/packages/react-instantsearch/src/components/GeoSearch.js
+++ b/packages/react-instantsearch/src/components/GeoSearch.js
@@ -33,6 +33,8 @@ class GeoSearch extends Component {
     this.fitViewToBounds();
   }
 
+  createRef = c => (this.element = c);
+
   onChange = () => {
     const { isRefineOnMapMove, setMapMoveSinceLastRefine } = this.props;
 
@@ -61,7 +63,7 @@ class GeoSearch extends Component {
 
     if (isFitBoundsEnable) {
       this.isUserInteraction = false;
-      this.mapElement.fitBounds(
+      this.element.fitBounds(
         items.reduce(
           (acc, item) =>
             acc.extend({ lat: item._geoloc.lat, lng: item._geoloc.lng }),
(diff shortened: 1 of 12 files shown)


packages/react-instantsearch/src/components/GeoSearch.js
```diff
@@ -0,0 +1,153 @@
...
+    const { setMapMoveSinceLastRefine } = this.props;
+
+    if (this.isMapAlreadyLoaded && this.isUserInteraction) {
+      this.isPendingRefine = true;
+      setMapMoveSinceLastRefine();
+    }
+  };
+
+  onIdle = () => {
+    const { isRefineOnMapMove } = this.props;
```
Haroenv: I don’t think this needs the is Prefix

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -63,24 +63,37 @@ export default function createConnector(connectorDesc) {
...
+        const initialUiState = connectorDesc.initialUiState
+          ? connectorDesc.initialUiState(props)
+          : {};
+
         this.state = {
-          props: this.getProvidedProps({ ...props, canRender }),
+          uiState: initialUiState,
+          props: this.getProvidedProps({
+            ...props,
+            ...initialUiState,
```
samouss: I don't like that, we should not merge that in the `props`. We should pass the `uiState` as an argument to the connector definition.

samouss:
Closing in favour of #1171.

PR #1093, opened 2018-03-20
chore: move examples

samouss:
**Summary**

This is a first step into moving the repo to a proper structure (with Workspaces). This PR only move the examples at the top level of the project. It decorrelates the examples from the main package (that will be split later on).

**Nest steps**

- move the main package to Workspaces
- move the docs to Workspaces


vvo:
So you are doing it progressively, and even if you are splitting packages, you still have the same import API? Neat!

samouss:
Yep that's the idea! I will create a "placeholder" package for `react-instantsearch` that will reproduce the same API that the package has right now.

PR #1095, opened 2018-03-21
chore: enable workspaces on packages

samouss:
**Summary**

This is a second step into moving the repo to a proper structure (with Workspaces). This PR enables the Workspaces only on the packages folder. All the `devDependencies` are hoisted on the top level packages. The `packages` will only have their own `dependencies`. I also get rid of the `boot` script replaced by `bootstrap`. Right now it cleans the project and run `yarn`.

**Nest steps**

- ~~move the main package to Workspaces~~
- move the docs to Workspaces


Haroenv:
```
3:55:33 PM: $ ./scripts/docs:build-preview.sh
3:55:34 PM: $ babel-node docgen/build.js && yarn docs:storybook-build-production
3:55:34 PM: /opt/build/repo/node_modules/babel-core/lib/transformation/plugin.js:131
3:55:34 PM:       throw new Error(messages.get("pluginInvalidProperty", loc, i, key));
3:55:34 PM:       ^
3:55:34 PM: Error: Plugin 0 specified in "/opt/build/repo/.babelrc.env.production" provided an invalid property of "__wrapped__"
3:55:34 PM:     at Plugin.init (/opt/build/repo/node_modules/babel-core/lib/transformation/plugin.js:131:13)
3:55:34 PM:     at Function.normalisePlugin (/opt/build/repo/node_modules/babel-core/lib/transformation/file/options/option-manager.js:152:12)
3:55:34 PM:     at /opt/build/repo/node_modules/babel-core/lib/transformation/file/options/option-manager.js:184:30
3:55:34 PM:     at Array.map (<anonymous>)
3:55:34 PM:     at Function.normalisePlugins (/opt/build/repo/node_modules/babel-core/lib/transformation/file/options/option-manager.js:158:20)
3:55:34 PM:     at OptionManager.mergeOptions (/opt/build/repo/node_modules/babel-core/lib/transformation/file/options/option-manager.js:234:36)
3:55:34 PM:     at OptionManager.init (/opt/build/repo/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
3:55:34 PM:     at compile (/opt/build/repo/node_modules/babel-register/lib/node.js:103:45)
3:55:35 PM:     at loader (/opt/build/repo/node_modules/babel-register/lib/node.js:144:14)
3:55:35 PM:     at Object.require.extensions.(anonymous function) [as .js] (/opt/build/repo/node_modules/babel-register/lib/node.js:154:7)
```

babel-node is not happy about the preview script apparently

package.json
```diff
@@ -1,8 +1,11 @@
 {
   "private": true,
   "version": "5.0.1",
+  "workspaces": [
+    "packages/*"
+  ],
   "scripts": {
-    "boot": "./scripts/boot.sh",
+    "bootstrap": "git clean -dfx && yarn",
```
vvo: Why not just "yarn"? Do we actually need to do git clean etc in here?

Wondering if we even now need any "bootstrap" script vs just `yarn`

netlify.toml
```diff
@@ -1,6 +1,6 @@
 [build]
   publish = "docs-production/"
-  command = "yarn boot && yarn docs:build-preview"
+  command = "yarn bootstrap && yarn docs:build-preview"
 
   [build.environment]
     YARN_VERSION = "1.3.2"
```
vvo: Maybe also update your Yarn version everywhere, good fixes I am sure for workspaces

packages/react-instantsearch/package.json
```diff
@@ -32,14 +32,9 @@
     "algoliasearch-helper": "^2.21.0",
     "classnames": "^2.2.5",
     "lodash": "^4.17.4",
-    "prop-types": "^15.5.10"
-  },
-  "devDependencies": {
-    "enzyme": "3.3.0",
-    "enzyme-adapter-react-16": "1.1.1",
```
vvo: If you have some dependencies that are only worth it for some packages you can also leave them in the package itself. What I am used to do is to have only the release dependencies (lerna) at the package root level while keeping every package self-contained and still be able to run npm test in every package.

wdyt?

package.json
Haroenv: yes, we should probably remove this script

samouss: The idea behind that is to have a "fresh" project. By running that command we reset the state of the  repository from the ground up (like a `git clone`). You can still run `yarn` since we deal with Workspaces the install will be correct. With the `bootstrap` the intent is a bit different.

packages/react-instantsearch/package.json
samouss: I saw a lots of different approach regarding that and I'm still not very convinced which one is the best fit. For us since we won't have a lots of packages (4-5 unlike Babel for example) your solution sounds legit. At the end we keep in the package dependencies all the one required by the scripts right? 

For example `babel` will be used by the `build` command so we should put it in **every** packages. On the other hand `eslint` will be run globally from the top level package so we can keep it only on the top level. Note that for the `babel` case if we want to correctly follow the self-contained rule we should have a `.babelrc` on every level also.

The other idea is to keep all the `devDependencies` on the top level package. So the packages only have their own `dependencies`. We break the self-contained rule but at the end it's a monorepo and everything is tight together.

WDYT?

netlify.toml
samouss: I'll do it in another PR.

packages/react-instantsearch/package.json
Haroenv: probably best to list it under devDeps in every “package”, so that you can pretend it’s its own package without the workspace too, seems the most consistent

samouss:
The message is collapsed because I edited the file but my remark is still valuable.

> https://github.com/algolia/react-instantsearch/pull/1095#discussion_r176340411

packages/react-instantsearch/package.json
samouss: I'll go for the solution that your proposed, having self-contained packages since we won't have hundred of packages it makes more sense. Thanks for the feedback!

package.json
```diff
@@ -22,7 +22,7 @@
     "test:regressions": "webpack --config test/regressions/webpack.config.js && rm -rf test/screenshots && happo run && babel-node ./test/regressions/normalize.js",
     "test:size": "bundlesize",
     "argos": "argos upload test/screenshots --token $ARGOS_TOKEN || true",
-    "dev": "jest --watch --bail",
+    "dev": "jest --watch --bail --projects packages/*",
```
samouss: I provide the `projects` option to the CLI because right now there is an issue with Jest when the projects target only one package. I'll remove this when the package will be split.

package.json
```diff
@@ -17,9 +20,8 @@
     "test:recipes": "./scripts/test:recipes.sh",
     "test:build": "NODE_ENV=production ./scripts/test:build.sh",
     "test:regressions": "webpack --config test/regressions/webpack.config.js && rm -rf test/screenshots && happo run && babel-node ./test/regressions/normalize.js",
-    "test:size": "bundlesize",
     "argos": "argos upload test/screenshots --token $ARGOS_TOKEN || true",
-    "dev": "jest --watch --bail",
+    "dev": "jest --watch --bail --projects packages/*",
```
samouss: I provide the `projects` option to the CLI because right now there is an issue with Jest when the projects target only one package. I'll remove this when the package will be split.

package.json
vvo: Ok got it. In which cases would you use bootstrap do you think? Since it can be dangerous to call git clean on the behalf of the user, I want to be sure we are providing the command for good reasons. Maybe you are already using it, let me know

vvo:
Aside from small comment, LGTM

package.json
samouss: I use it when I create a build / run test to be sure to run those scripts in a clean environment (no install / build cache). The option provided remove **only** the files that are ignored in your configuration (`.gitignore`). Untracked files are untouched.

PR #1098, opened 2018-03-21
Update Getting_started.md

iamclaytonray:
Just adds a space


Haroenv:
Thanks for this PR @iamclaytonray, unfortunately this doesn’t seem to fix it, maybe it needs to be in a `<p>`? Let’s think about a proper solution

iamclaytonray:
Ahh, yeah, I see that. Wouldn't a `<p />` drop it to the next line, though? Is ASCII supported in markdown? I could probably do a `&nbsp;`

Haroenv:
The issue was coming up because the container has display flex, all other times something like this was used, the actual content was in another tag, I've updated this to be consistent now. Thanks for noticing this issue!

Haroenv:
done @samouss, I thought it was already fixed based on a quick search :)

PR #1104, opened 2018-03-22
chore(travis): trigger build on every PR

samouss:
**Summary**

By removing this option and disabling the build on push on the Travis UI we are able to run the build only once on **every** PR. Even on PR that don't target `master`. Previously the build wasn't trigger on a PR that didn't target `master`.

screen shot 2018-03-22 at 16 32 38

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,23 @@
 language: node_js
+
 before_script:
   - yarn boot
-branches:
-  only:
-    - master
+
 addons:
   firefox: 56.0
+
 before_install:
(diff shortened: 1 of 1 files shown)


vvo:
I think I remember why it was like that, maybe to allow tools like renovate/greenkeeper to just push branches and have their tests ran, unsure still, will see :)

samouss:
Ok, but Renovate can't push directly on `master` (for this project) it will create a PR every time so even without this configuration the tests will be run.

PR #1116, opened 2018-03-27
chore: enable workspaces on docgen

samouss:
**Summary**

This is a third step into moving the repo to a proper structure (with Workspaces). This PR enables the Workspaces on the `docgen` package with a few other changes:

- move all the dependencies related to `docgen` on his own `package.json`
- set a default env for `docgen`
- scope the build artefacts under docgen:
  - `/docs` → `/docgen/docs-development`
  - `/docs-production` → `/docgen/docs-production`

I also removed the cache for Yarn on Travis. It should not change anything because the previous command `boot` was removing the cache folder before installing the dependencies. The install is fastest than before but wan can re-enable it if it's too slow. I'm not a huge fan of caching dependencies between build I prefer to rebuild everything from scratch every time especially for tests / builds. 

https://github.com/algolia/react-instantsearch/blob/60f648e26e1dd859bac0749fb689f289387ef1ca/scripts/boot.sh#L7

**Nest steps**

- ~~move the main package to Workspaces~~
- ~~move the docs to Workspaces~~


.travis.yml
```diff
@@ -12,12 +12,6 @@ before_install:
 install:
   - yarn bootstrap
 
-cache:
```
vvo: If you don't want to use the cache yarn: true then maybe add a comment saying "We explicitly disable this cache because X" so that no-one is tempted to add it back

vvo:
> I'm not a huge fan of caching dependencies between build I prefer to rebuild everything from scratch every time especially for tests / builds.

Do you know the actual build time difference?
Did you had any issues with cache yarn: true maybe?

I am convinced this works as expected but if you prefer without it, good too

vvo:
The tricky part about removing every packages and redoing installation on Travis is that then you have a difference between your dev environment and Travis. In dev when you will release/publish the doc you won't be removing all node_modules, you'll just trust "yarn" to clean everything. So why not doing it too on Travis?

Otherwise maybe do the same locally and remove node_modules when releasing but that's seems tricky to do so too and slow.

samouss:
> Do you know the actual build time difference?

Should be pretty much the same, but if we see a huge drop in performance we can add it back.

> Did you had any issues with cache yarn: true maybe?

Nope, but I probably misunderstood something. I'll re-enable the `yarn: true` but I won't keep the other cache directives. It means that we will have exactly the same environment between our local machine and Travis. WDYT?

> In dev when you will release/publish the doc you won't be removing all node_modules, you'll just trust "yarn" to clean everything. So why not doing it too on Travis?

I'm not sure to fully understand. Before releasing the doc you will have to redo the `bootstrap` to be sure to have a clean project and the dependencies up to date (I always clean everything before a deploy / release). So we should have the same environment as Travis. Maybe I miss something?

Haroenv:
I think React InstantSearch is the only library in which we have a clean/bootstrap step, it's probably not that necessary IMO

vvo:
> It means that we will have exactly the same environment between our local machine and Travis. WDYT?

Yep!

> Before releasing the doc you will have to redo the bootstrap to be sure to have a clean project and the dependencies up to date (I always clean everything before a deploy / release). So we should have the same environment as Travis. Maybe I miss something?

The difference if you disable cache on Travis with local will be:

locally:
node_modules is present
bootstrap runs
yarn runs

on travis:
node_modules is absent
bootstrap runs
yarn runs
node_modules is installed

cache:
	yarn: true

On travis means it will download the last node_modules and add it before running "yarn" that's it. That's why it's closer to what you have locally maybe.

No matter what I think we should rm -rf node_modules before running yarn only if we know there are actual issues, today I believe we did not encounter that but let me know.

samouss:
Finally I removed the `bootstrap` script in favor of just `yarn`. I also restore the `yarn` cache directive for Travis, but only the default one.

samouss:
- Before: `yarn boot` ~145s
- Now: `yarn` ~50s

vvo:
NICE!

vvo:
LGTM

PR #1118, opened 2018-03-27
fix: use lodash version of unsupported Array.{fill, find}

### LINKED ISSUE #1115: IE 11 doesn't support Array.find()
:wave: Hi team

IE 11 is throwing an error when I try to test on my local development server because of the use of Array.find(). I suspect it's the use here: https://github.com/algolia/react-instantsearch/blob/1a164d5536c82253b96619f4d2ba8d6964ae0992/packages/react-instantsearch/src/core/createConnector.js#L85-L93

which is only used when `process.env.NODE_ENV === 'development'`.

For now I'm just polyfilling it myself, but supporting the same browsers in production and development seems like the best solution. What do you say? 🙂

samouss:
**Summary**

Close #1115

### DIFF
--- a/packages/react-instantsearch/src/components/RatingMenu.js
+++ b/packages/react-instantsearch/src/components/RatingMenu.js
@@ -1,3 +1,4 @@
+import { find } from 'lodash';
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
@@ -147,7 +148,7 @@ class RatingMenu extends Component {
     const range = new Array(safeInclusiveLength)
       .fill(null)
       .map((_, index) => {
-        const element = values.find(item => item.value === limitMax - index);
+        const element = find(values, item => item.value === limitMax - index);
         const placeholder = { value: limitMax - index, count: 0, total: 0 };
 
         return element || placeholder;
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch/src/components/RatingMenu.js
```diff
@@ -1,3 +1,4 @@
+import { find, fill } from 'lodash';
```
Haroenv: probably best to depend on /find and /fill ?

samouss: The `babel-rollup-plugin` handle that for us.

Haroenv: didn't know we used it 🙌

vvo:
In the future if we want to get rid of lodash, maybe we can use the automatic polyfill features of babel? Any of you have experience using https://babeljs.io/docs/plugins/transform-runtime?

Haroenv:
Would be better to use core-js directly rather than babel, I think that seems to be more of the recommendation. I fear that core-js is actually bigger than loads usually too

samouss:
`babel-runtime` use `core-js` under the hood. I think it's the same at the end.

PR #1120, opened 2018-03-28
docs: update examples for RatingMenu & RefinementList

samouss:


samouss:
Yep it's completely that!

PR #1134, opened 2018-04-03
fix(release): install all the yarn dependencies

samouss:
**Summary**

The `release` script is trigger with `NODE_ENV` set to `production`. Yarn will install by default only the production dependencies. Previously the install was done with the `yarn boot` command, the `NODE_ENV` was override inside the script (see below). This PR fixes the issue by providing the option `--production=false` to yarn.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -41,7 +41,7 @@ git pull origin master
 git fetch origin --tags
 
 # printf "Release: install dependencies"
-yarn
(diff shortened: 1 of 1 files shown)


vvo:
Can you add a small comment to the code? This is the kind of lines that's not easy to grasp just by reading it, wdyt?

samouss:
@vvo Done.

PR #1135, opened 2018-04-03
fix: revert dependencies to devDependencies

samouss:
**Summary**

In the move to workspaces I accidentally move React as dependencies. It means that when you install React InstantSearch you also install React, React DOM & React Native. This PR revert this change, in a next version we should add those dependencies as peer.

### DIFF
--- a/packages/react-instantsearch/package.json
+++ b/packages/react-instantsearch/package.json
@@ -33,10 +33,7 @@
     "algoliasearch-helper": "^2.21.0",
     "classnames": "^2.2.5",
     "lodash": "^4.17.4",
-    "prop-types": "^15.5.10",
-    "react": "^16.2.0",
-    "react-dom": "^16.2.0",
-    "react-native": "^0.54.2"
+    "prop-types": "^15.5.10"
   },
   "devDependencies": {
     "babel-cli": "6.26.0",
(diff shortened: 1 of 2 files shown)


Haroenv:
why not move to peer dependency now?

samouss:
I'm not sure which version to set. I prefer to revert the changes and move on a proper `peerDependencies` after. When the next package architecture will be ready, we can test on which version the lib works and which version we can support.

packages/react-instantsearch/package.json
```diff
@@ -33,10 +33,7 @@
     "algoliasearch-helper": "^2.21.0",
     "classnames": "^2.2.5",
     "lodash": "^4.17.4",
-    "prop-types": "^15.5.10",
-    "react": "^16.2.0",
-    "react-dom": "^16.2.0",
-    "react-native": "^0.54.2"
+    "prop-types": "^15.5.10"
```
vvo: Are prop types used in production builds?

Haroenv: we can't strip them out in the library itself, since that way they'd only be useful for ourselves. People would strip them out at their bundling, not our I think

samouss: In production build they are still present yes. But they are are used only in development. You can strip them with a babel plugin as @Haroenv suggest.

PR #1142, opened 2018-04-04
docs(Conditional): fix snippet for <AllResults />

### LINKED ISSUE #1140: Documentation for conditional display with multiple indices appears to have a bug in it.
Hello! 

Thank you so much for this project as it has been an immense help!

I believe that while reading the documentation on how to implement a no search results found page (https://community.algolia.com/react-instantsearch/guide/Conditional_display.html) with multiple indices I have found a bug in the code snippet provided in the documentation.

The reduce function will return true if the last expression evaluates to `true`.  This means if you have a search result hit in the the first index, but not in the last index the function in the documentation will return true, and hide search results event though there are some results. 

Locally I have just changed the `reduce` to `every` and it seems to work fine.

`const array1 = [false, false, false, true];`
`const reducer = (accumulator, currentValue) => currentValue;`
`// Will output true`
`console.log(array1.reduce(reducer, false));`

Here is the screenshot for the code in question.  Please let me know if I'm missing something or any clarification is needed.

https://community.algolia.com/react-instantsearch/guide/Conditional_display.html
screen shot 2018-04-03 at 12 48 28 pm

samouss:
Close #1140

### DIFF
--- a/docgen/src/guide/Conditional_display.md
+++ b/docgen/src/guide/Conditional_display.md
@@ -119,13 +119,11 @@ const IndexResults = connectStateResults(
 );
 
 const AllResults = connectStateResults(({ allSearchResults, children }) => {
-  const noResults =
+  const hasResults =
     allSearchResults &&
-    Object.values(allSearchResults).reduce(
-      (acc, results) => results.nbHits === 0,
-      false
-    );
-  return noResults ? (
+    Object.values(allSearchResults).some(results => results.nbHits !== 0);
+
+  return !hasResults ? (
     <div>
       <div>No results in category, products or brand</div>
       <Index indexName="first" />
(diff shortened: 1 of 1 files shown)


docgen/src/guide/Conditional_display.md
```diff
@@ -119,13 +119,11 @@ const IndexResults = connectStateResults(
...
 const AllResults = connectStateResults(({ allSearchResults, children }) => {
-  const noResults =
+  const hasResults =
     allSearchResults &&
-    Object.values(allSearchResults).reduce(
-      (acc, results) => results.nbHits === 0,
-      false
-    );
-  return noResults ? (
+    Object.values(allSearchResults).some(results => results.nbHits !== 0);
```
Haroenv: Would've written `> 0` personally, seems clearer

samouss: Yep makes more sense the intent is a bit clearer 👍

PR #1145, opened 2018-04-05
docs: add requirements on faceting for range

samouss:
**Summary**

Inside the documentation every widgets that is related to the `connectRange` connector has a section "requirements" about the fact that the value must be a number. But we don't precise that the value must be also declared as `attributeForFaceting`. This PR adds those requirements in:

- `connectRange`
- `RangeInput`
- `RatingMenu`


samouss:
@Haroenv yep! Maybe some connectors & widgets have still missing requirements but this PR is only for range related components.

PR #1147, opened 2018-04-06
feat(createConnector): add uiState

samouss:
**Summary**

This PR adds to `createConnector` the ability to save UI related state and trigger re-render without triggering a search (will be used for the geo search connector). Here is the lifecycle of a connector when the `uiState` is provided:

- **mount** → `getInitialUiState` → `getProvidedProps` → `render`
- **setUiState** → `uiStateUpdater` → `getProvidedProps` → `render`

The implementation didn't change the current behaviour when the `uiState` is not provided. 

The concept of getting new props from a connector remains the same. The only way to get them is to return them from `getProvidedProps`. The introduction of `uiState` doesn't change this rule. We pass the `uiState` to `getProvidedProps` in order to get them. We should not use `uiState` & `setUiState` in a other place than the view. Hence the new parameters are only available in `getProvidedProps`.

I don't think we need to make this API public, because a user could implement his own logic outside of the connector in a parent component.

**API**:

- **getInitialState**: `(props) => uiState`
- **setUiState**: `(uiStateUpdater) => void`
- **uiStateUpdater**: `(prevUiState) => nextSliceUiState`

**Usage**

```js
const connect = createConnector({
  getInitialUiState: props => ({
    isMenuOpen: props.isOpen
  }),

  getProvidedProps(_, _, _, _, _, uiState, setUiState) {
    return {
      ...uiState,
      toggleMenu: () =>
        setUiState(prevUiState => ({
          isMenuOpen: !prevUiState.isMenuOpen
        }))
    };
  }
});

const Header = connect(({ isMenuOpen, toggleMenu }) => (
  <header>
    <button onClick={toggleMenu}>Toggle</button>

    {isMenuOpen && (
      <nav>
        <Link />
        <Link />
        <Link />
      </nav>
    )}
  </header>
));
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,7 @@
 module.exports = {
   extends: ['algolia/react', 'algolia/jest'],
   rules: {
+    'max-params': ['error', 7],
     'no-param-reassign': 'off',
     'import/no-extraneous-dependencies': [
       'error',
--- a/packages/react-instantsearch/src/core/createConnector.js
+++ b/packages/react-instantsearch/src/core/createConnector.js
@@ -269,7 +269,8 @@ export default function createConnector(connectorDesc) {
           searchResults,
           metadata,
           resultsFacetValues,
-          uiState
+          uiState,
+          this.setUiState
         );
       };
 
@@ -305,6 +306,28 @@ export default function createConnector(connectorDesc) {
         );
 
       cleanUp = (...args) => connectorDesc.cleanUp.call(this, ...args);
+
+      setUiState = uiStateUpdater =>
+        this.setState(prevState => {
+          const nextSliceUiState = uiStateUpdater(prevState.uiState);
+
+          if (!nextSliceUiState) {
+            return null;
+          }
+
+          const nextUiState = {
+            ...prevState.uiState,
+            ...nextSliceUiState,
+          };
+
+          return {
+            uiState: nextUiState,
+            props: this.getProvidedProps({
+              props: prevState.props,
+              uiState: nextUiState,
+            }),
+          };
+        });
 
       render() {
         if (this.state.props === null) {
--- a/packages/react-instantsearch/src/core/createConnector.js
+++ b/packages/react-instantsearch/src/core/createConnector.js
@@ -181,13 +181,18 @@ export default function createConnector(connectorDesc) {
         }
       }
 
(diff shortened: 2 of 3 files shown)


bobylito:
As far as I can tell, it looks good 👍 A second opinion might be good 😅

.eslintrc.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   extends: ['algolia/react', 'algolia/jest'],
   rules: {
+    'max-params': ['error', 7],
```
vvo: YOLO! For a next major I bet you would want to move over to an object param maybe?

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -63,24 +63,41 @@ export default function createConnector(connectorDesc) {
...
+              canRender,
+            },
+          }),
           canRender, // use to know if a component is rendered (browser), or not (server).
         };
 
         this.unsubscribe = store.subscribe(() => {
           if (this.state.canRender) {
-            this.setState({
+            this.setState(prevState => ({
```
vvo: is this.state strictly equivalent to prevState here? Just wondering if this can break in any way

Haroenv: this.state is actually the wrong way of doing it, since `setState` calls can be batched, and so it could be possible that `this.state` is the value from a previous render (especially when async rendering will be enabled)

.eslintrc.js
samouss: Clearly... I tried to find a path to easily migrate on named parameters (with `fn.length`) but without renaming the function I didn't found an easy way.

packages/react-instantsearch/src/core/createConnector.js
samouss: No it won't be strictly equivalent in some cases. But it should not be a problem, because the `canRender` value change only once (when the component `didMount`). When the condition is `true` we are sure that the `prevState` value is correct since it's the one that is the most up to date.

samouss:
This implementation won't solve all the issues that we have today. For example we have no way to update properly the `uiState` from the connector without triggering a warning in React (setState will not be pure anymore). It means that we can't update the state on an external change. I'll close this PR. In the next implementation we should stick as much possible to the React API (setState, lifecycle, etc..).

PR #1149, opened 2018-04-06
refactor(createInstantSearchServer): isolate test from each others

samouss:
**Summary**

There are two tests that were uncaught in case of error in the tests suite (see below). But they also shared some global state. In the PR I removed the global share state and make each test isolate from each others & fix the uncaught ones.

https://github.com/algolia/react-instantsearch/blob/c862cddca5f92ce89e6da8694301b6d2d5c2bc70/packages/react-instantsearch/src/core/createInstantSearchServer.test.js#L74-L80

https://github.com/algolia/react-instantsearch/blob/c862cddca5f92ce89e6da8694301b6d2d5c2bc70/packages/react-instantsearch/src/core/createInstantSearchServer.test.js#L162-L168


vvo:
Can't say LGTM because that's an advanced test code change that I don't master. I trust you though to do it right

Haroenv:
same as vincent

PR #1150, opened 2018-04-07
Resolved bug where connectRange connector fails to return results for queries with a max < 1

msied:
**Summary**

When calling `refine` through the `connectRange` connector using a max value below 1, the query is made properly and the expected results are returned, but immediate subsequent refinement queries are then called (not through this connector's refine method) that coerce the max and min values into integers, irregardless of the specified precision attribute, and thus return inaccurate results. This is problematic whenever precision is used, but discovered in my use case wherein the desired data contains an attribute with values between 0.1 and 0.9 (this queries for values between 0 and 0, which will result in nothing for my dataset). A slider component using `connectRange` was returning these erroneous results.

This PR resolves the issue by modifying the `connectRange` code to use `parseFloat` rather than `parseInt` within the `getCurrentRefinement` method.

**Result**

For my dataset, when I query for an attribute with values between 0.1 and 0.9, the correct results appear and are maintained even as other filters are applied with other widgets.

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectRange.js
+++ b/packages/react-instantsearch/src/connectors/connectRange.js
@@ -73,14 +73,27 @@ function getCurrentRefinement(props, searchState, currentRange, context) {
     `${namespace}.${getId(props)}`,
     {},
     currentRefinement => {
-      let { min, max } = currentRefinement;
-      if (typeof min === 'string') {
-        min = !props.precision ? parseInt(min, 10) : parseFloat(min);
+      const { min, max } = currentRefinement;
+      const isFloatPrecision = Boolean(props.precision);
+
+      let nextMin = min;
+      if (typeof nextMin === 'string') {
+        nextMin = isFloatPrecision
+          ? parseFloat(nextMin)
+          : parseInt(nextMin, 10);
       }
-      if (typeof max === 'string') {
-        max = !props.precision ? parseInt(max, 10) : parseFloat(max);
+
+      let nextMax = max;
+      if (typeof nextMax === 'string') {
+        nextMax = isFloatPrecision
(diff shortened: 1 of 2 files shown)


Haroenv:
This looks correct, but I’ll let @samouss take another look at this. Thanks a lot for contributing

samouss:
Hi @msied, thanks a lot for the contribution! 

I didn't manage to reproduce the issue and I don't understand how it's possible that those values are retrieved as strings. Because we always set them as numbers in the `searchState`. Could you provide a reproducible example? It will help us to find a potential another issue. We provide template to avoid you the boilerplate part. You can find in it on [CodeSandbox](https://codesandbox.io/s/github/algolia/instantsearch-templates/tree/master/src/React%20InstantSearch). 

Thanks!

msied:
@samouss I looked into this further and found that searchState was being pulled in from query parameters in the URL (as to persist a linked search). This means that the issue can be resolved without this PR. However, the `connectRange` connector does appear to try to attempt to handle handle string values for min and max (as seen in the lines changed in this diff), irrespective of the statement that "we always set them as numbers in the `searchState". `parseInt` remains the wrong implementation for this in the existing code if the precision attribute is to be applied, as it coerces the number to an integer when a string value is provided.

samouss:
Hi @msied thanks a lot for the clarification! What I meant with "we always set them as numbers in the `searchState`" is that internally the connector always set value as number in the `searchState` (with the `refine` function). But in your case the connector is not in charge of putting those values in the `searchState`. 

At the end your are right it's an issue and we need to fix it. Do you want to add a test case to be sure that we won't have regressions on that case? If not I will do it no worries.

PR #1171, opened 2018-04-18
feat(geo): add connector [PART-1]

samouss:
This is the first part of the work for the GeoSearch widget. The second part will be the widget itself implemented (with Google Maps) with this connector. Most of the API is fixed but I will update the PR if something is missing during the widget implementation.

At the end in this version we don't support all the APIs that InstantSearch support. It's because with the current implementation of React InstantSearch it's very complicated to update the internal state of the widget from the outside. So we only manage the state of the search in the connector.

**Issue:**

The part that might change is the one where we compute the current position. This position is used (mainly) to move the map to the correct bounds when we don't have results. It might change because with the current implementation we can only read the position from the path `aroundLatLng` in the `searchState`. It's perfect for a custom widget (like Places for example), the widget needs to write the position at this path (an example will be built in Storybook). But when the position is set by `Configure` we need to look at a different path `configure.aroundLatLng`. We have some solutions:

- allow to configure the path where we want to find the value as prop
- search at multiple paths stop on the first match `aroundLatLng` → `configure.aroundLatLng`
- implement the position, radius & precision as props of the connector (like InstantSearch). This solution remove the needs to look at different paths, we can only keep the top level one

**Update**: 

- I updated the PR to be able to override the `shouldComponentUpdate` hook of `createConnector` from the connector definition. To enable a full customisation we should be able to re-render the content of the marker. But the `createConnector` prevent this update based on his props & state. Now we can bypass this limitation by implementing the `shouldUpdate` function on the connector.

### DIFF
--- a/docgen/src/stylesheets/components/_documentation.sass
+++ b/docgen/src/stylesheets/components/_documentation.sass
@@ -3,7 +3,7 @@ $offset-height: 60px
 
 .documentation-section,
 .examples-section
-  padding-bottom: 300px
+  padding-bottom: 320px
 
   .container
     article
@@ -136,7 +136,6 @@ $offset-height: 60px
       background: #fff
       border: 1px solid #d8d8d8
       border-radius: 2px 2px 0 0
-      margin: -8px 0 0
       padding: 0.75em 1em
       font-family: $paragraphs-font-family
       border-radius: 6px 6px 0 0
@@ -201,6 +200,10 @@ $offset-height: 60px
         display: block
         height: $offset-height
         // margin: (-$offset-height) 0 0
+
+    .sub-component-title
+      &:before
+        height: 10px
 
     .anchor
       margin-left: .2em
--- a/docgen/src/widgets/GeoSearch.md
+++ b/docgen/src/widgets/GeoSearch.md
@@ -1,0 +1,735 @@
+---
+mainTitle: Widgets
+title: GeoSearch
+layout: widget.pug
+category: widget
+showInNav: true
+navWeight: 0
+external: true
+---
+
+## Description
+
+The `GeoSearch` widget displays the list of results from the search on a Google Maps. It also provides a way to search for results based on their position. The widget provides some of the common GeoSearch patterns like search on map interaction.
+
+<div class="storybook-section">
+  <a class="btn btn-cta" href="https://community.algolia.com/react-instantsearch/storybook?selectedKind=GeoSearch&selectedStory=default" target="_blank">
+    See live example
+  </a>
+</div>
+
+## Requirements
+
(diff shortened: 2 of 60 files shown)


packages/react-instantsearch/src/connectors/connectGeoSearch.js
```diff
@@ -0,0 +1,138 @@
...
+} from '../core/indexUtils';
+
+/**
+ * The GeoSearch connector provides the logic to build a widget that will display the results on a map.
+ * It also provides a way to search for results based on their position. The connector provides function to manage the search experience (search on map interaction).
+ * @name connectGeoSearch
+ * @kind connector
+ * @requirements Note that the GeoSearch connector uses the [geosearch](https://www.algolia.com/doc/guides/searching/geo-search) capabilities of Algolia.
+ * Your hits **must** have a `_geoloc` attribute in order to be passed to the rendering function.
+ * Currently, the feature is not compatible with multiple values in the `_geoloc` attribute.
```
vvo: Maybe add: "Example: a restaurant in multiple locations". If you are in this case, you can deduplicate records and use distinct (link to doc). wdyt?

packages/react-instantsearch/src/connectors/connectGeoSearch.js
```diff
@@ -0,0 +1,228 @@
...
+      lng: parseFloat(values[1]),
+    },
+    southWest: {
+      lat: parseFloat(values[2]),
+      lng: parseFloat(values[3]),
+    },
+  };
+};
+
+const latLngRegExp = /^(-?\d+(?:\.\d+)?),\s*(-?\d+(?:\.\d+)?)$/;
```
vvo: Just checking, why the need for a different algorithm to parse the lat lng (regex vs split)?

packages/react-instantsearch/src/connectors/connectGeoSearch.js
```diff
@@ -0,0 +1,228 @@
...
+    const results = getResults(searchResults, this.context);
+
+    // We read it from both becuase the SearchParameters & the searchState are not always
+    // in sync. When we set the refinement the searchState is used when we clear the refinement
+    // the SearchParameters is used. In the first case when we render the results are not there
+    // so we can't find the value from the results. The most up to date value is the searchState.
+    // But whren we clear the refinement the searchState is immediatly clear even when the items
+    // retrieve are still the one from the previous query with the bounding box. It leads to some
+    // issue with the position of the map. We should rely on 1 source of truth or at least always
+    // be sync.
```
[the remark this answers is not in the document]
samouss: Yes, in all other connectors we never rely on the helper state only on the `searchState`. In that case we can't because both states are not always sync. Most of the time it's not an issue but in case of the geo search it leads to inconsistencies for the view.

PR #1214, opened 2018-05-04
feat(geo): custom Marker [PART-7]

samouss:
**Summary**

This is the seventh part of the GeoSearch widget. The PR introduces the ability to create custom marker on the map. The marker accepts any React elements. We use the `createPortal` feature of React 16.x but when it's not available we use the `unstable_renderSubtreeIntoContainer` that is shipped with React 15.x. For now we don't support any events directly on the marker, it will come next.

untitled

This PR includes:

- **`createHTMLMarker`**: class that creates the marker from the `OverlayView` of Google Maps. The implementation is bit different than the one in InstantSearch, I will probably apply the changes from this implementation to the other. At some point we can externalise this code in a external package.
- **`<CustomMarker>`**: the component that will wrap the `HTMLMarker` class into a React component (like `<Marker>` component already does). 

**Usage**

```jsx
import { GoogleMapsLoader, GeoSearch, CustomMarker } from "react-instantsearch-dom-geo"; // name will change

<GoogleMapsLoader apiKey="GOOGLE_MAPS_API_KEY">
  {google => (
    <GeoSearch google={google}>
      {({ hits }) => (
        <Fragment>
          {hits.map(hit => (
            <CustomMarker key={hit.objectID} hit={hit} className="marker-element">
              <span>{hit.price}</span>
            </CustomMarker>
          ))}
        </Fragment>
      )}
    </GeoSearch>
  )}
</GoogleMapsLoader>
```

**Result**

You can use the widget in [Storybook](https://deploy-preview-1214--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=GeoSearch&selectedStory=with%20%3CHTMLMarker%3E&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs).


vvo:
Naming iteration/idea: `CustomMarker`? HTMLMarker even if that how it's implemented might gives the impression it's really only HTML, while this is still JSX

samouss:
Indeed in InstantSearch we use `customHTMLMarker`. We can go for `CustomMarker`, any other ideas?

Haroenv:
For the package name I still think description + platform is better than main + platform + subthing

So my idea would be `React-InstantSearch-maps-web` or as second choice `-geo-Dom` Or a permutation

PR #1215, opened 2018-05-04
feat: Upgrade Client and Helper

francoischalifour:
This PR upgrades to `algoliasearch@3.27.1` and `algoliasearch-helper@2.26.0`, using the promisified version of `client.search()`. It is necessary for adding `searchClient` support to React InstantSearch.

I had to update the tests with a not-so-great hack because of the way we internally handle search with promises now:

```javascript
Promise.resolve()
  .then(() => {})
  .then(() => {
    // test...
  });
```

### DIFF
--- a/packages/react-instantsearch/src/core/createInstantSearchManager.derived.test.js
+++ b/packages/react-instantsearch/src/core/createInstantSearchManager.derived.test.js
@@ -22,22 +22,7 @@ jest.mock('algoliasearch-helper/src/algoliasearch.helper.js', () => {
 });
 
 const client = algoliaClient('latency', '249078a3d4337a8231f1665ec5a44966');
-client.addAlgoliaAgent = () => {};
-client.search = jest.fn((queries, cb) => {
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch/src/core/createInstantSearchManager.derived.test.js
```diff
@@ -76,46 +61,46 @@ describe('createInstantSearchManager', () => {
...
-            params.setQuery('second query 2');
-
-          ism.widgetsManager.update();
-
-          return Promise.resolve().then(() => {
-            jest.runAllTimers();
-
-            const store1 = ism.store.getState();
+        return Promise.resolve()
+          .then(() => {})
```
Haroenv: I wonder if “await await” is allowed? I think a comment here would be useful too

samouss: Agree for comment!

PR #1258, opened 2018-05-23
chore(packages): create core package [PART-1]

### LINKED ISSUE #229: [dev env] convert to a proper monorepo structure
Today we're using the traditional Lerna file structure, where we have a `packages` folder, in which multiple `package.json`s are used. We have our own proprietary `yarn boot` script, which links up these dependencies manually. 

We want to switch to a library to handle this for us, so we don't have issues with that anymore (see https://github.com/algolia/react-instantsearch/commit/a80dcd5db3a4c9d595a83c63aa6b190927084b3d).

There are two ways to go now: 

- switch to Lerna now
  - maybe have the problems we had earlier combining it with Yarn (cc @rayrutjes)
- switch to Yarn workspaces soon
  - not yet released, should definitely work with Yarn 😂 

cc @rarkins

samouss:
**Summary**

Close #229 

This is the first part for the split of the package into multiple ones. This PR create the `core` package that contain all the logic to create a new React binding for InstantSearch. In the next PR we will create two bindings the `DOM` one and the `Native` one. 

This PR contains:

- the `core` package that contain all the core logic for React InstantSearch (manager, store, etc...), factories for `InstantSearch` & `Index`, the connectors and the widgets without a rendering (only `Configure` at the moment).
- uses the `core` package as dependency for the main React InstantSearch package. All the "core" imports are delegate to the `core` package.
- setup of Lerna (we only use a subset of the features available for now)
- move the tests in `__tests__` folder near by the implementation

**Note**:

- The packages are built in a standalone spirit, it means that they can be build and run independently. Right now we have some duplication (`.babelrc`, `rollup.config.js`). We can change this pattern later using a common base for those files. Since we won't have 100 modules under packages this might never happen though.
- You can see in the PR that we have a special entry point for one of the UMD file of the main React InstantSearch package. It's because Rollup is not able to correctly tree shake the bundle. For this reason we use the "private" import to build the exact same file for compatibility reason. Otherwise the file size will have a huge jump 8KB => 40KB.
- You can see that we also update the size of the `connectors` UMD file of the main React InstantSearch package from 38KB => 40KB. It's because the connectors are now import from another package that is already transpiled (with Babel). So each file will have his own Babel helpers instead of one for the all bundle when Rollup is doing the transpilation. The files impacted are only the UMD. We can probably avoid this but I didn't looked that much since the UMD files are rarely used with React (we do not document this usage).


packages/react-instantsearch-core/src/connectors/__tests__/connectPoweredBy.js
```diff
@@ -1,6 +1,6 @@
-import connect from './connectPoweredBy';
+import connect from '../connectPoweredBy';
 
-jest.mock('../core/createConnector');
+jest.mock('../../core/createConnector', () => x => x);
```
Haroenv: where's x defined?

samouss: Nowhere it's a function that returns a function.

Haroenv: should it not be `() => () => undefined` or `() => () => {}` then?

samouss: Nope because we use the return value of the connector for the test. The function above takes the connector declaration (the object with `getProvidedProps`, etc...) and return it. This way we can test each function independently. It was already the case, I inline the mock instead of rely on the `__mock__` file.

packages/react-instantsearch-core/.babelrc
```diff
@@ -0,0 +1,85 @@
...
+      ]
+    },
+    "es": {
+      "presets": [
+        [
+          "env",
+          {
+            "targets": {
+              "browsers": [
+                "last 2 versions",
```
vvo: I read a nice article about the use of last 2 versions: https://jamie.build/last-2-versions

is this relevant for us too?

samouss: It's a tricky question 🙂 I think we can revamp our range when we are sure which browsers we want to support. But for example we won't drop the IE support (soon).

PR #1259, opened 2018-05-23
chore(packages): create DOM package [PART-2]

samouss:
**Summary**

This is the second part for the split of the package. This PR create the `DOM` package that contains the React DOM binding for InstantSearch. The package expose all the DOM widgets available but also the connectors and other API. The previous package still works to avoid a breaking change compatibility but we can use this one instead.

```js
// Before
import { createConnector } from 'react-instantsearch';
import { InstantSearch, SearchBox } from 'react-instantsearch/dom';
import { connectMenu } from 'react-instantsearch/connectors';
import { createInstantSearch } from 'react-instantsearch/server';

// After
import { createConnector, InstantSearch, SearchBox, connectMenu } from 'react-instantsearch-dom';
import { createInstantSearch } from 'react-instantsearch-dom/server';
```

With the previous import we have to use a different import in order to use the ES Modules files to take advantage of tree shaking. With something like `react-instantsearch/es/dom`. Now it's transparent for the users since we only have the top level import it's up to the bundler to pick [the correct format](https://webpack.js.org/configuration/resolve/#resolve-mainfields) (`main` or `module`). It was not possible with the previous import because of the relative path.

|    |  `xxx/dom` (CJS) | `xxx/dom` (CJS + ESM) | `xxx-dom` (ESM)
| - | :---: | :---: | :---: |
| **CRA + Webpack 3** | 108 kB  | 105 kB (-2.7%)  | 105 kB (-2.7%)
| **Webpack 4** | 104 kB  | 100 kB (-3.8%)  | 87 kB (-16.3%)

> The sizes are after gzip. The sample application use: Configure, SearchBox, ClearRefinements, RefinementList, Hits, Custom Hits (with connectHits), Pagination.

This PR contains:

- the `dom` package that exports all the binding for the DOM (widgets), widget without render (`Configure`), the connectors and the public API (`createConnector`).
- uses the `dom` package as dependency for the main React InstantSearch package. All the "dom" imports are delegate to the `dom` package.
- move the tests in `__tests__` folder near by the implementation


Haroenv:
If both connectors and widgets are in the same package, same import; does that mean that they all get included? the size difference looks smaller than I expected. For example some bigger widgets like the Hierarchical isn't used

samouss:
Nope they are not included when they are unused.

**Before**

image

**After**

image

PR #1260, opened 2018-05-23
chore(packages): create native package [PART-3]

samouss:
**Summary**

This is the third part for the split of the package. This PR create the `native` package that contains the React Native binding for InstantSearch. The package expose the connectors and other API since we don't have any built-in widgets for React Native. The previous package still works to avoid a breaking change compatibility but we can use this one instead.

```js
// Before
import { createConnector } from 'react-instantsearch';
import { InstantSearch } from 'react-instantsearch/native';
import { connectMenu } from 'react-instantsearch/connectors';

// After
import { createConnector, InstantSearch, connectMenu } from 'react-instantsearch-native';
```

This PR contains:

- the `native` package that exports all the binding for React Native (widgets), widget without render (`Configure`), the connectors and the public API (`createConnector`).
- uses the `native` package as dependency for the main React InstantSearch package. All the "native" imports are delegate to the `native` package.


packages/react-instantsearch-native/.babelrc
```diff
@@ -0,0 +1,85 @@
+{
+  "env": {
```
vvo: Since we have the same .babelrc file everywhere, should we make it a .babelrc.js and have a single one that is required from those files? May or may not be relevant.

Haroenv: we can only do .js in v7 of babel I think

samouss: Indeed it's only available in Babel 7 (still in beta). But sill it depends if we want to keep the spirit of self-contained package. We can move the `.babelrc` at the top level (for Babel & Rollup at least) but it means that the child packages will depend on the top level one. For now I choose the most simplest way to implement it, the duplication. It's only the first iteration, we can choose to move all the configuration at the top in the next iteration. WDYT?

Haroenv: I'd even prefer to change it to a shared _actual_ dependency like `kcd-scripts`, it can be in the same monorepo, that way it avoids any issues with duplication. WDYT?

samouss: Yes indeed might be a solution.

PR #1261, opened 2018-05-23
chore(packages): update legacy build [PART-4]

samouss:
**Summary**

This is the fourth part for the split of the package. This PR update the build step for the "legacy" package. It use the same scripts than the other packages to be built with Lerna.

This PR contains:

- update the scripts for the "legacy" package. It doesn't rely on bash or external scripts anymore
- remove useless `production` key in the `.babelrc`


docgen/.babelrc
```diff
@@ -0,0 +1,56 @@
...
+              ]
+            }
+          }
+        ],
+        "react",
+        "stage-2"
+      ],
+      "plugins": [
+        "lodash",
+        "inline-json-import",
```
Haroenv: This is the plugin that accidentally gives the whole json object for just the version

packages/react-instantsearch-native/.babelrc
```diff
@@ -48,7 +48,7 @@
             "targets": {
               "browsers": [
                 "last 2 versions",
-                "ie >= 10"
+                "ie >= 9"
```
Haroenv: why this one 9 specific? I assume RN uses JSC, which has far above IE9 capabilities (even has modules IIRC)

docgen/.babelrc
samouss: Yes I removed it in the others `.babelrc`. I kept it in `docgen` because we require the whole `package.json` for the configuration.

.babelrc
```diff
@@ -1,122 +1,64 @@
...
-            ],
-          },
-        }],
+        [
+          "env",
+          {
+            "targets": {
+              "browsers": [
+                "last 2 versions",
+                "ie >= 9"
```
vvo: new packages are ie>=10, is this an issue?

packages/react-instantsearch/rollup.config.js
```diff
@@ -11,23 +11,19 @@ const clear = x => x.filter(Boolean);
 const version = process.env.VERSION || 'UNRELEASED';
 const algolia = '© Algolia, inc.';
 const link = 'https://community.algolia.com/react-instantsearch';
-const license = `/*! ReactInstantSearch ${version} | ${algolia} | ${link} */`;
+const createLicence = () =>
+  `/*! ReactInstantSearch ${version} | ${algolia} | ${link} */`;
```
vvo: `React InstantSearch` should be the actual name (sorry yes it was wrong since the begining)

packages/react-instantsearch/scripts/build-and-publish.sh
```diff
@@ -1,14 +0,0 @@
-#!/usr/bin/env bash
```
vvo: Bye shell scripts, we won't miss you :D

packages/react-instantsearch-native/.babelrc
samouss: I will revert all the `.babelrc` to `ie >= 9` it's the value that we have today in the project. For React Native the impact of changing this to a lower version is very low because we mainly use the `core` package. This dependency is already transpile with a IE9 preset.

PR #1262, opened 2018-05-23
chore(packages): update scripts [PART-5]

samouss:
**Summary**

This is the fifth part for the split of the package. This PR update the scripts to match the new structure, mainly the release script to use Lerna capabilities. We are not using the full power of the tool. We already plan to revamp the scripts part later this year. We will take advantage of the full capabilities of Lerna at that time.

This PR contains:

- update the release scripts
- add the `watch` command to rebuild on packages change
- update the contribution guidelines to match the new structure


CONTRIBUTING.md
```diff
@@ -48,13 +54,13 @@ Files are automatically formatted with prettier.
 ## Release
 
 ```sh
-npm run release
+yarn release
```
Haroenv: does this work now? I remember this throwing annoying errors earlier

scripts/release.sh
```diff
@@ -61,23 +60,47 @@ if $beta; then
...
+# update version in source file
+echo "export default '$newVersion';" > $versionFilePath;
+
+# update version in top level package
 mversion $newVersion
 
+printf "\n"
+
+# update version in packages & dependencies
+lerna publish \
```
Haroenv: yay!

scripts/test.sh
```diff
@@ -4,9 +4,10 @@
 set -e
 
 if [ "$CI" = "true" ]
-  then jest --projects packages/* --runInBand
-  else jest --projects packages/*
+  then jest --runInBand
```
Haroenv: I wonder if it's still useful to have --runInBand (something to check)

CONTRIBUTING.md
```diff
@@ -4,28 +4,34 @@
 
 To run this project, you will need:
 
-- Node.js >= v7.10.0, use nvm - [install instructions](https://github.com/creationix/nvm#install-script)
-- Yarn >= v0.23.4 - [install instructions](https://yarnpkg.com/en/docs/install#alternatives-tab)
+- Node.js >= v8.10.0, use nvm - [install instructions](https://github.com/creationix/nvm#install-script)
```
vvo: Maybe use the latest LTS Node and Yarn now?

CONTRIBUTING.md
vvo: Last time I did it, it worked (latest yarn)

CONTRIBUTING.md
samouss: I synced those values with the `engines` property of the `package.json`. We can upgrade them after the merge of those PRs.

scripts/test.sh
samouss: Indeed it's mostly for debugging (see [docs](https://facebook.github.io/jest/docs/en/cli.html#runinband)). I think we can remove it. Thanks!

Haroenv: sometimes it's also faster on CI; but something to check out

PR #1289, opened 2018-05-31
feat(geo): add docs [PART-10]

samouss:
**Summary**

This is the last part (\o/) of the GeoSearch, the documentation. You can't see the live version yet because the branch is not up to date with `master`. Once it will be the preview will be available. But even without the preview the content remains the same.

- HTML tables rather the markdown because I didn't find a way to have a table without headings. But anyway the style won't be correct if we use the markdown.
- the usage of specific markup for the sub title is intended. It avoid to have a huge space between the sub section of a same component.

[Live preview](https://deploy-preview-1289--react-instantsearch.netlify.com/react-instantsearch/widgets/geosearch)

### DIFF
--- a/docgen/src/stylesheets/components/_documentation.sass
+++ b/docgen/src/stylesheets/components/_documentation.sass
@@ -201,6 +201,10 @@ $offset-height: 60px
         height: $offset-height
         // margin: (-$offset-height) 0 0
 
+    .sub-component-title
+      &:before
+        height: 10px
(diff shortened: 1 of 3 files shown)


Haroenv:
Why not use markdown tables and _do_ add headers (name, type, description)?

samouss:
Because we span all the columns for the description, it might be a bit confusing.

PR #1304, opened 2018-06-04
chore(next): use correct babel configuration

samouss:
**Summary**

With the current configuration the Next application will load the `.babelrc` file that is located at the top level of the project. But Next require a specific configuration to work the extends `next/babel`. You can find more information [in the documentation](https://nextjs.org/docs#customizing-babel-config).

**Before**

localhost_3000_

**After**

localhost_3000_ 1

### DIFF
--- a/examples/next/.babelrc
+++ b/examples/next/.babelrc
@@ -1,0 +1,6 @@
+{
+  "presets": [
+    "next/babel"
+  ],
+  "plugins": []
+}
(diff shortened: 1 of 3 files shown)


samouss:
Yes it’s failing because of the `test` environment in Jest. I will give a try tomorrow.

samouss:
Should be good now, I've updated the dependencies to match the expectation.

PR #1345, opened 2018-06-18
fix(ReactNative): expo demo couldn't be updated

mthuret:
- Icon requires now strictly that images are squared.
- Broken link


samouss:
The snapshots of the examples needs to be updated.

PR #1346, opened 2018-06-18
feat(createInstantSearch): enable _useRequestCache

samouss:
**Summary**

This PR enable the option `_useRequestCache` on the default client. It doesn't impact clients that don't support this option. We only apply the option on the default implementation not on the `searchClient`.


vvo:
Curious: does this solve any issue like duplicate requests right now? 🙏

Haroenv:
probably some other issues too

samouss:
I didn't manage to have a working demo because CodeSandbox doesn't download the correct version of the client. But you can download the sandbox it works locally (see below). The App simulate a code splitting application where the `SearchBox` is mounted after the rest of the page. By default it triggers two requests, but with the option it triggers only one.

> https://codesandbox.io/s/zzqnllyqq3

**Before**

before

**After**

after

PR #1349, opened 2018-06-19
use real location object in example app test

yotammanor:
See example usage in our react-instantsearch based app, 

**Summary**
The existing solution mocks a specific behaviour of the history.location object, but when trying to create tests that require manipulation of the browser location or history, that won't do.

The suggested change shows how to create a real location object, that is usable in tests. 

**Result**
- See our use of this pattern on our react-instantsearch based app: https://github.com/masschallenge/directory/blob/development/src/components/__tests__/AlgoliaSearch.test.js

- Test suite is fully passing locally:
  ```
  Ran all test suites.
  ✨  Done in 475.28s.
  ```

- **Note:** Added dependency `history` already exists as a dependency on react-router ([reference](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/history.md)), and is added for explicitness.


samouss:
Thanks for your contribution!

PR #1358, opened 2018-06-21
fix(maps): avoid cascade update

samouss:
**Summary**

With the current implementation of `Redo` & `Control` components we perform a synchronous call to `setState` on `componentDidMount`. It leads to a wasteful re-render right after the components are mount. All the state is managed by the `Connector` component. When a user uses the `Redo` component it implicitly disable the refine on map move. To do so we call a setState callback from the `Redo.componentDidMount`. It trigger the wasteful re-render for the full tree of the `GeoSearch` component.

This PR solves this issue by removing the implicitness when a user uses the `Redo` component. Now we can disable the refinement on map move with a prop on the `GeoSearch` component. Since the value is directly control by the user we don't need to update the state when the components are rendered. The usage of `Redo` doesn't change. The `Control` component doesn't accept props anymore the value is directly read from the context.

**Usage**

```jsx
// Before with Redo
<GeoSearch google={google}>
  {({ hits }) => <Redo />}
</GeoSearch>

// After with Redo
<GeoSearch google={google} enableRefineOnMapMove={false}>
  {({ hits }) => <Redo />}
</GeoSearch>
```

```jsx
// Before with Control
<GeoSearch google={google}>
  {({ hits }) => <Control enableRefineOnMapMove={false} />}
</GeoSearch>

// After with Control
<GeoSearch google={google} enableRefineOnMapMove={false}>
  {({ hits }) => <Control />}
</GeoSearch>
```

**Metrics**

On the first screenshot you can notice that there is something going wrong during the commit phase, due to the synchronous `setState` (see the ⛔️). Then after the commit phase the all tree is re-render once again. But on the second screenshot we can observe that after the commit phase nothing happens anymore 👌

screen shot 2018-06-21 at 09 57 05 10 31 17

screen shot 2018-06-21 at 09 59 05

### DIFF
--- a/packages/react-instantsearch-dom-geo/src/Connector.js
+++ b/packages/react-instantsearch-dom-geo/src/Connector.js
@@ -9,6 +9,7 @@ export class Connector extends Component {
   static propTypes = {
     hits: PropTypes.arrayOf(PropTypes.object).isRequired,
     isRefinedWithMap: PropTypes.bool.isRequired,
+    enableRefineOnMapMove: PropTypes.bool.isRequired,
     refine: PropTypes.func.isRequired,
     children: PropTypes.func.isRequired,
     position: LatLngPropType,
@@ -16,7 +17,7 @@ export class Connector extends Component {
   };
 
   state = {
-    isRefineOnMapMove: true,
+    isRefineOnMapMove: this.props.enableRefineOnMapMove,
     hasMapMoveSinceLastRefine: false,
   };
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,5 +2,8 @@ module.exports = {
   extends: ['algolia/react', 'algolia/jest'],
   rules: {
     'no-param-reassign': 'off',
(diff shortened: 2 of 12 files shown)


packages/react-instantsearch-dom-geo/src/Connector.js
```diff
@@ -9,14 +9,15 @@ export class Connector extends Component {
   static propTypes = {
```
francoischalifour: Why don't we have the other options [like in InstantSearch.js](https://community.algolia.com/instantsearch.js/v2/widgets/geoSearch.html#usage)?

samouss: We do have the other options when it make sense. The only option that is missing is `paddingBoundingBox` because I didn't manage to implement it properly. The rest of the options is either implemented as component (`Redo`, `Control`)
 or can be achieved with other widgets (`Configure`, `ClearRefinements`).

PR #1367, opened 2018-06-22
test(e2e): remove GeoSearch it is too brittle

samouss:
**Summary**

This PR removes the `GeoSearch` tests in the E2E, because they are too flaky:

- Google Maps is loaded aync
- Tiles of the map are also loaded async

### DIFF
--- a/test/regressions/tests.js
+++ b/test/regressions/tests.js
@@ -14,7 +14,13 @@ function normalize(string) {
   return string.replace(/[ -/]/g, '_');
 }
 
-const stories = require.context('../../stories', true, /\.stories\.js$/);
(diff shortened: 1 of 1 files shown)


test/regressions/tests.js
```diff
@@ -14,7 +14,13 @@ function normalize(string) {
   return string.replace(/[ -/]/g, '_');
 }
 
-const stories = require.context('../../stories', true, /\.stories\.js$/);
+// Remove the GeoSearch stories from the tests
+// beacuse they are too much brittle
```
Haroenv: Nit: bec**au**se they are **too** brittle

PR #1371, opened 2018-06-22
docs(guide): Clarify parent requirement

dardub:
Clarify required parent component for connected components. This is important information that shouldn't be hidden in comments.

**Summary**

**Result**


Haroenv:
Thanks!

Issue #1042, opened 2018-03-05
More aggressive shallow equal in willReceiveProps


samouss:
## Problem

In the current implementation on every render we check if the given props are shallow equal or not in order to trigger a new request. We check **all** the props and not only the ones used by the connector. It can lead to useless requests when we pass an props that will change on every render (ex: Anonymous function, value from a Redux store, plain object, ...).

See https://github.com/algolia/react-instantsearch/issues/1002 https://github.com/algolia/react-instantsearch/issues/1037.

https://github.com/algolia/react-instantsearch/blob/5ecec858f41ff3964bce0d26431804f8fe67308f/packages/react-instantsearch/src/core/createConnector.js#L166-L175

## Solution

A possible solution will be to let the inner connector do the shallow equal. The connector has more knowledge about his props than `createConnector`. By doing that we can specify which props needs to be check in order to trigger a new request (ex: `attribute`). We should also split the props in two categories UI / search. An update on the UI props must not trigger a new request (ex: `translations`).

rtman:
Just wondering if this has progressed at all? I still see occasions when searching is triggered and it shouldn't be.

samouss:
Nope not yet! We don't have any ETA on it neither.

sarahdayan:
As we're deprecating React InstantSearch v6 to the benefit of React InstantSearch Hooks, we'll most likely not work on this optimization in v6.

PR #1389, opened 2018-06-26
poc: Will receive update

samouss:
**Problem**

This is a first attempt to solve this issue #1042. With the current implementation we trigger a change on the widget manager each time a props change on a connector. We detect those changes with the [`isEqual`](https://lodash.com/docs/4.17.10#isEqual) function of Lodash. The widget manager update schedule new queries and a re-render. The queries are (most of the time) already cached so we don't see any network request. But in case we have a connector inside another one it can leads to an infinite loop:

```jsx
const Connected = connectStateResults(() => (
  <RefinementList
    attribute="colors"
    translations={{
      showMore: extended => (extended ? "Less" : "More")
    }}
  />
));

const App = () => (
  <InstantSearch>
    <Connected />
  </InstantSearch>
);
``` 

In that case we have an infinite loop because the translations prop contain a function. On each render a new function is created. It means that on each render React InstantSearch will trigger a new change on the widget manager that will also trigger a new render (∞). We have our infinite loop.

**Solution**

The responsibility to trigger a new change is hold by the `createConnector` function. But this abstraction don't have enough information to know wether or not a new request is required. The component that have the more knowledge about that is the connector. We can define a new hook (called `shouldWidgetUpdate`) on the connector description to determine when the widget needs to trigger a new request.

```jsx
const connectRefinementList = createConnector({
  shouldWidgetUpdate(props, nextProps) {
    return (
      props.attribute !== nextProps.attribute ||
      props.operator !== nextProps.operator ||
      props.limit !== nextProps.limit ||
      props.showMoreLimit !== nextProps.showMoreLimit ||
      props.showMore !== nextProps.showMore ||
      !isEqual(props.defaultRefinement, nextProps.defaultRefinement)
    );
  }
});
```

Now we trigger only a new change on the widget manager when on of those props change. I took `RefinementList` because it's one of the widget that has the most props implied in the request. For most of the widget the test will be less "complicated" than that.

**Pros**:

- more control on the lifecycle
- avoid wasteful loop to the widget manager
- we can now provide anonymous function to props
- nested connectors is a bit more safe (mount / unmount issue is still there)

**Cons**:

- one more function added to the connector API
- it feels (is) an escape hatch

-----

Let me know what do you think or if you have a better ideas!

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectRefinementList.js
+++ b/packages/react-instantsearch-core/src/connectors/connectRefinementList.js
@@ -1,3 +1,4 @@
+import { isEqual } from 'lodash';
 import PropTypes from 'prop-types';
 import createConnector from '../core/createConnector';
 import {
@@ -260,5 +261,17 @@ export default createConnector({
             ]
           : [],
     };
+  },
+
+  shouldWidgetUpdate(props, nextProps) {
+    const shouldUpdate =
+      props.attribute !== nextProps.attribute ||
+      props.operator !== nextProps.operator ||
+      props.limit !== nextProps.limit ||
+      props.showMoreLimit !== nextProps.showMoreLimit ||
+      props.showMore !== nextProps.showMore ||
+      !isEqual(props.defaultRefinement, nextProps.defaultRefinement);
+
+    return shouldUpdate;
   },
 });
--- a/packages/react-instantsearch-core/src/core/createConnector.js
+++ b/packages/react-instantsearch-core/src/core/createConnector.js
@@ -42,6 +42,7 @@ export default function createConnector(connectorDesc) {
   const hasTransitionState = has(connectorDesc, 'transitionState');
   const hasCleanUp = has(connectorDesc, 'cleanUp');
   const hasShouldComponentUpdate = has(connectorDesc, 'shouldComponentUpdate');
+  const hasShoulWidgetdUpdate = has(connectorDesc, 'shouldWidgetUpdate');
   const isWidget = hasSearchParameters || hasMetadata || hasTransitionState;
 
   return Composed =>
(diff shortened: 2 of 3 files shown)


packages/react-instantsearch-core/src/connectors/connectRefinementList.js
```diff
@@ -1,3 +1,4 @@
+import { isEqual } from 'lodash';
```
Haroenv: We import like this normally to avoid having the whole bundle: `import isEqual from 'lodash/isEqual'` right?

stories/RefinementList.stories.js
```diff
@@ -1,9 +1,14 @@
-import React from 'react';
+import React, { Component } from 'react';
 import orderBy from 'lodash.orderby';
```
Haroenv: since it's a story and we have a dependency on lodash already, you can change this to `import orderBy from 'lodash/orderby';`. That way this devdep can go :)

samouss: Actually no I explicitly use the `lodash.orderby` because Rollup messed up when `lodash` is present as a top level dependency on the Workspaces.

bobylito:
Do you have another solutions that you want to explore @samouss?

samouss:
> Do you have another solutions that you want to explore @samouss?

Nope I didn't found one that doesn't require a major change.

packages/react-instantsearch-core/src/core/createConnector.js
```diff
@@ -42,6 +42,7 @@ export default function createConnector(connectorDesc) {
   const hasTransitionState = has(connectorDesc, 'transitionState');
   const hasCleanUp = has(connectorDesc, 'cleanUp');
   const hasShouldComponentUpdate = has(connectorDesc, 'shouldComponentUpdate');
+  const hasShoulWidgetdUpdate = has(connectorDesc, 'shouldWidgetUpdate');
```
bobylito: There is a typo here, in the name of the variable, should not shoul

samouss:
Close in favor of #1394

PR #1390, opened 2018-06-26
docs(GeoSearch): add missing prop enableRefine

samouss:


docgen/src/widgets/GeoSearch.md
```diff
@@ -143,6 +143,16 @@ const App = () => (
...
       </td>
     </tr>
+    <tr>
+      <td>enableRefine</td>
+      <td>Type: <code>boolean</code></td>
+      <td>Default: <code>true</code></td>
+    </tr>
+    <tr>
+      <td colspan="3">
+        <p>If false, you can't refine with the map.</p>
```
Haroenv: If this is false, this map is for display purposes only, not for refining the search

PR #1394, opened 2018-06-27
feat(core): implement shouldWidgetUpdate

### LINKED ISSUE #1042: More aggressive shallow equal in willReceiveProps

samouss:
**Summary**

Fix #1042 

This PR implements the hook `shouldWidgetUpdate` in `createConnector` following what we discuss in #1389. I will create one PR for each connector it will be easier to review with this branch as a base.

The diff in the test file is a bit long because I moved another test on the top level. What really matters for the PR is in this commit https://github.com/algolia/react-instantsearch/pull/1394/commits/e5cfa22396be74bfa92fb3b1d2addea44d292bdd.

**Usage**

```jsx
const customConnector = createConnector({
  shouldWidgetUpdate(props, nextProps) {
    return props.attribute !== nextProps.attribute;
  }
});
```

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createConnector.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createConnector.js
@@ -585,72 +585,6 @@ describe('createConnector', () => {
         { canRender: true, props: null }
       );
     });
-
-    describe('unmounting', () => {
-      const Connected = createConnector({
-        displayName: 'CoolConnector',
-        getProvidedProps: () => null,
-        getMetadata: () => null,
-        cleanUp: () => ({ another: { state: 'state', key: {} }, key: {} }),
-      })(() => null);
-      const unregister = jest.fn();
-      const setState = jest.fn();
-      const onSearchStateChange = jest.fn();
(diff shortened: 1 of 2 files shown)


samouss:
Close for the moment, will update the branch later.

PR #1404, opened 2018-07-02
chore: rename maps folder to maps

samouss:
**Summary**

This PR rename the folder for the `maps` package and make it public for the release.


samouss:
> ## Requirements
>
> The API of this widget is a bit different than the others that you can find in React InstantSearch. The API is component driven rather than options driven. We chose the former because it brings more flexibility to the widget. Since the geo search pattern is not a use case for every applications we decided to ship the widget in a separate package. Be sure to install it before using it:
>
> ```shell
> yarn add react-instantsearch-dom-maps
> ```

You are talking about that?

> https://github.com/algolia/react-instantsearch/blob/master/docgen/src/widgets/GeoSearch.md#requirements

Haroenv:
ah perfect

PR #1405, opened 2018-07-02
chore: drop story for multi highlight for now

samouss:
**Summary**

The story is failing because we don't have access to the index anymore. I dropped it for now to avoid issue with E2E. We should find a proper solution to datasets with missing features. In that case highlighted an array attribute.


samouss:
The best would be to have a dataset that handle most of the use cases (that we support).

PR #1422, opened 2018-07-10
docs: Revamp README and CONTRIBUTING

francoischalifour:
This is a revamp of the README and the CONTRIBUTING guides.

## Changes

- Update logo
- Add "Why" section
- Add `npm` command in "Installation" section
- Add "Getting started" section
- Add "Demos" section
- Add "Playground" section
- Add "License" section

## Questions

The recommended way of using React InstantSearch is now with the `searchClient` option. Should we also update the [installation guide](https://community.algolia.com/react-instantsearch/Getting_started.html#install-react-instantsearch) of the website?

## Related

* algolia/instantsearch.js#3007


samouss:
Why don't we use the same structure than the one on InstantSearch?

francoischalifour:
I was wondering if we wanted to go the same way with React InstantSearch. I'm not sure we are addressing to the same kind of developers.

I'm OK to go this way though, let me know!

samouss:
We have an introduction, (quick) getting started to see how it’s simple, demo, etc... Whatever the target of developers IMO it still valuable to have those kind of information on the README.

francoischalifour:
I updated the content to be more aligned with [InstantSearch.js](https://github.com/algolia/instantsearch.js/blob/0d90276af4ba40bc32a5da5ba9c6d5dca9eab9f0/README.md).

Do we also want to add the "We're hiring" section?

PR #1445, opened 2018-07-18
Fix documentation for VirtualMenu

kwent:
Hi,

`connectMenu` should be imported from `react-instantsearch/connectors`

Regards


francoischalifour:
Hi @kwent, thank you for getting involved.

Since the 5.2.0 release, `react-instantsearch-dom` is the recommended way of importing the package. It contains both widgets and connectors.

You can read more on the [migration guide on package import](https://community.algolia.com/react-instantsearch/guide/Migration_guide_package_import.html#migration-for-react-dom).

Let us know if there's anything we haven't communicated clearly!

PR #1459, opened 2018-07-26
HeirachialMenu empty array parent--item

digitalcraft:
There is more context in AlgoliaPartners#ps-digital8 channel.
Image of Bug

The bug happens when a product is in a level 1 category and a level 2 category.

### DIFF
--- a/packages/react-instantsearch-dom/src/components/List.js
+++ b/packages/react-instantsearch-dom/src/components/List.js
@@ -61,7 +61,7 @@ class List extends Component {
   };
 
   renderItem = (item, resetQuery) => {
-    const items = item.items && (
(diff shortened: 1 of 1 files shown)


Haroenv:
This probably needs a new test to show the behaviour this fixes, WDYT? 

Thanks for your contribution though! ❤️ 

Also listing is crying because I think you need to write it on more lines or something

samouss:
I wasn't able to update the PR directly so I merged the branch in a new branch that I can edit (thanks @Haroenv for the idea). Here is the new PR https://github.com/algolia/react-instantsearch/pull/1472.

cc @digitalcraft

PR #1461, opened 2018-07-27
[prop-types] Allow `object` as type for `root.Root` (Fixes #1446)

### LINKED ISSUE #1446: Failed prop type: Invalid prop `root.Root` warning
**example code causing the issue:**

(taken from https://community.algolia.com/react-instantsearch/Getting_started_React_native.html#add-the-%3Cinstantsearch%3E-component)
``` javascript
<InstantSearch appId="latency" apiKey="6be0576ff61c053d5f9a3225e2a90f76" indexName="ikea">
    <Text>Congrats 🎉! Your application is now connected to Algolia!</Text>
</InstantSearch>
```

Warning is being thrown: 
```
Warning: Failed prop type: Invalid prop `root.Root` supplied to `CreateInstantSearch`.
    in CreateInstantSearch (at Search.js:21)
    ... further stack incoming
```

No warning I guess? However, the search still works, it's just a warning, not an error.

**Bug: Which versions?**

RN@0.56.0, babel@7.0.0-beta.52, react-instantsearch@5.2.2

JonnyBurger:
**Summary**

Removes two warnings you get in React Native 0.56.
<View/> now uses React.forwardRef() (exact commit in react-native 0.56: https://github.com/facebook/react-native/commit/3e534b9aab5156adac67762877b2457408fe8934)

Therefore:

```js
import {View} from 'react-native';
console.log(typeof View) // 'object'
```

This PR adds support for that.

**Result**

Two warnings disappear.
Fixes #1446


Haroenv:
Thanks, seems to be a long list now, I wonder if what we're using this for satisfies `PropTypes.element`; Not sure if that includes the string

JonnyBurger:
Tried it out quickly and it seems like `PropTypes.element` is *not* satisfied with `View`.

samouss:
Thanks for the fix, I will merge it as it is. There is an ongoing discussion to support `forwardRef` directly in [the PropTypes package](https://github.com/facebook/prop-types/issues/200). Will follow how the discussion evolve and we might revert the changes depending on the outcome. Thanks again!

JonnyBurger:
Thanks a lot!
When can I expect this to be released?

samouss:
Probably beginning of next week.

PR #1472, opened 2018-08-06
fix(List): render children list only when required

samouss:
**Summary**

This PR is a followup of https://github.com/algolia/react-instantsearch/pull/1459. It adds tests to the `List` component (only for the changes not the all component). Now we render the children list only when the `item` contains an non empty array of `items`. The same condition is applied on the class `item--parent`.

**Before**

screen shot 2018-08-06 at 14 25 23

**After**

screen shot 2018-08-06 at 14 25 45

Note: The arrow is not present on the latter since we removed the class.

### DIFF
--- a/packages/react-instantsearch-dom/src/components/List.js
+++ b/packages/react-instantsearch-dom/src/components/List.js
@@ -61,7 +61,7 @@ class List extends Component {
   };
 
   renderItem = (item, resetQuery) => {
-    const isItemHasChildren = item.items && Boolean(item.items.length);
+    const itemHasChildren = item.items && Boolean(item.items.length);
 
     return (
       <li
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-dom/src/components/List.js
```diff
@@ -61,13 +61,7 @@ class List extends Component {
...
 
   renderItem = (item, resetQuery) => {
-    const items = item.items && (
-      <ul className={this.props.cx('list', 'list--child')}>
-        {item.items
-          .slice(0, this.getLimit())
-          .map(child => this.renderItem(child, item))}
-      </ul>
-    );
+    const isItemHasChildren = item.items && Boolean(item.items.length);
```
Haroenv: `hasChildren` should be clear enough too right? This name now is pretty unclear :)

samouss: Indeed. But I prefer to be explicit with something like `itemHasChildren`. WDYT?

Haroenv: itemHasChildren also would be fine :)

PR #1484, opened 2018-08-09
docs: add Rheostat warning about version

samouss:
**Summary**

This PR adds a warning about the Rheostat usage inside the docs & Storybook.

screen shot 2018-08-09 at 16 27 15

screen shot 2018-08-09 at 16 27 32

### DIFF
--- a/packages/react-instantsearch-dom/src/widgets/RangeSlider.js
+++ b/packages/react-instantsearch-dom/src/widgets/RangeSlider.js
@@ -10,12 +10,13 @@ import React from 'react';
  * @kind widget
  * @example
  *
- * // Here's an example showing how to connect the airbnb rheostat slider to React InstantSearch using the
- * // range connector
- import React, {Component} from 'react';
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-dom/src/widgets/RangeSlider.js
```diff
@@ -10,12 +10,13 @@ import React from 'react';
  * @kind widget
  * @example
  *
- * // Here's an example showing how to connect the airbnb rheostat slider to React InstantSearch using the
- * // range connector
- import React, {Component} from 'react';
+ * // Here's an example showing how to connect the AirBnb Rheostat Slider to React InstantSearch
+ * // using the range connector. ⚠️ This example only works with the version 2.x of Rheostat.
+
```
Haroenv: would add in the docs maybe another line saying that it's because the classes in Rheostat 3.x are not deterministic?

samouss: It's not true, they are deterministic but the setup is much more harder for the users. The main reason is that our theme only works with the v2.

PR #1485, opened 2018-08-09
chore(release): avoid interactive release

samouss:
**Summary**

Yarn can now publish without the interactive prompt even without providing the version.


samouss:
Yep it's the latest one actually.

PR #1515, opened 2018-08-20
chore: switch from Travis to CircleCI

samouss:
**Summary**

This PR move the build task from [Travis](https://travis-ci.org/) to [CircleCI](https://circleci.com/). The main argument for this move is **speed**.  On Travis we have a huge number of projects so our jobs are most of the time queued for a moment (up to 1 ~ 2 hrs). For the moment we don't have this problem on CircleCI. But even without talking about the queuing time CircleCI runs the jobs faster than Travis.

- **Travis**: **13 minutes** build time / **33 minutes** total time (average on the last 5 builds)
- **CircleCI**: - / **8 minutes** (average on the last 5 builds)

The workflow is divided into 4 jobs - 3 of them run under 4 minutes most of the time. The last one install & build and test all the examples, it takes ~8 minutes. The next step is to move those examples on the workspaces to avoid the install step for some of them.


samouss:
Yes but at the end the configuration is not that complicated (once you read the docs).

Issue #1521, opened 2018-08-23
Bug in createInstantSearchServer/connectRange


BrentLayne:
**Bug: What is the current behavior?**

When providing a `min` and/or `max` prop to `connectRange` (https://community.algolia.com/react-instantsearch/connectors/connectRange.html), server-side rendering throws an error.

```
TypeError: Cannot read property 'min' of undefined
 at getCurrentRefinement (......node_modules/react-instantsearch-core/dist/cjs/connectors/connectRange.js:118:35)
 at Object.getSearchParameters (......node_modules/react-instantsearch-core/dist/cjs/connectors/connectRange.js:264:33)
 at searchParameters.filter.reduce._algoliasearchHelper.SearchParameters._extends.index (......node_modules/react-instantsearch-dom/dist/cjs/core/createInstantSearchServer.js:78:50)
 at Array.reduce (<anonymous>)
 at findResultsState (......node_modules/react-instantsearch-dom/dist/cjs/core/createInstantSearchServer.js:77:8)
```

This error is caused by a bug in `createInstantSearchServer`.  When reducing over the `searchParameters` in `findResultsState`, `searchParameters.getSearchParameters` is being `call`ed with the wrong `this` context.

This incorrect `this` context causes the error in connectRange.  connectRange's method `getSearchParameters` calls `getCurrentRefinement`, which tries to access the `min` and `max` properties of `this._currentRange` if `min` and/or `max` props are provided.  Because `getSearchParameters` has been bound to the wrong `this` context, `this._currentRange` is undefined instead of an object.

**Bug: What is the expected behavior?**
It is expected that you can pass a `min` and/or `max` prop to connectRange and that it will not throw an error during server-side rendering.  

**Bug: What browsers are impacted? Which versions?**
All environments

PR #1522, opened 2018-08-23
Fix bug with createInstantSearchServer/connectRange

BrentLayne:
**Summary**
In `createInstantSearchServer`, when reducing over the searchParameters in `findResultsState`, `searchParameters.getSearchParameters` was being `call`ed with the wrong `this` context.

This led to a bug in connectRange during server-side rendering. `this._currentRange` was `undefined` when `getSearchParameters` was being called, due to `this` being incorrect.

This bug is hit when a `min` and/or `max` prop is provided to connectRange. `createInstantSearchServer` calls connectRange's `getSearchParameters`  and an if block is executed inside `getCurrentRefinement`  that tries to access `min` and `max` properties on the undefined `this._currentRange`, causing an error

See the issue I created for more details: #1521 

**Result**
This PR resolves the problem of being unable to provide a `min` and/or `max` prop to connectRange by ensuring that each connector is spreading it's own `this` context into the `call` of `getSearchParameters` by `createInstantSearchServer`


packages/react-instantsearch-core/src/core/createConnector.js
```diff
@@ -162,7 +162,8 @@ export default function createConnector(connectorDesc) {
           this.context.ais.onSearchParameters(
             connectorDesc.getSearchParameters,
```
samouss: Rather than providing the `this` as argument to the function we can bound the `this` to the function `getSearchParamters`. It avoids to pass the self on the parent function.

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -60,7 +62,10 @@ const createInstantSearchServer = algoliasearch => {
       .reduce(
         (acc, searchParameter) =>
           searchParameter.getSearchParameters.call(
-            { context: searchParameter.context },
+            {
```
samouss: Since the function is now bound to his instance we can call it directly without the `call`.

BrentLayne:
@samouss Thanks for the feedback! I have addressed your changes if you can please take another look.  I tried to re-request your review but I don't seem to have permissions to do that.  Cheers

packages/react-instantsearch-core/src/core/__tests__/createConnector.js
```diff
@@ -417,7 +417,7 @@ describe('createConnector', () => {
       });
 
       expect(onSearchParameters.mock.calls).toHaveLength(1);
-      expect(onSearchParameters.mock.calls[0][0]).toBe(getSearchParameters);
+      expect(onSearchParameters.mock.calls[0][0]).toMatchSnapshot();
```
samouss: We can remove the snapshot and use a matcher to expect the argument to be a function.

```js
expect(onSearchParameters.mock.calls[0][0]).toEqual(expect.any(Function));
```

BrentLayne:
@samouss No problem - thanks for your feedback.  I've used a matcher to ensure onSearchParameters is being passed a `Function` and I wrote an additional test to ensure that `getSearchParameters` is being bound to the correct `this` context, if you can please review one more time.  Thanks!

BrentLayne:
@Haroenv if you can please re-request @samouss for review, I would greatly appreciate it.

Issue #1419, opened 2018-07-09
"Edit this page" button URL is broken


samouss:

JureSotosek:
I see this still hasn't been resolved. Is anyone working on it? Can i give it a try?

Haroenv:
Yes, if you feel like contributing, this definitely would be welcome. It's likely something that changed when we changed the folder structure. Source for the documentation is https://github.com/algolia/react-instantsearch/tree/master/docgen

JureSotosek:
@Haroenv One quick question. If I'm not mistaken, docs are partly generated from comments from inside some of the files right?

So for example [connectHits page](https://community.algolia.com/react-instantsearch/connectors/connectHits.html) is generated from the comments from [this](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/connectors/connectHits.js) file and if somebody clicked "Edit this page" on the connectHits page they should be redirected to [this](https://github.com/algolia/react-instantsearch/edit/master/packages/react-instantsearch-core/src/connectors/connectHits.js), an edit page for the file I linked.

Or am i missing something?

Haroenv:
yes, the link should be towards the source code, since that's what is needed to be updated :)

PR #1562, opened 2018-09-09
fix: Edit this page button

JureSotosek:
This is a solution to #1419.

The button stoped working because the folder structure was changed. I made changes in the editThisPage.js file where the path for the source code was generated to take in to account the new file structure.

Note: This fix is very barebones, for a more robust solution the way how the pathname is generated would have to be "optimised" or at least minor changes in the file structure should be done, more specifically where the source of the "server-side-rendering" docs comes from. I can make a simple PR if anyone is interested.


JureSotosek:
The deployment preview does not work because it, for some reason, lowercases the path url when navigating trough the documentation. That does not happen on my local machine or the actual website. It is again, a problem that could be avoided if the pathname generation process was different.

JureSotosek:
And, I guess, the deployment of the preview should be changed since it acts differently than the production deployment.

francoischalifour:
I wonder if we should skip the magical URL mapping and have the GitHub source hardcoded in the frontmatter [as we do for InstantSearch.js](https://github.com/algolia/instantsearch.js/blob/develop/docgen/src/guides/calendar-widget.md).

JureSotosek:
@francoischalifour That's also what i was thinking, or maybe in the jsdoc-parsing process store the path of each doc entry, jsdoc-parse actually supplies a `data.meta.path`, in a variable and then supply it to the editthispage button.

But all of that would require a way bigger and extensive change to the code base.

JureSotosek:
So is there or isn't there a problem with the InstantSearch and Index widgets?

francoischalifour:
@JureSotosek They are correct! My bad for the misunderstandings.

PR #1568, opened 2018-09-10
Fix: Compute canRefine on the transformed items

### LINKED ISSUE #1524: Compute canRefine on the transformed items

BUG.

With the current implementation of `connectCurrentRefinements` the `canRefine` props is computed on items rather than the transformed items. It can leads to issue when we filter some values from items.

Compute `canRefine` on the transformed items.

**Connectors impacted**:

- `connectCurrentRefinements`
- `connectHierarchicalMenu`
- `connectMenu`
- `connectNumericMenu`
- `connectRefinementList`
- `connectRefinementList`

ltk:
Resolves #1524

**Summary**
There was previously a bug where `canRefine` props were calculated based on untransformed item lists. This caused unintended behaviors such as a ClearRefinements button remaining active when a refinement being ignored via a filter in `transformItems` was present.

**Result**
This PR calculates `canRefine` on the transformed list of items, correcting this behavior.

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectCurrentRefinements.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectCurrentRefinements.js
@@ -112,10 +112,11 @@ describe('connectCurrentRefinements', () => {
   });
 
   it('computes canRefine based on the length of the transformed items list', () => {
-    const transformItems = items => [];
+    const transformItems = () => [];
     const props = getProvidedProps({ transformItems }, null, null, [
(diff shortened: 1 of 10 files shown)


ltk:
@samouss yep, I can add those tests in the next couple of days.

ltk:
@samouss, better late than never :-). Just pushed up those new tests.

samouss:
One last change and we are ready to merge! You have to run Prettier to fix the style issue. Thanks a lot for your contributions!

PR #1588, opened 2018-09-18
feat(connectToggleRefinement): implement canRefine & count

samouss:
**Summary**

This PR adds `canRefine` & `count` on the provided props of the `connectToggleRefinement` connector. I didn't implement the count on the widget because it will be a breaking change only in the connector. 

The count object has the following shape (see below). When the `currentRefinement` is `false` (unchecked) the user can pick the `unchecked` attribute. It displays the number of results that match the value provided to the Toggle widget. Once it's checked the count displays the total number of results (compute from the facet values).

```ts
type Count = {
  checked: number | null;
  unchecked: number | null;
};
```

**Changes**:

- the computation of the `SearchParameters` (conjunctiveFacet -> disjunctiveFacet)
- the computation of the `providedProps` (adds `canRefine` & `count`)
- adds the `noRefinement` class on the Toggle root element

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectToggleRefinement.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectToggleRefinement.js
@@ -3,7 +3,6 @@ import connect from '../connectToggleRefinement';
 
 jest.mock('../../core/createConnector', () => x => x);
 
-let props;
 let params;
 
 describe('connectToggleRefinement', () => {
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch-core/src/connectors/__tests__/connectToggleRefinement.js
```diff
@@ -192,30 +537,90 @@ describe('connectToggleRefinement', () => {
...
+              },
+            },
+          },
+        }
+      );
+
+      expect(params.getDisjunctiveRefinements('facet')).toEqual(['val']);
+    });
+
+    it('does not refines the corresponding facet with `false`', () => {
```
francoischalifour: `refines` → `refine`

packages/react-instantsearch-core/src/connectors/connectToggleRefinement.js
```diff
@@ -62,20 +64,48 @@ export default createConnector({
...
+    const results = getResults(searchResults, this.context);
     const currentRefinement = getCurrentRefinement(
       props,
       searchState,
       this.context
     );
-    return { currentRefinement };
+
+    const facetValues = results && results.getFacetValues(attribute);
+    const facetValue =
```
francoischalifour: These two variable names are confusing.

packages/react-instantsearch-core/src/connectors/connectToggleRefinement.js
```diff
@@ -62,20 +64,48 @@ export default createConnector({
...
+      // Use null to always be consistent with type of the value
+      // count: number | null
+      facetValues && facetValues.length
+        ? find(facetValues, item => item.name === value.toString())
+        : null;
```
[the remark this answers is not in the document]
samouss: You get `null` on the first render because `searchResults` is an empty object.

packages/react-instantsearch-core/src/connectors/connectToggleRefinement.js
samouss: Any suggestions? I'm didn't find something else.

francoischalifour: When I deal with both singular and plural variable names, I tend to go for `allFacetValues` and `facetValue`. WDYT?

samouss: To be honest I don't see a huge benefit, I'm used to plural singular like the original one. But I get your point.

francoischalifour: When I came across these two lines I thought you were referring to the same variable. The benefit is just to avoid this confusion. This is subjective though.

samouss: Yep I've renamed it.

VincentCtr:
@samouss it seems that this PR is causing a crash when we use ToggleRefinement and change the index used by <InstantSearch />

```
app.js:sourcemap:406316 Error: **FaceName** is not a retrieved facet.
    at SearchResults../node_modules/algoliasearch-helper/src/SearchResults/index.js.SearchResults.getFacetValues (app.js:sourcemap:55188)
    at Connector.getProvidedProps (app.js:sourcemap:281799)
    at Connector.getProvidedProps (app.js:sourcemap:282134)
    at new Connector (app.js:sourcemap:281975)
    at constructClassInstance (app.js:sourcemap:250944)
    at updateClassComponent (app.js:sourcemap:252666)
    at beginWork (app.js:sourcemap:253265)
    at performUnitOfWork (app.js:sourcemap:255591)
    at workLoop (app.js:sourcemap:255629)
    at renderRoot (app.js:sourcemap:255708)

```

I spent the whole day trying to figure out what happen and finally came back to 5.2.3

VincentCtr:
Hi @samouss, thank you for this very quick fix !

samouss:
The PR has been merged and released ([v5.3.1](https://github.com/algolia/react-instantsearch/blob/master/CHANGELOG.md#531-2018-09-26)), let me know if you still have any issues.

PR #1609, opened 2018-09-21
Hit Component: include Index Position as a data attribute

stomachfat:
**Summary**

Summary

Use Case: As a consumer of Algolia/react-instancesearch, I want to be able to track the position of the hit my user has click.

Solution: pass `indexPosition` to `HitComponent`

Friendly Note: If there's already a way to do this that isn't hacky please let me know. As of currently I'm grabbing the hitID and reaching into the dom and determining the position of the hit manually.

Alternative Implementation: We could also add a data-attribute on the `Hit` component. Please see https://github.com/algolia/react-instantsearch/pull/1608

**Result**

### DIFF
--- a/packages/react-instantsearch-dom/src/components/Hits.js
+++ b/packages/react-instantsearch-dom/src/components/Hits.js
@@ -10,9 +10,12 @@ const Hits = ({ hits, className, hitComponent: HitComponent }) => (
   // ex: <HitComponent {...hit} key={hit.objectID} />
   <div className={classNames(cx(''), className)}>
     <ul className={cx('list')}>
-      {hits.map(hit => (
+      {hits.map((hit, indexPosition) => (
         <li className={cx('item')} key={hit.objectID}>
-          <HitComponent hit={hit} />
+          <HitComponent
+            hit={hit}
+            indexPosition={indexPosition}
+          />
         </li>
       ))}
     </ul>


Haroenv:
We actually had this in the backlog, but instead with absolute position (with page and hitsPerPage taken in account). What exactly is the usecase you had for getting the index on the page? 

Thanks for your contribution!

stomachfat:
@Haroenv  heyyy!! Thanks for the reply : D It's cool to talk to someone on the other side!

So the use case is: 

We've implemented Algolia Search in our tool that our users like to use. And we need to track how good the search matches to the query. We can do that by tracking absolute position. (For example, perfect match is first position, good match in the first few, weaker matches). 

We want to use that information to adjust what we index in Algolia and what we make searchable, in order to make our search experience optimized for users.

Haroenv:
Makes sense, however this means that we should in fact give the absolute position, which is index + page * hitsPerPage.

Haroenv:
It’ll probably be best (at least that was my plan) to add the position in the connector, and then expose it to the widget too. A data attribute is also possible, but not required for now. 

By the way, see also: https://github.com/algolia/search-insights.js/tree/master/examples/react-instantsearch for the current workaround to do it in user land :)

samouss:
Hi @stomachfat, like @Haroenv said, we planned to expose the **absolute** position to the component not the **relative** one. Would you interested to update your PR? If not I'm gonna close it, we'll implement the proposed solution in the coming months.

PR #1610, opened 2018-09-21
Hit Component: include Index Position as a data attribute

stomachfat:
**Summary**

Use Case: As a consumer of Algolia/react-instancesearch, I want to be able to track the position of the hit my user has click. 

Solution: put the index position as a data-attribute on the Hits

Friendly Note: If there's already a way to do this that isn't hacky please let me know. As of currently I'm grabbing the hitID and reaching into the dom and determining the position of the hit manually.

Alternative Implementation: We could also make `index-position` an argument that we can pass into the `hitComponent` as a prop. Please see https://github.com/algolia/react-instantsearch/pull/1609

**Result**


samouss:
Close in favour of #1609.

PR #1613, opened 2018-09-25
fix(toggle): ensure facet is present

samouss:
**Summary**

This PR fixes an issue when we tried to access the facet values on a `SearchResults` that don't have the expected facet. We can ensure this with `getFacetByName` - this method is deprecated on the helper but we don't have a proper alternative yet.

**Before**

before

**After**

after


samouss:
Yep indeed, or throw only in development. It can be handy to have this information when you develop.

samouss:
Yes for sure the error about the facet is thrown inside the helper actually.

PR #1623, opened 2018-09-27
docs: update tourism example with Google Maps widget

samouss:
**Summary**

Closes https://github.com/algolia/react-instantsearch/issues/308

This PR updates the Tourism example to use the Google Maps widget of React InstantSearch.

**Result**

untitled

### DIFF
--- a/docgen/src/examples/tourism/App.js
+++ b/docgen/src/examples/tourism/App.js
@@ -61,8 +61,8 @@ function Header() {
 function Filters() {
   return (
     <div className="container-fluid">
-      <div className="row">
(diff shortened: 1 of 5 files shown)


docgen/src/examples/tourism/App.js
```diff
@@ -1,7 +1,8 @@
-/* eslint react/prop-types: 0 */
+/* eslint react/prop-types: 0, import/no-extraneous-dependencies: 0 */
```
Haroenv: why is this added? :)

samouss: Otherwise ESLint complaint about the maps package but not the normal one.

PR #1647, opened 2018-10-08
docs(refresh): use setState callback

samouss:
**Summary**

The previous implementation of the examples was relying on the `onSearchStateChange` callback to reset the `refresh` value. But the callback is not triggered when the cache is refreshed. It means that the value remains `true` until the search **actually** change. 

This PR reset the value immediately with the help of the `setState` callback.

### DIFF
--- a/docgen/src/guide/Refreshing_cache.md
+++ b/docgen/src/guide/Refreshing_cache.md
@@ -35,21 +35,18 @@ class App extends Component {
   }
 
   refresh = () => {
-    this.setState({ refresh: true });
-  };
-
-  onSearchStateChange = () => {
-    this.setState({ refresh: false });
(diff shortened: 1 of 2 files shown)


docgen/src/guide/Refreshing_cache.md
```diff
@@ -35,21 +35,18 @@ class App extends Component {
...
   };
 
   render() {
     return (
       <InstantSearch
-        appId="yourAppId"
-        apiKey="yourApiKey"
-        indexName="yourIndexName"
+        appId="latency"
+        apiKey="6be0576ff61c053d5f9a3225e2a90f76"
```
Haroenv: searchClient?

PR #1675, opened 2018-10-17
Add option to disable empty search

### LINKED ISSUE #1111: Feature Request - Ability to prevent search on init
By default `<InstantSearch>` performs search with an empty query on init. 

**Feature: What is your use case for such a feature?**
Performance and network optimization. Currently page doesn't finish loading until this initial search is finished. Also, I want to prevent searches with less than a few characters.

This is available in `instantsearch.js` with the `searchFunction` hook. Docs: https://community.algolia.com/instantsearch.js/v1/documentation/#hide-results-on-init

I'd love if that hook can be used via a prop on `<InstantSearch>`. Ideally, every possible option for `instantsearch.js` would automatically be usable on `<InstantSearch>`.

Thank you!

janza:
Closes #1111 

Added an extra prop to `InstantSearch` which disables searching for empty queries.

As suggested in the issue it might also be an option to add a `searchFunction` hook which can be used to disable searching depending on the query. I'm open to implementing it if that would be a prefered solution.

### DIFF
--- a/packages/react-instantsearch-core/src/components/InstantSearch.js
+++ b/packages/react-instantsearch-core/src/components/InstantSearch.js
@@ -35,6 +35,7 @@ function validateNextProps(props, nextProps) {
  * @propType {func} [createURL] - Function to call when creating links, useful for [URL Routing](guide/Routing.html).
  * @propType {SearchResults|SearchResults[]} [resultsState] - Use this to inject the results that will be used at first rendering. Those results are found by using the `findResultsState` function. Useful for [Server Side Rendering](guide/Server-side_rendering.html).
  * @propType {number} [stalledSearchDelay=200] - The amount of time before considering that the search takes too much time. The time is expressed in milliseconds.
+ * @propType {boolean} [disableEmptySearch=false] - Flag to activate when searching for empty query is not desired.
  * @propType {{ Root: string|function, props: object }} [root] - Use this to customize the root element. Default value: `{ Root: 'div' }`
  * @example
  * import React from 'react';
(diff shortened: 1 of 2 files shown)


Haroenv:
This will also prevent search on empty query for queries other than the initial. I think a more programmatic solution (like already mentioned by @samouss, not rendering the root e.g., or switching out searchClient for a noop version before loading) seems more long-term useful. 

Thanks for your PR though, definitely!

miguelci:
And if this filter was only referring to the click analytics part @Haroenv ?
i don't think empty searches are very valuable to have logged

Haroenv:
If you only want it for analytics, do this: 

```js
const ConditionalAnalytics = connectStateResults(({ searchState }) => (
  <Configure analytics={Boolean(searchState && searchState.query)} />
));
```

https://codesandbox.io/s/81xo2qrnm9

miguelci:
Thank you for the fast answer @Haroenv .
Was already trying something like that, but always got two requests. It seems that it happens one before the configure (because it's false) and then another (with true).
It seems that the same happens with your code.
Can you confirm?

Haroenv:
In my sandbox I only see one request per query. Can you try to make a replicating sandbox, preferably in a different issue? Thanks!

janza:
Would you still be interested in a PR with an option that disables the initial request only?  
Othewise it might make sense to close #1111 as a _wontfix_ for eventual future contributors.

gianpaj:
Looking forward to this PR being merged 🙃 

_sneaky +1 / friendly nudge to the wonderful maintainers and collaborators_ 😄

Haroenv:
@gianpaj, thanks for the ping, while we definitely agree there are use cases for avoiding the initial search, in the meantime you can use this PR and use it as your own fork, instead of installing it from npm. 

The reason why we don't simply merge this yet is because we'd rather have a more powerful API

petermiles:
What do you think about doing a componentDidUpdate and see if the last query existed, trigger an all results on empty hits, if not, trigger an empty query?

PR #1689, opened 2018-10-25
chore: pin Docker images

samouss:
Recently the E2E tests stop working without any changes on the packages dependencies. The test did pass locally but they only failed on CircleCI. It's because Docker image tags are **mutable**. They can change at any time - it appears that one of the image update breaks one of the tools that we use for the E2E tests. To fix this issue we now pin the Docker images with their digest hash. 

Renovate should take care of the update in a dedicated PR.

> https://renovatebot.com/blog/docker-mutable-tags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ aliases:
 defaults: &defaults
   working_directory: ~/react-instantsearch
   docker:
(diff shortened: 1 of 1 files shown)


Haroenv:
ugh wow mutable tags with semver names 🤦‍♂️

PR #1696, opened 2018-10-26
fix(sffv): clamp maxFacetHits to the allowed range

samouss:
**Summary**

With the current implementation the value used for [`maxFacetHits`](https://www.algolia.com/doc/api-reference/api-parameters/maxFacetHits) is the default limit of the widget or the user provided value. It's fine for most cases, but it happens that the value provided by the user is higher than 100. This leads to an issue because the Algolia engine only accepts values between 1 - 100. When the value is not in that range the API throws an error.

I chose to fix the issue inside InstantSearch rather than in the helper because it feels odd (to me) to change the value on behalf of the user. The helper can be directly manipulate by users so it makes sense to throw an error if the value is invalid. With InstantSearch it's a bit different we use the helper, they don't have any knowledge of it. Since we allow any value for the limit it makes sense to send the correct a value to the API.

The fix clamp the provided value in the range 1, 100. I've also create a default value of 10 for the `maxFacetHits` in case no value is provided. It avoid to gets a `NaN` value from `Math.{min,max}`. It doesn't breaks anything because `10` is the default value on the engine.

**Before**

before

**After**

after

### DIFF
--- a/packages/react-instantsearch-core/src/core/createInstantSearchManager.js
+++ b/packages/react-instantsearch-core/src/core/createInstantSearchManager.js
@@ -272,13 +272,17 @@ export default function createInstantSearchManager({
   }
 
   function onSearchForFacetValues({ facetName, query, maxFacetHits = 10 }) {
+    // The values 1, 100 are the min / max values that the engine accepts.
+    // see: https://www.algolia.com/doc/api-reference/api-parameters/maxFacetHits
+    const maxFacetHitsWithinRange = Math.max(1, Math.min(maxFacetHits, 100));
+
     store.setState({
       ...store.getState(),
       searchingForFacetValues: true,
     });
 
     helper
-      .searchForFacetValues(facetName, query, maxFacetHits)
+      .searchForFacetValues(facetName, query, maxFacetHitsWithinRange)
       .then(
         content => {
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -271,14 +271,18 @@ export default function createInstantSearchManager({
     search();
   }
 
-  function onSearchForFacetValues({ facetName, query, maxFacetHits }) {
+  function onSearchForFacetValues({ facetName, query, maxFacetHits = 10 }) {
+    // The values 1, 100 are the min / max values that the engine accepts.
+    // see: https://www.algolia.com/doc/api-reference/api-parameters/maxFacetHits
+    const maxFacetHitsWithinRange = Math.max(1, Math.min(maxFacetHits, 100));
```
Haroenv: Do we have this issue as well in InstantSearch.js?

samouss: We have kinda the same issue, but it's less visible because we don't use the `showMore` option to conditionally pick the value for `maxFacetHits`.

PR #1773, opened 2018-11-28
Fix Virtual Widget doc

babgyy:
Fixed import of connectMenu.

**Summary**
Previous import was not failing, however the below example was raising following mounting-time error: 
```Uncaught TypeError: Object(...) is not a function```

**Result**
I have found at least two ways to import connectMenu in the repo's source : 
- react-instantsearch/connectors
- react-instantsearch-core

Both seem to work with react-instantsearch v5.3.2.

### DIFF
--- a/docgen/src/guide/Virtual_widgets.md
+++ b/docgen/src/guide/Virtual_widgets.md
@@ -12,7 +12,8 @@ For example an online shop for clothes could have a page like `https://www.cloth
 that shows hoodies and only hoodies:
 
 ```jsx
-import { InstantSearch, SearchBox, connectMenu } from 'react-instantsearch-dom';
+import { InstantSearch, SearchBox } from 'react-instantsearch-dom';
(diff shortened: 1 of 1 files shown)


Haroenv:
The way written in the documentation should work though 🤔 https://unpkg.com/react-instantsearch-dom@5.3.2/dist/es/index.js, as you can see, all of the connectors are exported. Which version are you using that doesn't export it?

samouss:
The import of the package recently change. The sample from the documentation should work. You can find more information about the migration [inside the documentation](https://community.algolia.com/react-instantsearch/guide/Migration_guide_package_import.html). Here is [an example](https://codesandbox.io/s/wq6x33jw58) that import `connectMenu` from the `react-instantsearch-dom` without any issue. Could you provide an example with an import that failed? Here is [a template](https://codesandbox.io/s/github/algolia/create-instantsearch-app/tree/templates/react-instantsearch) to avoid the boilerplate part.

PR #1774, opened 2018-11-29
fix: Connector memory leak from store subscription during server side…

klummy:
… rendering

**Summary**

This fixes memory leaks in server side environments when connectors are rendered on the server. 

The memory leak was caused by the `store.subscribe(cb)` function in the constructor which because the `componentWillUnmount` method (which contains the unsubscribe from store method) is never called on the server, leads to compounding memory leak.

**Result**

Moving the subscription from the constructor to the `componentDidMount` method ensures that the subscription (and subsequent unsubscribe) only happens on the client.

### DIFF
--- a/packages/react-instantsearch-core/src/core/createConnector.js
+++ b/packages/react-instantsearch-core/src/core/createConnector.js
@@ -63,7 +63,7 @@ export default function createConnector(connectorDesc) {
         super(props, context);
 
         const {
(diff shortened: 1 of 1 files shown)


klummy:
Alright @samouss. Closing this.

samouss:
I've sent a PR #1855  to fix the issue.

PR #1823, opened 2018-12-20
test(createManagerResults): avoid to rely on helper internal

samouss:

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.result.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.result.js
@@ -1,5 +1,7 @@
 import createInstantSearchManager from '../createInstantSearchManager';
 
+const flushPendingMicroTasks = () => new Promise(setImmediate);
+
 const createSearchClient = () => ({
   search: jest.fn(() =>
     Promise.resolve({
@@ -21,9 +23,7 @@ const createSearchClient = () => ({
 
 describe('createInstantSearchManager with results', () => {
   describe('on search', () => {
-    it('updates the store on widget lifecycle', () => {
-      expect.assertions(7);
-
+    it('updates the store on widget lifecycle', async () => {
       const searchClient = createSearchClient();
 
       const ism = createInstantSearchManager({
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.result.js
```diff
@@ -1,323 +1,273 @@
...
-    describe('on external updates', () => {
-      it('updates the store and searches', () => {
-        expect.assertions(7);
+      ism.widgetsManager.registerWidget({
+        getSearchParameters: params => params.setQuery('search'),
+        props: {},
+        context: {},
+      });
 
-        const _dispatchAlgoliaResponse = jest.fn(function(state) {
```
tkrugg: it is definitely better now 👍

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.result.js
```diff
@@ -1,323 +1,273 @@
...
-        const ism = createInstantSearchManager({
-          indexName: 'index',
-          initialState: {},
-          searchParameters: {},
-          searchClient: client,
-        });
+describe('createInstantSearchManager with results', () => {
+  describe('on search', () => {
+    it('updates the store on widget lifecycle', () => {
+      expect.assertions(7);
```
tkrugg: same remark as #1824. I don't think it's necessary to have it

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.result.js
```diff
@@ -1,323 +1,273 @@
...
-    describe('on search for facet values', () => {
-      it('updates the store and searches', () => {
-        expect.assertions(4);
-
-        const searchForFacetValues = jest.fn(() =>
-          Promise.resolve({
-            facetHits: 'results',
-          })
-        );
+    it('updates the store on external updates', () => {
```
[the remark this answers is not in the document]
samouss: We only test that `onExternalStateUpdate` trigger a new search. The `nextSearchState` object provided in that case is useless because the registered widgets are not connected to the `searchState`.

PR #1824, opened 2018-12-20
test(createManagerDerived): drop non relevant test + avoid to rely on helper internal

samouss:

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
@@ -1,5 +1,7 @@
 import createInstantSearchManager from '../createInstantSearchManager';
 
+const flushPendingMicroTasks = () => new Promise(setImmediate);
+
 const createSearchClient = () => ({
   search(requests) {
     return Promise.resolve({
@@ -21,9 +23,7 @@ const createSearchClient = () => ({
 });
 
 describe('createInstantSearchManager with multi index', () => {
-  it('updates the store and searches', () => {
-    expect.assertions(5);
-
+  it('updates the store and searches', async () => {
     // <InstantSearch indexName="first">
     //   <SearchBox defaultRefinement="first query 1" />
     //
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
```diff
@@ -1,340 +1,392 @@
...
 
-        ism.widgetsManager.registerWidget({
-          getSearchParameters: params => params.setIndex('second'),
-          context: { multiIndexContext: { targetedIndex: 'second' } },
-          props: { indexName: 'second' },
-        });
+    expect(ism.store.getState().results).toBe(null);
+
+    return Promise.resolve()
+      .then(() => {})
```
tkrugg: Why do you need this?

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
```diff
@@ -1,340 +1,392 @@
...
+        // <Index indexName="second">
+        //   <SearchBox defaultRefinement="second query 2" />
+        // </Index>
+        ism.widgetsManager.getWidgets()[4].getSearchParameters = params =>
+          params.setQuery('second query 2');
+
+        // Simualte an udpate (see `createConnector`)
+        ism.widgetsManager.update();
+      })
+      .then(() => {})
```
tkrugg: since I see this twice, I suppose it's not a typo.
Are you trying to skip a tick?

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
```diff
@@ -1,340 +1,392 @@
...
+      },
+    ]);
+  },
 });
 
-const client = algoliaClient('latency', '249078a3d4337a8231f1665ec5a44966');
-client.search = jest.fn(() => Promise.resolve({ results: [{ hits: [] }] }));
+describe('createInstantSearchManager with multi index', () => {
+  it('updates the store and searches', () => {
+    expect.assertions(5);
```
tkrugg: I don't think you need this, since the promise your return resolves after all your expect have been ran.

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
samouss: Basically every time a `registerWidget` is called [we scheduled an update](url) on the next tick. The manager does not return the promise so we have no ways to execute an instruction once the delay is elapsed. That way we "wait" the next tick. The idea is similar to what Jest does for fake timers. There is [some discussions](https://github.com/facebook/jest/issues/2157) to add this feature to Jest.

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
samouss: I think I've mostly answer to the question inside [the previous comment](https://github.com/algolia/react-instantsearch/pull/1824#discussion_r243545151)

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
```diff
@@ -1,340 +1,371 @@
...
+    // </Index>
+    ism.widgetsManager.registerWidget({
+      getSearchParameters: params => params.setQuery('second query 1'),
+      context: { multiIndexContext: { targetedIndex: 'second' } },
+      props: {},
+    });
+
+    expect(ism.store.getState().results).toBe(null);
+
+    await flushPendingMicroTasks();
```
samouss: @tkrugg

PR #1826, opened 2018-12-21
test(createManagerError): avoid to rely on helper internal

samouss:

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.error.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.error.js
@@ -1,150 +1,258 @@
-import algoliaClient from 'algoliasearch/lite';
 import createInstantSearchManager from '../createInstantSearchManager';
 
-jest.useFakeTimers();
-
-jest.mock('algoliasearch-helper/src/algoliasearch.helper.js', () => {
-  let count = 0;
-  const Helper = require.requireActual(
-    'algoliasearch-helper/src/algoliasearch.helper.js'
-  );
-  Helper.prototype._dispatchAlgoliaResponse = function(state) {
-    if (count > 3) {
-      this.emit('result', { count: count++ }, state);
-    } else {
-      this.emit('error', { count: count++ }, state);
-    }
-  };
-  Helper.prototype.searchForFacetValues = () =>
-    Promise.reject(new Error('error'));
-  return Helper;
+const createSearchClient = () => ({
+  search: jest.fn(() =>
+    Promise.resolve({
+      results: [
+        {
+          hits: [],
+        },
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.error.js
```diff
@@ -1,150 +1,258 @@
...
-    });
 
-    describe('on external updates', () => {
-      it('updates the store and searches', () => {
-        expect.assertions(3);
+      expect(ism.store.getState().error).toBe(null);
+
+      return Promise.resolve()
+        .then()
+        .then()
```
Haroenv: I'm not following why these two promises are necessary 🤔

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.error.js
```diff
@@ -1,150 +1,258 @@
...
-    } else {
-      this.emit('error', { count: count++ }, state);
-    }
-  };
-  Helper.prototype.searchForFacetValues = () =>
-    Promise.reject(new Error('error'));
-  return Helper;
+const createSearchClient = () => ({
+  search: jest.fn(() =>
+    Promise.resolve({
```
tkrugg: maybe you can simplify this to just jest.fn(),
it seems you're always calling mockImplementation, which means it's unused?

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.error.js
```diff
@@ -1,150 +1,258 @@
...
-
-            ism.widgetsManager.update();
-          })
-          .then(() => {})
-          .then(() => {
-            const store1 = ism.store.getState();
-            expect(store1.error).toEqual(null);
-            expect(store1.results).toEqual({ count: 4 });
-          });
+    it('reset the error after a succesful search', () => {
```
tkrugg:
```suggestion
    it('reset the error after a successful search', () => {
```

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.error.js
```diff
@@ -1,150 +1,258 @@
...
+          expect(state.error).toEqual(new Error('API_ERROR_2'));
+          expect(state.results).toEqual(null);
         });
-      });
     });
 
-    describe('reset error after a succesful query', () => {
-      it('on widget lifecyle', () => {
-        expect.assertions(5);
+    it('reset the error after a succesful search', () => {
```
tkrugg:
```suggestion
    it('reset the error after a successful search', () => {
```

PR #1827, opened 2018-12-21
test(createManagerRaw): follow convention adopted in the other tests

samouss:


packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -1,35 +1,64 @@
...
+            nbPages: 0,
+            processingTimeMS: 4,
+            query: '',
+            nbHits: 0,
+            index: 'index',
+          },
+        ],
+      });
+
+      searchResultsPromises.push(results);
```
tkrugg: It seems you kept this variable from `makeManagedClient` but it's not exposed anymore and you don't use it in your tests.
Is it really needed?

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -231,58 +250,55 @@ describe('createInstantSearchManager', () => {
...
+      expect(searchClient.search).toHaveBeenCalledTimes(0);
+      expect(nextSearchClient.search).toHaveBeenCalledTimes(0);
 
-      ism.updateClient(client1);
+      ism.updateClient(nextSearchClient);
 
       return Promise.resolve().then(() => {
-        expect(client0.search).toHaveBeenCalledTimes(0);
-        expect(client1.search).toHaveBeenCalledTimes(1);
+        expect(searchClient.search).toHaveBeenCalledTimes(0);
```
tkrugg: Why is this promise.resolve needed ?
If I look at updateClient implem, it seems search method is called synchronously.
Is the .search method of the helper wrapper the client's search method?

PR #1829, opened 2018-12-21
Change gonna to going to in React getting started

dustincoates:
**Summary**

Change `gonna` to `going to`. In English `gonna` is probably too "slang-y" for what we want to convey.


Haroenv:
This doc is not really the source of the documentation anymore, is it also the case in algolia.com/doc?

PR #1833, opened 2018-12-26
feat(indexId): avoid to rely on the results.index  [PART-1]

samouss:
**Summary**

This is the first part of the work to support multi-index search on the same index. The PR doesn't add any feature to the lib, it's only an internal refactor to allow the support of the same index search. The previous implementation rely on an object like the one below to keep track of which results match which index:

```
{
  <indexNameSearchParameters>: <indexNameReactInstantSearchContext>
}
```

Then inside the success handler of the helper we rely on `content.index` to find which index match which content:

https://github.com/algolia/react-instantsearch/blob/07ad0390e1ca21024289b7f38ce7a7d129a00441/packages/react-instantsearch-core/src/core/createInstantSearchManager.js#L174

This solution doesn't work with a multi-index search on the same index because the value of `content.index` is the same for all the results. We have to use an external value to find back which results match which index.

**Solution**

We now bind the index identifier to the helper success handler. This identifier comes from the outside and it's not bound to the results. We can get rid of the map `indexMapping` because we now use the `indexId`. We also get rid of the `derivedHelpers` object because this information is already inside the helper.

We have the same issue as before though because we use the index name as identifier. But we gonna introduce the `indexId` property in a next PR.

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
@@ -3,16 +3,16 @@ import createInstantSearchManager from '../createInstantSearchManager';
 const runAllMicroTasks = () => new Promise(setImmediate);
 
 const createSearchClient = () => ({
-  search(requests) {
-    return Promise.resolve({
+  search: jest.fn(requests =>
+    Promise.resolve({
       results: requests.map(request => ({
         index: request.indexName,
         query: request.params.query,
         page: request.params.page,
         hits: [],
       })),
-    });
-  },
+    })
+  ),
   searchForFacetValues() {
     return Promise.resolve([
       {
@@ -118,7 +118,6 @@ describe('createInstantSearchManager with multi index', () => {
     ism.widgetsManager.getWidgets()[4].getSearchParameters = params =>
       params.setQuery('second query 2');
 
-    // Simualte an udpate (see `createConnector`)
     ism.widgetsManager.update();
 
     await runAllMicroTasks();
@@ -211,6 +210,70 @@ describe('createInstantSearchManager with multi index', () => {
         page: 0,
       })
     );
+  });
+
+  it('searches with N queries for N Index widgets', async () => {
+    // <InstantSearch indexName="first">
+    //   <Index indexName="first" />
+    //   <Index indexName="second" />
+    //   <Index indexName="third" />
+    //   <Index indexName="four" />
+    // </InstantSearch>;
+
+    const searchClient = createSearchClient();
+
+    const ism = createInstantSearchManager({
+      indexName: 'first',
+      initialState: {},
+      searchParameters: {},
+      searchClient,
+    });
+
+    // <Index indexName="first" />
+    ism.widgetsManager.registerWidget({
+      getSearchParameters: x => x.setIndex('first'),
+      context: {},
+      props: {
+        indexName: 'first',
+      },
+    });
+
+    // <Index indexName="second" />
+    ism.widgetsManager.registerWidget({
+      getSearchParameters: x => x.setIndex('second'),
+      context: {},
+      props: {
+        indexName: 'second',
+      },
+    });
+
+    // <Index indexName="third" />
+    ism.widgetsManager.registerWidget({
+      getSearchParameters: x => x.setIndex('third'),
+      context: {},
+      props: {
+        indexName: 'third',
+      },
+    });
+
+    // <Index indexName="four" />
+    ism.widgetsManager.registerWidget({
+      getSearchParameters: x => x.setIndex('four'),
+      context: {},
+      props: {
+        indexName: 'four',
+      },
+    });
+
+    await runAllMicroTasks();
+
+    expect(searchClient.search.mock.calls[0][0]).toHaveLength(4);
+
+    ism.widgetsManager.update();
+
+    await runAllMicroTasks();
+
+    expect(searchClient.search.mock.calls[1][0]).toHaveLength(4);
   });
 
   it('switching from mono to multi index', async () => {
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
@@ -9,6 +9,7 @@ const createSearchClient = () => ({
         index: request.indexName,
         query: request.params.query,
         page: request.params.page,
+        hitsPerPage: request.params.hitsPerPage,
         hits: [],
       })),
     })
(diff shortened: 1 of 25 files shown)


packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,36 @@ export default function createInstantSearchManager({
...
+      //
+      // - a -> main helper receives results
+      // - ap -> main helper receives results
+      // - app -> main helper + derived helpers receive results
+      //
+      // The quick fix is to avoid to detatch them on search but only once they
+      // received the resutls. But it means that in case of a stalled search
+      // all the derived helper not detached yet register a new search inside
+      // the helper. The number grows fast in case of a bad network and it's
+      // not deterministic.
```
samouss: It was already the case before the code change. I' added a note because I struggle with a weird behaviour for 2 hours because of this.

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
```diff
@@ -3,16 +3,16 @@ import createInstantSearchManager from '../createInstantSearchManager';
 const runAllMicroTasks = () => new Promise(setImmediate);
```
tkrugg: this is to be commited into master?

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,42 @@ export default function createInstantSearchManager({
...
+        // they won't receive intermediate resutls in case of a stalled search.
+        // Only the last results is dispatch by the derived helper beause they
+        // are not detached yet:
+        //
+        // - a -> main helper receives results
+        // - ap -> main helper receives results
+        // - app -> main helper + derived helpers receive results
+        //
+        // The quick fix is to avoid to detatch them on search but only once they
+        // received the resutls. But it means that in case of a stalled search
```
tkrugg:
```suggestion
        // received the results. But it means that in case of a stalled search
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,42 @@ export default function createInstantSearchManager({
...
-      Object.keys(derivedHelpers).forEach(key => derivedHelpers[key].detach());
-      derivedHelpers = {};
+
+      // We have to call slice because the method `detach` on the dervied
+      // helpers mutate the value `derivedHelpers`. The forEach loop does
+      // not iterate on each value and we're not able to correctly clear
+      // the previous derived helpers (memory leak + useless requests).
+      helper.derivedHelpers.slice().forEach(derivedHelper => {
+        // Since we detach the derived helpers on **every** new search
+        // they won't receive intermediate resutls in case of a stalled search.
```
```suggestion
        // they won't receive intermediate results in case of a stalled search.
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,42 @@ export default function createInstantSearchManager({
         mainIndexParameters,
         derivatedWidgets,
       } = getSearchParameters(helper.state);
-      Object.keys(derivedHelpers).forEach(key => derivedHelpers[key].detach());
-      derivedHelpers = {};
+
+      // We have to call slice because the method `detach` on the dervied
```
```suggestion
      // We have to call `slice` because the method `detach` on the derived
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,42 @@ export default function createInstantSearchManager({
...
+      // helpers mutate the value `derivedHelpers`. The forEach loop does
```
```suggestion
      // helpers mutates the value `derivedHelpers`. The `forEach` loop does
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,42 @@ export default function createInstantSearchManager({
...
+      // not iterate on each value and we're not able to correctly clear
+      // the previous derived helpers (memory leak + useless requests).
+      helper.derivedHelpers.slice().forEach(derivedHelper => {
+        // Since we detach the derived helpers on **every** new search
+        // they won't receive intermediate resutls in case of a stalled search.
+        // Only the last results is dispatch by the derived helper beause they
```
```suggestion
        // Only the last results are dispatched by the derived helper because they
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,42 @@ export default function createInstantSearchManager({
...
+        // are not detached yet:
+        //
+        // - a -> main helper receives results
+        // - ap -> main helper receives results
+        // - app -> main helper + derived helpers receive results
+        //
+        // The quick fix is to avoid to detatch them on search but only once they
+        // received the results. But it means that in case of a stalled search
+        // all the derived helper not detached yet register a new search inside
```
```suggestion
        // all the derived helpers not detached yet register a new search inside
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -162,39 +174,44 @@ export default function createInstantSearchManager({
...
-      nextIsSearchStalled = false;
-    }
+  function handleSearchSuccess({ indexId }) {
+    return content => {
+      const state = store.getState();
+      const isDerivedHelpersEmpty = !helper.derivedHelpers.length;
+
+      let results = state.results ? state.results : {};
+
+      /* if switching from mono index to multi index and vice versa,
```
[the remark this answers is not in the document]
samouss: Yep I replaced it in a later PR.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -124,8 +120,6 @@ export default function createInstantSearchManager({
         sharedParameters
       );
 
-    indexMapping[mainIndexParameters.index] = indexName;
-
     return { sharedParameters, mainIndexParameters, derivatedWidgets };
```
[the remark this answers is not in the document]
samouss: It's renamed inside #1840 to `derivedParameters`.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -136,24 +130,42 @@ export default function createInstantSearchManager({
...
+      });
 
       helper.setState(sharedParameters);
 
       derivatedWidgets.forEach(derivatedSearchParameters => {
         const index = derivatedSearchParameters.targetedIndex;
-        const derivedHelper = helper.derive(() => {
-          const parameters = derivatedSearchParameters.widgets.reduce(
+        const derivedHelper = helper.derive(() =>
+          derivatedSearchParameters.widgets.reduce(
```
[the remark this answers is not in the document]
samouss: It's updated inside #1840. It's still called `parameters` but it's inside the parameters of a function to me it comprehensible enough:

```js
derivedParameters.forEach(({ indexId, parameters }) => {
  const derivedHelper = helper.derive(() => parameters);
  // ...
});
```

PR #1835, opened 2018-12-27
feat(indexId): rely on indexId rather than indexName [PART-2]

samouss:
**Summary**

This is the second part of the work to support multi-index search on the same index. The PR introduce a new prop on the `Index` component called `indexId`. This prop is a unique identifier for the index. It has to be unique inside the App to let the manager linked an index, a request and the results. With different identifiers we are able to target the same index with different parameters:

```jsx
<InstantSearch indexName="instant_search" searchClient={searchClient}>
  <SearchBox />

  <Index indexId="instant_search_default" indexName="instant_search">
    <Configure hitsPerPage={3} />
    <Hits />
  </Index>

  <Index indexId="instant_search_apple" indexName="instant_search">
    <Configure hitsPerPage={3} filters="brand:Apple" />
    <Hits />
  </Index>

  <Index indexId="instant_search_samsung" indexName="instant_search">
    <Configure hitsPerPage={3} filters="brand:Samsung" />
    <Hits />
  </Index>
</InstantSearch>
```

The prop `indexId` should be required but it's a breaking change. The solution we came up with is to fallback the `indexId` to the `indexName` when it's not provided (we'll add a warning in a next PR). Then on the next major we can move the prop to be required. When the fallback happen it doesn't have an impact for our users because it's the exact same behaviour as before (`indexName` was the key). 

Internally the value exposes though the `Index` context has changed: it's now the `indexId` not the `indexName`. It should not have an impact neither because the places where we use the value from the context we actually want the identifier (we can expose both of them at some point if needed).

You can test the changes with the updated [multi-index stories](https://deploy-preview-1835--react-instantsearch.netlify.com/react-instantsearch/storybook/?selectedKind=%3CIndex%3E&selectedStory=MultiHits&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs). Only the client side code is working the SSR support will come in a next PR. Otherwise this PR would have embed too much changes.

**Changes**

- **`createInstantSearchManager`**: rely on `props.indexId` rather than `props.indexName`
- **`createIndex`**: provide `indexId` to the `Index` component with a fallback on `indexName`
- **`Index`**: expose the `indexId` through the context rather than `indexName`

### DIFF
--- a/packages/react-instantsearch-core/src/components/Index.js
+++ b/packages/react-instantsearch-core/src/components/Index.js
@@ -1,5 +1,3 @@
-/* eslint valid-jsdoc: 0 */
-
 import React, { Component, Children } from 'react';
 import PropTypes from 'prop-types';
 
--- a/packages/react-instantsearch-core/src/components/Index.js
+++ b/packages/react-instantsearch-core/src/components/Index.js
@@ -1,7 +1,8 @@
+/* eslint valid-jsdoc: 0 */
+
 import React, { Component, Children } from 'react';
 import PropTypes from 'prop-types';
 
-/* eslint valid-jsdoc: 0 */
 /**
  * @description
  * `<Index>` is the component that allows you to apply widgets to a dedicated index. It's
@@ -32,18 +33,17 @@ import PropTypes from 'prop-types';
  * );
  */
 class Index extends Component {
-  constructor(props, context) {
-    super(props);
-    const {
-      ais: { widgetsManager },
-    } = context;
+  constructor(...args) {
+    super(...args);
 
     /*
      we want <Index> to be seen as a regular widget.
      It means that with only <Index> present a new query will be sent to Algolia.
      That way you don't need a virtual hits widget to use the connectAutoComplete.
     */
-    this.unregisterWidget = widgetsManager.registerWidget(this);
+    this.unregisterWidget = this.context.ais.widgetsManager.registerWidget(
+      this
+    );
   }
 
   componentWillMount() {
@@ -67,7 +67,7 @@ class Index extends Component {
   getChildContext() {
     return {
       multiIndexContext: {
-        targetedIndex: this.props.indexName,
+        targetedIndex: this.props.indexId,
       },
     };
   }
@@ -89,6 +89,7 @@ class Index extends Component {
 Index.propTypes = {
   // @TODO: These props are currently constant.
   indexName: PropTypes.string.isRequired,
+  indexId: PropTypes.string.isRequired,
   children: PropTypes.node,
   root: PropTypes.shape({
     Root: PropTypes.oneOfType([
--- a/packages/react-instantsearch-core/src/components/__tests__/Index.js
+++ b/packages/react-instantsearch-core/src/components/__tests__/Index.js
@@ -1,100 +1,160 @@
 import React from 'react';
-import Enzyme, { mount } from 'enzyme';
+import Enzyme, { shallow } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import { SearchParameters } from 'algoliasearch-helper';
 import Index from '../Index';
 
 Enzyme.configure({ adapter: new Adapter() });
 
 describe('Index', () => {
-  const DEFAULT_PROPS = {
-    indexName: 'foobar',
+  const createContext = () => ({
+    ais: {
+      onSearchParameters: jest.fn(),
+      widgetsManager: {
+        registerWidget: jest.fn(),
+        update: jest.fn(),
+      },
+    },
+  });
+
+  const requiredProps = {
+    indexName: 'indexName',
+    indexId: 'indexId',
     root: {
       Root: 'div',
     },
   };
 
-  const registerWidget = jest.fn();
-  const update = jest.fn();
-  const widgetsManager = { registerWidget, update };
+  it('register itself on mount', () => {
+    const context = createContext();
 
-  let context = {
-    ais: {
-      widgetsManager,
-      onSearchParameters: () => {},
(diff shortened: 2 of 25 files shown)


packages/react-instantsearch-core/src/components/Index.js
```diff
@@ -1,7 +1,8 @@
+/* eslint valid-jsdoc: 0 */
```
tkrugg: I read this rule is deprecated so I guess it can be disabled globally?
I'm not sure what replaces it

packages/react-instantsearch-core/src/core/createIndex.js
```diff
@@ -9,14 +9,16 @@ import Index from '../components/Index';
  * @return {object} a Index root
  */
 const createIndex = defaultRoot => {
-  const CreateIndex = ({ indexName, root, children }) => (
-    <Index indexName={indexName} root={root}>
+  const CreateIndex = ({ indexName, indexId, root, children }) => (
+    <Index indexName={indexName} indexId={indexId || indexName} root={root}>
```
tkrugg: is there a reason we require (or we'll require later) the user to provide an indexId?
Can't we just generate them from the indexName and make sure they are unique?

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -187,14 +187,14 @@ export function getCurrentRefinementValue(
 }
 
 export function cleanUpValue(searchState, context, id) {
-  const indexName = getIndex(context);
+  const index = getIndex(context);
```
tkrugg: it seems getIndex is returning an indexId. Shouldn't we make that clear by calling it getIndexId ?

packages/react-instantsearch-core/src/components/__tests__/Index.js
```diff
@@ -1,100 +1,160 @@
...
-        </Index>,
-        { context }
-      );
-    }).not.toThrow();
-
-    expect(registerWidget.mock.calls).toHaveLength(3);
+    expect(context.ais.onSearchParameters).toHaveBeenCalledTimes(1);
+  });
+
+  it('call update if indexName prop change', () => {
```
francoischalifour:
```suggestion
  it('calls update if indexName prop changes', () => {
```

packages/react-instantsearch-core/src/core/__tests__/createIndex.js
```diff
@@ -9,26 +9,76 @@ Enzyme.configure({ adapter: new Adapter() });
...
-    const wrapper = shallow(<CustomIndex indexName="name" root={root} />);
+    const wrapper = shallow(<CustomIndex {...props} />);
 
-    expect(wrapper.is(Index)).toBe(true);
-    expect(wrapper.props().root).toEqual(root);
-    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.props().indexId).toBe('indexId');
+  });
+
+  it("expect to create an Index with an indexId that fallback to indexName when it's not provided", () => {
```
francoischalifour: This doesn't seem to be prettified (double quotes).

packages/react-instantsearch-core/src/core/createIndex.js
tkrugg: yes, only on this layer, but it's required in `<Index />` and my point what that if we generate indexId we won't need to expose it and anticipate the case it isn't provided.

Discussed this live with @samouss. It turns out we can't because
- indexIds do not have to be unique 
- there are use cases where user would want to reuse an indexId

samouss: Yep like @tkrugg said: it can be re-used in some cases. On [the latest PR](https://github.com/algolia/react-instantsearch/pull/1844) I asked about the fact to put it required or not. I'm not sure it brings value to put it required, it brings more constraints to user that won't use it.

packages/react-instantsearch-core/src/core/indexUtils.js
samouss: Yep I hesitate to update it: the first reason why I didn't do it in this PR it's because it adds more changes to the diff. The other reason is linked to the situation describes above:

> Internally the value exposes though the `Index` context has changed: it's now the `indexId` not the `indexName`. It should not have an impact neither because the places where we use the value from the context we actually want the identifier (we can expose both of them at some point if needed).

I didn't change it to avoid confusion inside the connectors implementations that use the value return by the function. But it seems that it brings more confusion to use `getIndex` rather than `getIndexId`. I'll change it in a dedicated PR.

packages/react-instantsearch-core/src/components/Index.js
```diff
@@ -89,6 +89,7 @@ class Index extends Component {
 Index.propTypes = {
   // @TODO: These props are currently constant.
   indexName: PropTypes.string.isRequired,
+  indexId: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
samouss: Yes it is. The prop is not required for `createIndex` but it is for `Index`. This component is not exposes through the public API. It's value returned by [`createIndex`](https://github.com/algolia/react-instantsearch/blob/feat/part-2-index-id/packages/react-instantsearch-core/src/core/createIndex.js) that is exposed not the `Index` component itself.

samouss:
> The tests first word should be using the third-person singular (`expect` → `expects`).

Done.

> What do you think of the name `indexKey` instead of `indexId`?

I didn't want to use the word `key` to avoid confusion with the React term. The `key` have to be unique among siblings it's not the case with the `indexId`. The value can be re-used to share an `Index` at the same level.

PR #1840, opened 2018-12-28
feat(indexId): refactor createInstantSearchManager [PART-3]

samouss:
**Summary**

This PR does not introduce new feature, it's a refactor of `createInstantSearchManager`. The logic to detect shared, main and derived parameters was hard to follow. We tried to put name on this logic to help the reader to follow what happens. Now all search parameters are computed inside the same function, previously it was spread across `getSearchParameters` and `search`.

**Changes**

- **`createInstantSearchManager`**: exposes the `getSearchParameters` function to be able to test it. We don't use this function elsewhere inside the codebase but since the logic is "sensitive" it's a viable trade off IMO.
- **`createInstantSearchManager`**: the derived search parameters are now computed directly inside `getSearchParameters`. The main impact is that the shared parameters don't have to be exposed anymore.
- **`createInstantSearchManager`**: the helper creation is now inlined, we removed the useless step that create the `SearchParameters`. The [third parameters of the helper](https://github.com/algolia/algoliasearch-helper-js/blob/5d7d915331c3b8782114be393171268e5489c6c7/src/algoliasearch.helper.js#L129) is an object that is used to create the state.

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
@@ -150,6 +150,237 @@ describe('createInstantSearchManager', () => {
     });
   });
 
+  describe('getSearchParameters', () => {
+    it('expect a widget top level to be shared between main and derived parameters', () => {
+      // <InstantSearch indexName="index">
+      //   <SearchBox defaultRefinement="shared" />
+      //   <Index indexId="main" indexName="main" />
+      // </InstantSearch>
+
+      const ism = createInstantSearchManager({
+        indexName: 'index',
+        searchClient: createSearchClient(),
+      });
+
+      // <SearchBox defaultRefinement="shared" />
+      ism.widgetsManager.registerWidget({
+        getSearchParameters(state) {
+          return state.setQuery('shared');
+        },
+        context: {},
+        props: {},
+      });
+
+      // <Index indexId="main" indexName="main" />
+      ism.widgetsManager.registerWidget({
+        getSearchParameters(state) {
+          return state.setIndex('main');
+        },
+        props: {
+          indexId: 'main',
+        },
+      });
+
+      const { mainParameters, derivedParameters } = ism.getSearchParameters();
+
+      expect(mainParameters).toEqual(
+        expect.objectContaining({
+          index: 'index',
+          query: 'shared',
+        })
+      );
+
+      expect(derivedParameters).toEqual([
+        {
+          indexId: 'main',
+          parameters: expect.objectContaining({
+            index: 'main',
+            query: 'shared',
+          }),
+        },
+      ]);
+    });
+
+    it('expect a widget with the same id than the indexName to be a main parameters', () => {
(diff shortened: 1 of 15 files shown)


packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -150,6 +150,237 @@ describe('createInstantSearchManager', () => {
...
+          indexId: 'main',
+          parameters: expect.objectContaining({
+            index: 'main',
+            query: 'shared',
+          }),
+        },
+      ]);
+    });
+
+    it('expect a widget with the same id than the indexName to be a main parameters', () => {
```
tkrugg: is it not a little confusing that `indexId` on the `<Index />` component maps to `indexName` on `<InstantSearch />`, while  `<Index />` has an `indexName` prop, too?

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -1,8 +1,20 @@
...
-import algoliasearchHelper, { SearchParameters } from 'algoliasearch-helper';
+import { omit } from 'lodash';
+import algoliasearchHelper from 'algoliasearch-helper';
 import createWidgetsManager from './createWidgetsManager';
 import createStore from './createStore';
 import { HIGHLIGHT_TAGS } from './highlight';
+import { hasMultipleIndex } from './indexUtils';
+
+const isMultiIndexContext = widget => hasMultipleIndex(widget.context);
+const isTargetedIndexEqualIndex = (widget, index) =>
```
tkrugg: I think if you called it indexId it would make it more clear what it is.

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
samouss: Yes it's intended. We test that the actual value of the index that we want to target is the `indexName` of the `Index` component not the one of the `InstantSearch` component.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -1,8 +1,20 @@
-import { omit, find } from 'lodash';
-import algoliasearchHelper, { SearchParameters } from 'algoliasearch-helper';
+import { omit } from 'lodash';
+import algoliasearchHelper from 'algoliasearch-helper';
 import createWidgetsManager from './createWidgetsManager';
 import createStore from './createStore';
 import { HIGHLIGHT_TAGS } from './highlight';
+import { hasMultipleIndex } from './indexUtils';
```
Haroenv: could be called `hasMultipleIndices`

samouss: Will do it in another PR this function is used in multiple places.

samouss: Fixed inside https://github.com/algolia/react-instantsearch/pull/1852

PR #1842, opened 2018-12-28
feat(indexId): rely on indexId inside SSR [PART-4]

samouss:
**Summary**

This is the fourth part of the work to support multi-index search on the same index. The PR updates how the search parameters are computes inside `createInstantSearchServer`. We now rely on the `indexId` rather than `indexName`.

**Changes**

- **`createInstantSearchServer`**: the value returned by `findResultsState` have now an `indexId` to be able to reconstruct the result on the client side. Previously we were relying on the `indexName`
 we can't anymore because the same index can be used multiple times.

**Result**

single-index

multi-index


tkrugg:
On your gif you can see a query triggered after the first render. I would suppose it's unneeded on SSR.
Is it hard/impossible to prevent?

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -197,18 +199,52 @@ describe('createInstantSearchServer', () => {
...
 
       const wrapper = shallow(<InstantSearch {...props} />);
 
-      expect(wrapper.props().resultsState.getFacetByName).toBeDefined();
+      expect(wrapper.props().resultsState).toBeInstanceOf(SearchResults);
       expect(wrapper.props().resultsState.query).toEqual('query');
     });
 
     describe('find results', () => {
+      it('results shoud be SearchResults and SearchParameters from the helper', async () => {
```
tkrugg: wording `should be instances of ...` would better match the assertions made

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -601,14 +646,63 @@ describe('createInstantSearchServer', () => {
...
+        expect(first.state.query).toBe('iPhone');
+
+        const [, second] = data;
+
+        expect(second._internalIndexId).toBe('index1_with_refinement');
+        expect(second.state.index).toBe('index1');
+        expect(second.state.query).toBe('iPad');
+      });
     });
   });
```
tkrugg: shouldn't we test the behaviour of this too?
```js
     <InstantSearch {...props}>
         <Index indexId="index1" indexName="index1" /> 
         <Index indexId="index1" indexName="index2" />
      </InstantSearch>	      

```

samouss: At the moment nothing prevents this usage of the API. The latest component mounted on the page will have the "advantage" (we can image the merge like an `Object.assign`).

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -299,32 +335,82 @@ describe('createInstantSearchServer', () => {
...
       expect(wrapper.props().resultsState.index1.query).toBe('query1');
-      expect(wrapper.props().resultsState.index1.getFacetByName).toBeDefined();
+      expect(wrapper.props().resultsState.index1).toBeInstanceOf(SearchResults);
       expect(wrapper.props().resultsState.index2.query).toBe('query2');
-      expect(wrapper.props().resultsState.index2.getFacetByName).toBeDefined();
+      expect(wrapper.props().resultsState.index2).toBeInstanceOf(SearchResults);
     });
 
     describe('find results', () => {
+      it('results shoud be instance of SearchResults and SearchParameters', async () => {
```
Haroenv:
```suggestion
      it('results should be instance of SearchResults and SearchParameters', async () => {
```

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -1,4 +1,4 @@
-import { isEmpty } from 'lodash';
+import { isEmpty, zipWith } from 'lodash';
```
Haroenv: I think this is the only place in the code base using `zipWith`, so it would increase bundle size I think

PR #1851, opened 2019-01-03
feat(indexId): rename getIndex -> getIndexId [PART-6]

samouss:
**Summary**

Follow up to https://github.com/algolia/react-instantsearch/pull/1835#discussion_r244946397


Haroenv:
I guess this is breaking somehow that the `index` key in the widgets now returns the id instead of the name?

samouss:
It's not a "breaking change" because it has the exact same behaviour for a "regular" mono and multi-index search. It's only for a multi-index search that users might realise that they have now the `indexId` rather than the `indexName`. At that time we can introduce `indexId` and `indexName` inside the metadata. But for now I think this behaviour is fine.

PR #1855, opened 2019-01-04
refactor(createConnector): update subscriptions mechanism

samouss:
**Summary**

The PR changes the way we subscribe to the `store` and `widgetManager`. Until now the subscriptions was done inside the constructor. It leads to some issues with our implementation. 

The first issue occurs with SSR. We have a callback (`onSearchParameters`) to collect and compute search parameters on the server. The callback is called inside `componentWillMount` only when it's provided. The issue is that the constructor is also executed on the server. It means that we trigger at least two requests on the server: one scheduled by the SSR (this is expected) and the others by the manager (this is **not** expected). The latter is not expected because we have no way to "wait" for it and use its results to re-render the App. Only the request schedules by the SSR is required.

The other issue comes from the `canRender` prop. This prop is used to detect whether or not we are on the client or the server. The value is initialised to `false` in the constructor and then we call `setState` inside `componentDidMount`. This lifecycle is only called on the client. It means that we are now safe to access browser API like `window.location`. The issue with it is that we schedule an extra useless render to set the value to `true`. React triggers a warning inside the flame graph (see below).

**Solution**

To solve both issues we can move the subscription from the constructor to `componentDidMount`. The listeners are now only attached on the client once the component is actually mounted on the page. It solves the first issue because the manager doesn't schedule a new request on the server, only the one scheduled by the SSR is sent. It solves also our second issue because now the listeners are called on the client only. It means that we can provide directly the value to `canRefine` rather than relying on internal state. It's also the recommended way to handle subscriptions [by the React team](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#adding-event-listeners-or-subscriptions).

The `canRender` prop is almost useless though because we can rely on something else to detect whether or not we are on the server. We dropped his usage inside the connectors but we have to keep it until the next major inside `createConnector`. It's a public API so every users that rely on this value inside a custom connector could be impacted.

**Result**

screenshot 2019-01-02 at 12 02 19

screenshot 2019-01-02 at 11 59 55

### DIFF
--- a/packages/react-instantsearch-core/src/core/createConnector.js
+++ b/packages/react-instantsearch-core/src/core/createConnector.js
@@ -70,7 +70,7 @@ export default function createConnector(connectorDesc) {
             ...props,
             // @MAJOR: We cannot drop this beacuse it's a breaking change. The
             // prop is provided to `createConnector.getProvidedProps`. All the
-            // custom connector are impacted by this change. It should be fine
+            // custom connectors are impacted by this change. It should be fine
             // to drop it in the next major though.
             canRender: false,
           }),
--- a/packages/react-instantsearch-core/src/core/createConnector.js
+++ b/packages/react-instantsearch-core/src/core/createConnector.js
@@ -59,6 +59,8 @@ export default function createConnector(connectorDesc) {
         multiIndexContext: PropTypes.object,
       };
 
+      isUnmounting = false;
+
       constructor(props, context) {
         super(props, context);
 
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch-core/src/core/createConnector.js
```diff
@@ -59,31 +59,23 @@ export default function createConnector(connectorDesc) {
...
-        } = context;
-        const canRender = false;
         this.state = {
-          props: this.getProvidedProps({ ...props, canRender }),
-          canRender, // use to know if a component is rendered (browser), or not (server).
+          props: this.getProvidedProps({
+            ...props,
+            // @MAJOR: We cannot drop this beacuse it's a breaking change. The
+            // prop is provided to `createConnector.getProvidedProps`. All the
+            // custom connector are impacted by this change. It should be fine
```
francoischalifour:
```suggestion
            // custom connectors are impacted by this change. It should be fine
```

klummy:
Hey @samouss,

Nice work on the PR! Please, what's the ETA for publishing a version containing these changes to NPM?

Thanks

samouss:
I’ll release a beta version mid next week. I’ll keep you posted.

samouss:
The change has been released with the version [5.4.0-beta.0](https://discourse.algolia.com/t/react-instantsearch-v5-4-0-beta-0-is-out/6859).

klummy:
Thanks @samouss

PR #1859, opened 2019-01-04
chore: move on Babel 7 + Storybook 4 (WIP)

samouss:
**Summary**

This PR moves the project to `babel@7` and `storybook@4`. We did both at the same time otherwise it was a nightmare to support both at the same time. Some places use `babel@6`:

- examples: it can be done in a later PR

The bundle size increases because Babel sticks more to the spec now. It will be fixed in a later PR where we gonna use `babel/runtime` to re-use helpers. The size should be smaller than before this pull request.

At some point we'll hoist Babel, Rollup, Jest configurations at the top to share the logic. At first we didn't do it because it's was complex to setup but we'll do it at some point it'll simplify a lot how we track the dependencies and manage the project.


samouss:
We'll open many PRs for this migration.

PR #1879, opened 2019-01-09
fix: sync algoliasearch version

samouss:
**Summary**

Without this change the UMD build of the legacy package is broken. There are a `require` calls inside the build. With this change they disappear from the output.

With issue:

> https://unpkg.com/react-instantsearch@5.4.0-beta.0/umd/Dom.js

Without issue:

> https://unpkg.com/react-instantsearch@5.3.2/umd/Dom.js


samouss:
The issue is now fixed with the version `v5.4.0-beta.1`.

> https://unpkg.com/react-instantsearch@5.4.0-beta.1/umd/Dom.js

PR #1888, opened 2019-01-10
chore(babel): migration to Babel 7 [PART-1]

samouss:
**Summary**

This PR moves packages, Storybook, docgen to Babel 7. Only the examples are still built with Babel 6 but they are outside of the workspaces. We can update them later. The major change is that we now use the [project-wide configuration](https://babeljs.io/docs/en/config-files#project-wide-configuration) `babel.config.js` rather than scoped `.babelrc`. We only have one point of contact for the all projects. We can write overrides per files / folders to apply a specific set of plugins to the matches (like we do for `maps` package).

The downside of this approach is that since we called `babel` from a nested level we have to provide a flag to tell Babel to find the proper configuration. Previously it was automatic, but not anymore. There is mode to call [`upward`](https://babeljs.io/docs/en/options#rootmode) to walks and find the correct configuration. Note that we are gonna hoist all the build step at the top of the package so this mode won't be used anymore at some point (at least for the `packages`).

Last downside of this change is the bundle size, since [Babel 7 is closer to the spec](https://babeljs.io/docs/en/v7-migration#spec-compliancy) it generates more helpers to actually check if the code is valid or not. There is a `loose` mode to avoid this but we can get rid of this increase with `@babel/runtime`. It's implemented in the next PR.


packages/react-instantsearch-dom-maps/package.json
```diff
@@ -34,8 +34,8 @@
     "clean": "rm -rf dist",
     "watch": "yarn build:cjs --watch",
     "build": "yarn build:cjs && yarn build:es && yarn build:umd",
-    "build:cjs": "babel src --out-dir dist/cjs --ignore __tests__,__mocks__ --quiet",
-    "build:es": "BABEL_ENV=es babel src --out-dir dist/es --ignore __tests__,__mocks__ --quiet",
+    "build:cjs": "babel src --root-mode upward --out-dir dist/cjs --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
+    "build:es": "BABEL_ENV=es babel src --root-mode upward --out-dir dist/es --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
     "build:umd": "BABEL_ENV=rollup rollup -c rollup.config.js",
```
francoischalifour: `BABEL_ENV` was already named like that but why `rollup` instead of `umd`?

`rollup` seems to be a detail, we want to detect if it's a UMD bundle.

samouss: It doesn't matter to me, we can rename it.

PR #1889, opened 2019-01-10
chore(babel): setup @babel/runtime [PART-2]

samouss:
**Summary**

This PR leverages the plugin [`@babel/plugin-transform-runtime`](https://babeljs.io/docs/en/babel-plugin-transform-runtime) and [`@babel/runtime`](https://babeljs.io/docs/en/babel-runtime) to avoid to inline the Babel helpers for each files. It rather rely on dependency to re-use them as much as possible.

### DIFF
--- a/packages/react-instantsearch-core/package.json
+++ b/packages/react-instantsearch-core/package.json
@@ -39,6 +39,7 @@
     "release:beta": "yarn clean && yarn build && yarn publish --tag beta --non-interactive"
   },
   "dependencies": {
+    "@babel/runtime": "^7.1.2",
     "algoliasearch-helper": "^2.26.0",
     "lodash": "^4.17.4",
     "prop-types": "^15.5.10"
--- a/packages/react-instantsearch-core/package.json
+++ b/packages/react-instantsearch-core/package.json
@@ -33,7 +33,7 @@
     "build": "yarn build:cjs && yarn build:es && yarn build:umd",
     "build:cjs": "babel src --root-mode upward --out-dir dist/cjs --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
     "build:es": "BABEL_ENV=es babel src --root-mode upward --out-dir dist/es --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
-    "build:umd": "BABEL_ENV=es rollup -c rollup.config.js",
+    "build:umd": "BABEL_ENV=rollup rollup -c rollup.config.js",
     "test": "jest",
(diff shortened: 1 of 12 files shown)


packages/react-instantsearch-core/package.json
```diff
@@ -33,12 +33,13 @@
...
     "build:cjs": "babel src --root-mode upward --out-dir dist/cjs --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
     "build:es": "BABEL_ENV=es babel src --root-mode upward --out-dir dist/es --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
-    "build:umd": "BABEL_ENV=es rollup -c rollup.config.js",
+    "build:umd": "BABEL_ENV=rollup rollup -c rollup.config.js",
     "test": "jest",
     "release": "yarn clean && yarn build && yarn publish --non-interactive",
     "release:beta": "yarn clean && yarn build && yarn publish --tag beta --non-interactive"
   },
   "dependencies": {
+    "@babel/runtime": "^7.1.2",
```
Haroenv: is there some way a user might end up with multiple babel runtimes because it's a dependency?

packages/react-instantsearch-core/package.json
```diff
@@ -33,12 +33,13 @@
     "build": "yarn build:cjs && yarn build:es && yarn build:umd",
     "build:cjs": "babel src --root-mode upward --out-dir dist/cjs --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
     "build:es": "BABEL_ENV=es babel src --root-mode upward --out-dir dist/es --ignore '**/__tests__/**/*.js','**/__mocks__/**/*.js' --quiet",
-    "build:umd": "BABEL_ENV=es rollup -c rollup.config.js",
+    "build:umd": "BABEL_ENV=rollup rollup -c rollup.config.js",
```
francoischalifour: See https://github.com/algolia/react-instantsearch/pull/1888/files#r247415753.

packages/react-instantsearch-core/package.json
samouss: Yes it's a possibility like any other dependencies. I've used an older version to minimize this. But like with any other dependencies it might happen. The behaviour that we had previously is the same as having multiple `@babel/runtime` since the helpers are inlined inside the source code.

samouss:
> Looks good, smaller change in bundle size than I expected? Is it because gz takes most of the duplication?

It depends for which bundle, the one have the most duplication is `Dom.js` and the gain is more than 4kb gzipped it's not that bad for a small change like that.

PR #1914, opened 2019-01-16
chore(typescript): setup Babel, Rollup, ESLint [PART-1]

samouss:
**Summary**

This is the first PR to setup TypeScript into the project. Here is the notable changes:

- TypeScript intall
- allow Babel to transpile TypeScript file
- allow Rollup to resolve / transpile TypeScript file
- allow ESLint to resolve TypeScript file
- move Babel & Rollup dependencies top level

There is at least one thing that we have to setup on the TypeScript side: create the definition file. It can be done later once we have a proper coverage of the project.


packages/react-instantsearch-core/rollup.config.js
```diff
@@ -17,12 +17,14 @@ const createLicence = name =>
...
   babel({
     exclude: ['../../node_modules/**', 'node_modules/**'],
+    extensions: ['.js', '.ts', '.tsx'],
     rootMode: 'upward',
     runtimeHelpers: true,
   }),
   resolve({
     browser: true,
     preferBuiltins: false,
+    extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json'],
```
Haroenv: do we actually use `.mjs`?

tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "target": "ESNEXT",
+    "module": "es2015",
+    "moduleResolution": "node",
+    "jsx": "preserve",
+    "noEmit": true,
+    "strict": true,
+    // It's mandatory to disable this strict check at the moment otherwise we
+    // have to type every import we do. The migration is an incremental process
+    // so it happens that we import JavaScript module from TypeScript. The
+    // compiler translates those imports as impliciy any.
```
Haroenv: add an `@TODO`?

packages/react-instantsearch-core/rollup.config.js
samouss: Nope but a dependency could use it.

samouss:
> I'd expect a `tsc` call in the test config, are you doing that in another PR?

Yes in another PR because you have without any TypeScript file the built fail.

PR #1918, opened 2019-01-17
chore(typescript): setup TSLint [PART-3]

samouss:
**Summary**

This is the third PR to setup TypeScript into the project. Here is the notable changes:

- setup TSLint - we extend from the recommended preset. I've not create a configuration from scratch. I've just disabled some rule that are annoying IMO. We can refine them later though. We'll also see how to externalise this configuration across flavours.


tslint.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": ["tslint:recommended", "tslint-react", "tslint-config-prettier"],
+  "rulesDirectory": ["tslint-plugin-prettier"],
+  "rules": {
+    // Avoid to always prefix interface with `IXxxx` but allow `Xxxx`
```
francoischalifour:
```suggestion
    // Avoid always prefixing interfaces with `IXxxx` but allow `Xxxx`
```

tslint.json
```diff
@@ -0,0 +1,15 @@
...
+    "interface-name": [true, "never-prefix"],
+    // Interface are preferred for exported types but type litreals are very
```
francoischalifour:
```suggestion
    // Interfaces are preferred for exported types but type literals are very
```

tslint.json
```diff
@@ -0,0 +1,15 @@
...
+    // handy to define "private" simple type like  `Props` inside a component.
```
francoischalifour:
```suggestion
    // handy to define "private" simple types like `Props` inside a component.
```

francoischalifour:
Do you have any ideas in mind for sharing this configuration across projects?

samouss:
Yes we can create an external package like we did for [`eslint-config-algolia`](https://github.com/algolia/eslint-config-algolia). It might be possible to do the same for the `tsconfig.json` too. I'll probably give it a try once everything is merged.

PR #1919, opened 2019-01-17
chore(typescript): setup Storybook [PART-4]

samouss:
**Summary**

This is the fourth PR to setup TypeScript into the project. Here is the notable changes:

- allow Storybook to resolve TypeScript module
- enable TypeScript compilation in the CI
- move to non pinned dependencies for types - we might have duplicate identifier otherwise that trigger errors at build time.


renovate.json
```diff
@@ -3,5 +3,13 @@
     "config:js-lib",
     ":automergeMinor",
     ":automergePr"
+  ],
+  "packageRules": [
+    {
+      "packagePatterns": [
+        "@types/*"
+      ],
+      "rangeStrategy": "replace"
```
francoischalifour: `"replace"` seems to already be the default: https://renovatebot.com/docs/configuration-options/#rangestrategy

PR #1920, opened 2019-01-17
chore(typescript): migrate Hits [PART-5]

samouss:
**Summary**

This is the last part that actually migrate one module. The type is far from perfect because this component have to probably be generic. But we can refine this later once we have the connector setup, etc...


packages/react-instantsearch-dom/src/components/Hits.tsx
```diff
@@ -0,0 +1,63 @@
...
+      {hits.map(hit => (
+        <li className={cx('item')} key={hit.objectID}>
+          <HitComponent hit={hit} />
+        </li>
+      ))}
+    </ul>
+  </div>
+);
+
+const DefaultHitComponent: React.SFC<HitProps> = props => (
```
Haroenv:
```suggestion
const DefaultHitComponent: React.FC<HitProps> = props => (
```

packages/react-instantsearch-dom/src/components/Hits.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const cx = createClassNames('Hits');
+
+const Hits: React.SFC<Props> = ({
+  hits,
+  className = '',
+  hitComponent: HitComponent = DefaultHitComponent,
+}) => (
+  <div className={classNames(cx(''), className)}>
+    <ul className={cx('list')}>
```
Haroenv: I thought this was an `ol` now? Did we not do that breaking change on React InstantSearch yet?

samouss: Nope the change came after the release of the V5 of React InstantSearch.

packages/react-instantsearch-dom/src/components/Hits.tsx
```diff
@@ -0,0 +1,63 @@
...
+    ...
+  </div>
+);
+
+const HitPropTypes = PropTypes.shape({
+  objectID: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+    .isRequired,
+});
+
+Hits.propTypes = {
```
francoischalifour: How do we deal with these proptypes redundancy?

samouss: Right now we type them since we have to use them for non TypeScript users. We can probably generate them automatically with a plugin but for now it's fine to type them twice. TypeScript manage to help you (sometimes) when there is a mismatch between the two.

packages/react-instantsearch-dom/src/components/__tests__/Hits.tsx
```diff
@@ -1,25 +1,22 @@
...
+  type Props = { hit: any };
 
-  Hit.propTypes = {
-    hit: PropTypes.object,
-  };
+  const Hit = ({ hit }: Props) => <div id={hit.objectID} />;
 
   it('accepts a hitComponent prop', () => {
-    const hits = [{ objectID: 0 }, { objectID: 1 }, { objectID: 2 }];
+    const hits = [{ objectID: '0' }, { objectID: '1' }, { objectID: '2' }];
```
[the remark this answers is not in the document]
samouss: For now particular reason, mostly because at the time I've changed that I only allow the type `string`. I'll revert.

PR #1721, opened 2018-11-04
Add HeatmapLayer component

fitimvata:
HeatmapLayer  [Google docs](https://developers.google.com/maps/documentation/javascript/heatmaplayer)

**Summary**

In the company where we worked we wanted to build a heatmap of users with algolia.
I saw in the source code that react-instantsearch did not have one, so I made the fork and created it.

**Result**

heatmap-layer

PR #2000, opened 2019-02-08
feat(DOMMaps): expose withGoogleMaps HOC [PART-1]

samouss:
**Summary**

This PR introduces a new HOC for the maps package. It's now part of the public API because we export this component. The HOC gives access to the `google` object and the `googleMapsInstance` which is the instance of the map. 

Users might want to leverage this to create their own layer on top of our geo search widget. It was already the case with the [heatmap](https://github.com/algolia/react-instantsearch/pull/1721) PR. Now it's possible to build it in user-land.

We choose the HOC over a render prop because it's easier to build a component with lifecycle hooks out of an HOC. With a render prop you have to create a layer to extract the value and provide it to the component that will use it. We can built an HOC out of render prop so if someone ask for it we can refactor the API to provide both.

**Usage**

```jsx
const Heatmap = withGoogleMaps({ google, googleMapsInstance }) => null;

const App = () => (
  <GeoSearch google={google}>
    {() => (
      <Fragment>
        <Heatmap />
      </Fragment>
    )}
  </GeoSearch>
);
```

### DIFF
--- a/packages/react-instantsearch-dom-maps/src/withGoogleMaps.tsx
+++ b/packages/react-instantsearch-dom-maps/src/withGoogleMaps.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
+type Subtract<T, K> = Omit<T, keyof K>;
+
+// @TODO: Move to GoogleMaps once it's migrated to TypeScript
+export interface GoogleMapsContext {
+  __ais_geo_search__google_maps__: {
+    google: typeof google;
+    instance: google.maps.Map;
+  };
+}
+
+export interface WithGoogleMapsProps {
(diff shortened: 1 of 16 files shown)


packages/react-instantsearch-dom-maps/src/withGoogleMaps.tsx
```diff
@@ -0,0 +1,50 @@
...
+    instance: google.maps.Map;
+  };
+}
+
+export interface WithGoogleMapsProps {
+  google: typeof google;
+  googleMapsInstance: google.maps.Map;
+}
+
+const withGoogleMaps = <Props extends WithGoogleMapsProps>(
```
Haroenv: naming is somewhat inconsistent, but it makes sense

samouss: You mean `withXXX` vs `connectXXX` ?

PR #2002, opened 2019-02-08
feat(DOMMaps): use withGoogleMaps for Marker [PART-3]

samouss:
**Summary**

This PR updates the `Marker` implementation to leverage `withGoogleMaps` HOC.


Haroenv:
bundle size will need to be updated

samouss:
> bundle size will need to be updated

Not sure I think the last PR will have a size that match the current limit.

PR #2004, opened 2019-02-08
feat(DOMMaps): use withGoogleMaps for Control [PART-5]

samouss:
**Summary**

This PR updates the `Control` implementation to leverage `withGoogleMaps` HOC.

### DIFF
--- a/packages/react-instantsearch-dom-maps/src/Control.js
+++ b/packages/react-instantsearch-dom-maps/src/Control.js
@@ -2,12 +2,13 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { createClassNames, translatable } from 'react-instantsearch-dom';
 import { STATE_CONTEXT } from './Provider';
-import { GOOGLE_MAPS_CONTEXT } from './GoogleMaps';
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch-dom-maps/src/Control.js
```diff
@@ -69,4 +62,4 @@ export class Control extends Component {
 export default translatable({
   control: 'Search as I move the map',
   redo: 'Redo search here',
-})(Control);
+})(withGoogleMaps(Control));
```
Haroenv: complex bracketing here, but is expected I guess. Don't we usually write this in the other order?

samouss: The order doesn't matter, there is no dependency between the two HoCs. The proper way to write this would be with a `compose` function.

```js
export default compose(
  withGoogleMaps,
  translatable({ ... })
)(Control)
```

Haroenv: the confusion was that `withGoogleMaps`. isn't `() => () => () =>`, but `() => () =>`, so it doesn't need its own invocation for options.

PR #2019, opened 2019-02-12
chore(website): drop docgen [PART-9]

samouss:
**Summary**

This PR drops the `docgen` website! It also removes all the occurrences from, `gitignore`, `eslintignore`, etc...


Haroenv:
Don't we use these image somewhere? (readme or contributing) Or has that already been changed

samouss:
> Don't we use these image somewhere? (readme or contributing) Or has that already been changed

It has already been changed.

Haroenv:
I wonder if it's the babel change, but the bundles are significantly bigger, did you notice?

samouss:
Yep I suspect it's because of the workspaces. I removed a lot of dependencies with `docgen` the `yarn.lock` changed a lot. It's possible that we have duplicate dependencies.

francoischalifour:
The average bundle size has still increased a tiny bit after the deduplication of Lodash.

Any idea why?

samouss:
The version of lodash changed from `4.17.5` to `4.17.11` that might explain why.

PR #2023, opened 2019-02-13
fix(DOMMaps): use new React Context API

francoischalifour:
**Summary**

This updates the `react-instantsearch-dom-maps` package to the new React Context API introduced in 16.3.0.

**Result**

Check the [stories](https://deploy-preview-2023--react-instantsearch.netlify.com/storybook/?selectedKind=GeoSearch).


packages/react-instantsearch-dom-maps/src/__tests__/CustomMarker.js
```diff
@@ -554,7 +554,7 @@ describe('CustomMarker', () => {
     });
   });
 
-  describe('Connected', () => {
+  describe.skip('Connected', () => {
```
francoischalifour: I'm not sure how to update this test suit.

packages/react-instantsearch-dom-maps/src/__tests__/Marker.js
```diff
@@ -238,7 +238,7 @@ describe('Marker', () => {
     });
   });
 
-  describe('Connected', () => {
+  describe.skip('Connected', () => {
```
francoischalifour: I'm not sure how to update this test suit.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -9,7 +9,7 @@ import GoogleMaps from '../GoogleMaps';
 
 Enzyme.configure({ adapter: new Adapter() });
 
-describe('GoogleMaps', () => {
+describe.skip('GoogleMaps', () => {
```
francoischalifour: This won't be trivial to update.

We now have an inner functional component, wrapped by the component owning the logic. The thing is the content of this component is not as easily accessible because of the new `Provider` as first child.

packages/react-instantsearch-dom-maps/src/__tests__/withGoogleMaps.tsx
```diff
@@ -1,14 +1,11 @@
...
-import withGoogleMaps, {
-  GoogleMapsContext,
-  WithGoogleMapsProps,
-} from '../withGoogleMaps';
+import withGoogleMaps, { WithGoogleMapsProps, } from '../withGoogleMaps';
 
 Enzyme.configure({ adapter: new Adapter() });
 
-describe('withGoogleMaps', () => {
+describe.skip('withGoogleMaps', () => {
```
francoischalifour: To be tested correctly, this will probably also need an inner component to which we pass the context as a prop.

packages/react-instantsearch-dom-maps/src/GeoSearchContext.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type GeoSearchContextProps = {
+  isRefineOnMapMove: boolean;
+  hasMapMoveSinceLastRefine: boolean;
+  toggleRefineOnMapMove: () => void;
+  setMapMoveSinceLastRefine: () => void;
```
samouss: The function takes a boolean as argument: `(value: boolean) => void`

packages/react-instantsearch-dom-maps/src/GeoSearchContext.ts
```diff
@@ -0,0 +1,13 @@
...
+  refineWithInstance: () => void;
```
samouss: The function takes the map instance as argument: `(value: google.maps.Map) => void`

packages/react-instantsearch-dom-maps/src/GeoSearchContext.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+const GeoSearchContext = React.createContext<Partial<GeoSearchContextProps>>({});
```
samouss: We should avoid the partial otherwise we'll have to use the `!` operator when we use the context. We can provide default value.

packages/react-instantsearch-dom-maps/src/Control.js
```diff
@@ -1,65 +1,58 @@
...
+    refineWithInstance: PropTypes.func.isRequired,
+  }),
+};
 
-  getStateContext() {
-    return this.context[STATE_CONTEXT];
-  }
+const ControlWrapper = props => (
+  <GeoSearchContext.Consumer>
+    {context => <Control {...props} context={context} />}
```
samouss: We don't have to provide the prop `context` it's an implementation detail. We can pick and provide which value is really useful for this component.

packages/react-instantsearch-dom-maps/src/Control.js
```diff
@@ -1,65 +1,58 @@
...
+    hasMapMoveSinceLastRefine: PropTypes.bool.isRequired,
+    refineWithInstance: PropTypes.func.isRequired,
+  }),
+};
 
-  getStateContext() {
-    return this.context[STATE_CONTEXT];
-  }
+const ControlWrapper = props => (
+  <GeoSearchContext.Consumer>
```
samouss: Not a fan of this approach but I guess we don't have the choice.

packages/react-instantsearch-dom-maps/src/Control.js
```diff
@@ -1,65 +1,58 @@
...
-            onClick={() => refineWithInstance(googleMapsInstance)}
-          >
-            {translate('redo')}
-          </button>
-        )}
-      </div>
-    );
-  }
-}
+ControlWrapper.displayName = 'Control';
```
samouss: I don't think it's valuable to rename it, we have two times the same component in the tree.

screenshot 2019-02-14 at 10 10 48

packages/react-instantsearch-dom-maps/src/GoogleMaps.js
```diff
@@ -145,16 +141,17 @@ class GoogleMaps extends Component {
...
+    const { google, children } = this.props;
     const { isMapReady } = this.state;
 
     return (
-      <div className={cx('')}>
-        <div ref={c => (this.element = c)} className={cx('map')} />
-        {isMapReady && children}
-      </div>
+      <GoogleMapsContext.Provider value={{ google, instance: this.instance }}>
+        <GoogleMaps isMapReady={isMapReady} onRef={ref => (this.element = ref)}>
```
samouss: We can inline this component, the logic inside it is straightforward.

samouss: We can use [`createRef`](https://reactjs.org/docs/react-api.html#reactcreateref) rather than the callback.

packages/react-instantsearch-dom-maps/src/GoogleMaps.js
```diff
@@ -145,16 +141,17 @@ class GoogleMaps extends Component {
...
-    const { children } = this.props;
+    const { google, children } = this.props;
     const { isMapReady } = this.state;
 
     return (
-      <div className={cx('')}>
-        <div ref={c => (this.element = c)} className={cx('map')} />
-        {isMapReady && children}
-      </div>
+      <GoogleMapsContext.Provider value={{ google, instance: this.instance }}>
```
samouss: We should not use inline value here, it will re-render the `Consumer` on each render. We have [to always provide the same reference](https://reactjs.org/docs/context.html#caveats) unless the state changes.

packages/react-instantsearch-dom-maps/src/GeoSearchContext.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type GeoSearchContextProps = {
```
samouss: It's not really a props it's a state that is used as props.

packages/react-instantsearch-dom-maps/src/GoogleMapsContext.ts
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export type GoogleMapsContextProps = {
+  google: typeof google;
+  instance: google.maps.Map;
+}
+
+const GoogleMapsContext = React.createContext<Partial<GoogleMapsContextProps>>({});
```
samouss: We should avoid the partial otherwise we'll have to use the `!` operator when we use the context. We can provide default value.

packages/react-instantsearch-dom-maps/src/Provider.js
```diff
@@ -121,13 +99,25 @@ class Provider extends Component {
...
-    return children({
-      onChange: this.onChange,
-      onIdle: this.onIdle,
-      shouldUpdate: this.shouldUpdate,
-      boundingBox,
-      boundingBoxPadding,
-    });
+    return (
+      <GeoSearchContext.Provider
+        value={{
```
samouss: We should not use inline value here, it will re-render the `Consumer` on each render. We have [to always provide the same reference](https://reactjs.org/docs/context.html#caveats) unless the state changes.

packages/react-instantsearch-dom-maps/src/Redo.js
```diff
@@ -1,49 +1,47 @@
...
+    refineWithInstance: PropTypes.func.isRequired,
+  }),
+};
 
-  getStateContext() {
-    return this.context[STATE_CONTEXT];
-  }
+const RedoWrapper = props => (
+  <GeoSearchContext.Consumer>
+    {context => <Redo {...props} context={context} />}
```
samouss: We don't have to provide the prop `context` it's an implementation detail. We can pick and provide which value is really useful for this component.

packages/react-instantsearch-dom-maps/src/Redo.js
```diff
@@ -1,49 +1,47 @@
...
-          disabled={!hasMapMoveSinceLastRefine}
-          onClick={() => refineWithInstance(googleMapsInstance)}
-        >
-          {translate('redo')}
-        </button>
-      </div>
-    );
-  }
-}
+RedoWrapper.displayName = 'Redo';
```
samouss: I don't think it's valuable to rename it, we have two times the same component in the tree.

packages/react-instantsearch-dom-maps/src/GoogleMapsContext.ts
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export type GoogleMapsContextProps = {
```
samouss: It's not really a props it's a state that is used as props.

packages/react-instantsearch-dom-maps/src/__tests__/Control.js
```diff
@@ -42,16 +40,15 @@ describe('Control', () => {
...
-      ...defaultProps,
-    };
-
     const context = {
       ...defaultContext,
-      [STATE_CONTEXT]: {
-        ...getStateContext(defaultContext),
-        isRefineOnMapMove: false,
-      },
+      ...getStateContext(defaultContext),
```
samouss: I think this call is useless we already do a copy the line above.

packages/react-instantsearch-dom-maps/src/__tests__/Control.js
```diff
@@ -63,17 +60,16 @@ describe('Control', () => {
...
-    };
-
     const context = {
       ...defaultContext,
-      [STATE_CONTEXT]: {
-        ...getStateContext(defaultContext),
-        isRefineOnMapMove: false,
-        hasMapMoveSinceLastRefine: true,
-      },
+      ...getStateContext(defaultContext),
```
samouss: I think this call is useless we already do a copy the line above.

packages/react-instantsearch-dom-maps/src/__tests__/Control.js
```diff
@@ -84,16 +80,15 @@ describe('Control', () => {
...
-      ...defaultProps,
-    };
-
     const context = {
       ...defaultContext,
-      [STATE_CONTEXT]: {
-        ...getStateContext(defaultContext),
-        toggleRefineOnMapMove: jest.fn(),
-      },
+      ...getStateContext(defaultContext),
```
samouss: I think this call is useless we already do a copy the line above.

packages/react-instantsearch-dom-maps/src/__tests__/Control.js
```diff
@@ -114,19 +109,18 @@ describe('Control', () => {
   it('expect to call refineWithInstance on button click', () => {
     const mapInstance = createFakeMapInstance();
 
+    const context = {
+      ...defaultContext,
+      ...getStateContext(defaultContext),
```
samouss: I think this call is useless we already do a copy the line above.

packages/react-instantsearch-dom-maps/src/__tests__/Redo.js
```diff
@@ -63,24 +55,21 @@ describe('Redo', () => {
...
-      googleMapsInstance: mapInstance,
-    };
-
     const context = {
       ...defaultContext,
-      [STATE_CONTEXT]: {
-        ...getStateContext(defaultContext),
-        refineWithInstance: jest.fn(),
-      },
+      ...getStateContext(defaultContext),
```
samouss: I think this call is useless we already do a copy the line above.

packages/react-instantsearch-dom-maps/src/__tests__/CustomMarker.js
samouss:
```
<GoogleMapsContext.Provider value={context}>
  <Connected {...props}>
    <span>This is the children.</span>
  </Connected>
</GoogleMapsContext.Provider>
```

packages/react-instantsearch-dom-maps/src/__tests__/Marker.js
samouss: https://github.com/algolia/react-instantsearch/pull/2023#discussion_r256762319

packages/react-instantsearch-dom-maps/src/__tests__/withGoogleMaps.tsx
samouss: https://github.com/algolia/react-instantsearch/pull/2023#discussion_r256762319

packages/react-instantsearch-dom-maps/src/GoogleMaps.js
francoischalifour: I separated the component in two because I don't think it's be possible to test it otherwise.

francoischalifour:
@samouss Sorry about the amount of changes, will keep that in mind. It was my first time touching this codebase and using the Context API. Needless to say I was not comfortable.

francoischalifour:
I applied the changes.

The remaining part is to convert the `GoogleMaps` tests.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -306,7 +300,7 @@ describe('GoogleMaps', () => {
     });
   });
 
-  describe('context', () => {
+  describe.skip('context', () => {
```
francoischalifour: Not sure how to get the context from here.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -388,8 +382,8 @@ describe('GoogleMaps', () => {
 
       expect(mapInstance.fitBounds).toHaveBeenCalledTimes(0);
 
-      expect(mapInstance.setZoom).toHaveBeenCalledTimes(1); // cDM
-      expect(mapInstance.setCenter).toHaveBeenCalledTimes(1); // cDM
+      expect(mapInstance.setZoom).toHaveBeenCalledTimes(2); // cDM
```
francoischalifour: Since we store the context in the state, the component renders twice now.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -522,13 +516,13 @@ describe('GoogleMaps', () => {
       simulateMapReadyEvent(google);
       simulateEvent(mapInstance, 'center_changed');
 
-      expect(wrapper).toMatchSnapshot();
+      expect(wrapper.html()).toMatchSnapshot();
```
francoischalifour: That's the easiest solution I found not to change all the tests.
Otherwise, we can't see the generated tree because of the first component being the `Provider`.

packages/react-instantsearch-dom-maps/src/__tests__/Control.js
```diff
@@ -117,23 +72,14 @@ describe('Control', () => {
...
-      },
-    };
-
-    const wrapper = shallow(<Control {...props} />, {
-      context,
-    });
+    const wrapper = shallow(<Control {...props} />);
 
-    const { refineWithInstance } = getStateContext(context);
+    const { refineWithInstance } = props;
```
samouss: We can inline the call inside `expect`.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -47,10 +47,8 @@ describe('GoogleMaps', () => {
       }
     );
 
-    expect(wrapper).toMatchSnapshot();
-    expect(wrapper.state()).toEqual({
-      isMapReady: false,
-    });
+    expect(wrapper.html()).toMatchSnapshot();
```
samouss: The `.html` might be useless if we render the `Provider` at the right time.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
samouss: The number of calls should not change if we render the `Provider` at the right time.

packages/react-instantsearch-dom-maps/src/__tests__/Provider.js
```diff
@@ -388,11 +388,7 @@ describe('Provider', () => {
 
       const wrapper = shallow(<Provider {...props}>{x => x}</Provider>);
 
-      expect(wrapper.instance().getChildContext()).toEqual({
-        [STATE_CONTEXT]: expect.objectContaining({
-          isRefineOnMapMove: true,
-        }),
-      });
+      expect(wrapper.props().value.isRefineOnMapMove).toBe(true);
```
samouss: We can use a `Consumer` to access the value rather than a direct access on the `props`.

packages/react-instantsearch-dom-maps/src/__tests__/Redo.js
```diff
@@ -66,21 +44,12 @@ describe('Redo', () => {
...
-      },
-    };
-
-    const wrapper = shallow(<Redo {...props} />, {
-      context,
-    });
+    const wrapper = shallow(<Redo {...props} />);
 
-    const { refineWithInstance } = getStateContext(context);
+    const { refineWithInstance } = props;
```
samouss: We can inline the call inside `expect`.

packages/react-instantsearch-dom-maps/src/GoogleMapsContext.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export type GoogleMapsContextState = {
+  google: typeof google | null;
```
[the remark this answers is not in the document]
francoischalifour: Can you elaborate?

Mounting the `Provider` only when the map is ready will change the component's lifecycle.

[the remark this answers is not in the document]
samouss: Right now we mount the `Provider` even when the map is not ready. We don't have to because the children that consume the value are only mounted once the map is ready. We probably can do the same with the `Provider`.

https://github.com/algolia/react-instantsearch/blob/d4c71892efb9d1dbd86544d7bf5c22e6fe0548d8/packages/react-instantsearch-dom-maps/src/GoogleMaps.js#L138-L143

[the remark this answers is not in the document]
francoischalifour: Right now we rely on `componentDidMount` to set the Google Maps instance based on the map ref:

https://github.com/algolia/react-instantsearch/blob/d4c71892efb9d1dbd86544d7bf5c22e6fe0548d8/packages/react-instantsearch-dom-maps/src/GoogleMaps.js#L35

Mounting `Provider` only when the map is ready means that we don't have access to this ref anymore.

In order for this to work, we'd need to change the whole component's lifecycle. Is that something that we want to address in this change?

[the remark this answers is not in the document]
samouss: I'm not sure to understand the argument of the `ref`. What prevent us to do something like this?

```jsx
<div className={cx("")}>
  <div ref={this.mapRef} className={cx("map")} />
  {isMapReady && (
    <GoogleMapsContext.Provider value={this.state.value}>
      {children}
    </GoogleMapsContext.Provider>
  )}
</div>
```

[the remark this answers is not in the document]
francoischalifour: Ah, now I understand.

packages/react-instantsearch-dom-maps/src/Provider.js
```diff
@@ -19,38 +18,28 @@ class Provider extends Component {
...
-  }
+    const bounds = instance.getBounds();
+
+    refine({
+      northEast: bounds.getNorthEast().toJSON(),
+      southWest: bounds.getSouthWest().toJSON(),
+    });
+  };
+
+  mapValue = {
```
samouss: I said something that was wrong on the previous review. The value `isRefineOnMapMove` and `hasMapMoveSinceLastRefine` do change across render. The rest is not though we can keep the same reference. We have to update their values otherwise since we always re-use the same object the `Consumer`s don't re-render. We do have the issue on the stories: `Control` and `Redo` don't work anymore.

francoischalifour:
This branch needs to be rebased against a `v6` branch before getting merged.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -41,16 +42,11 @@ describe('GoogleMaps', () => {
...
-        disableLifecycleMethods: true,
-      }
+      </GoogleMaps>
     );
 
     expect(wrapper).toMatchSnapshot();
-    expect(wrapper.state()).toEqual({
-      isMapReady: false,
-    });
+    expect(wrapper.state().isMapReady).toBe(false);
```
samouss: We can drop this assertion and test that the children are not there.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -63,17 +59,12 @@ describe('GoogleMaps', () => {
 
     const wrapper = shallow(
       <GoogleMaps {...props}>
-        <div>This is the children</div>
-      </GoogleMaps>,
-      {
-        disableLifecycleMethods: true,
-      }
+        <div className="children">This is the children</div>
```
samouss: We could use a `testId` rather than `className` in this case.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -63,17 +59,12 @@ describe('GoogleMaps', () => {
...
     );
 
     expect(wrapper).toMatchSnapshot();
-    expect(wrapper.state()).toEqual({
-      isMapReady: false,
-    });
+    expect(wrapper.state().isMapReady).toBe(false);
 
     // Simulate didMount
     wrapper.instance().componentDidMount();
```
samouss: This call is useless now since we don't disable the lifecycle methods anymore.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -63,17 +59,12 @@ describe('GoogleMaps', () => {
...
-      }
+        <div className="children">This is the children</div>
+      </GoogleMaps>
     );
 
     expect(wrapper).toMatchSnapshot();
-    expect(wrapper.state()).toEqual({
-      isMapReady: false,
-    });
+    expect(wrapper.state().isMapReady).toBe(false);
```
samouss: We can drop this assertion and test that the children are not there.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -83,10 +74,8 @@ describe('GoogleMaps', () => {
     // Trigger the update
     wrapper.update();
 
-    expect(wrapper).toMatchSnapshot();
-    expect(wrapper.state()).toEqual({
-      isMapReady: true,
-    });
+    expect(wrapper.find('.children')).toBeTruthy();
+    expect(wrapper.state().isMapReady).toBe(true);
```
samouss: We can drop this assertion and test that the children are there.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -83,10 +74,8 @@ describe('GoogleMaps', () => {
     // Trigger the update
     wrapper.update();
 
-    expect(wrapper).toMatchSnapshot();
-    expect(wrapper.state()).toEqual({
-      isMapReady: true,
-    });
+    expect(wrapper.find('.children')).toBeTruthy();
```
samouss: We can rely on a snapshot like we do for the first assertion.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +296,55 @@ describe('GoogleMaps', () => {
   });
 
   describe('context', () => {
-    it('expect to expose the google object through context', () => {
-      const google = createFakeGoogleReference();
+    it('expect not to expose the context only when map is not ready', () => {
```
samouss:
```suggestion
    it('expect to not expose the context when map is not ready', () => {
```

packages/react-instantsearch-dom-maps/src/GoogleMapsContext.ts
```diff
@@ -0,0 +1,13 @@
...
+  instance: google.maps.Map | null;
```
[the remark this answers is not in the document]
francoischalifour: > Worst case we cast the default values.

What do you mean by this?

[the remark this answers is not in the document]
samouss:
```ts
const GoogleMapsContext = React.createContext<GoogleMapsContextState>({
  google: {} as typeof google,
  instance: {} as google.maps.Map,
});
```

francoischalifour:
@samouss The `Provider` `mapValue` reference tests (https://github.com/algolia/react-instantsearch/pull/2023#discussion_r261580234) were fixed in 523aed341dcf1b192d7ffdc00af776018a610ab0.

packages/react-instantsearch-dom-maps/src/GoogleMapsContext.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export type GoogleMapsContextState = {
+  google: typeof google;
+  instance: google.maps.Map;
+};
+
+const GoogleMapsContext = React.createContext<GoogleMapsContextState>({
+  google: {} as typeof google,
```
samouss:
```suggestion
  // We mount the context only once the map is created that's why we can assume
  // that the value provided through the context is never `undefined`. We can't
  // create an instance at that point, hence the cast.
  google: {} as typeof google,
```

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +290,55 @@ describe('GoogleMaps', () => {
   });
 
   describe('context', () => {
-    it('expect to expose the google object through context', () => {
-      const google = createFakeGoogleReference();
+    it('expect to not expose the context when map is not ready', () => {
+      expect.assertions(1);
```
samouss: We can remove this.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +290,55 @@ describe('GoogleMaps', () => {
...
+
+      const mapInstance = createFakeMapInstance();
```
samouss: `mapInstance` is useless in this test.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +290,55 @@ describe('GoogleMaps', () => {
   });
 
   describe('context', () => {
-    it('expect to expose the google object through context', () => {
-      const google = createFakeGoogleReference();
+    it('expect to not expose the context when map is not ready', () => {
```
samouss:
```suggestion
    it('expect to not expose the context when the map is not ready', () => {
```

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +290,55 @@ describe('GoogleMaps', () => {
...
 
-      expect(wrapper.instance().getChildContext()).toEqual({
-        // eslint-disable-next-line camelcase
-        __ais_geo_search__google_maps__: expect.objectContaining({
-          google,
-        }),
+      expect(wrapper.find(GoogleMapsContext.Consumer).exists()).toBe(false);
+    });
+
+    it('expect to expose the context only when created', () => {
```
samouss:
```suggestion
    it('expect to expose the context only when the map is created', () => {
```

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +290,55 @@ describe('GoogleMaps', () => {
...
+      expect.assertions(1);
```
samouss: We can remove this.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +290,55 @@ describe('GoogleMaps', () => {
...
+
+      const mapInstance = createFakeMapInstance();
```
samouss: `mapInstance` is useless in this test.

packages/react-instantsearch-dom-maps/src/__tests__/GoogleMaps.js
```diff
@@ -307,27 +290,55 @@ describe('GoogleMaps', () => {
...
+        </GoogleMaps>
+      );
+
+      simulateMapReadyEvent(google);
+
+      expect(wrapper.find(GoogleMapsContext.Consumer).exists()).toBe(true);
     });
 
     it('expect to expose the map instance through context only when created', () => {
+      expect.assertions(1);
```
samouss: We can remove this.

packages/react-instantsearch-dom-maps/src/__tests__/Provider.js
```diff
@@ -444,11 +432,99 @@ describe('Provider', () => {
...
+    });
+
+    it('expect to pass new mapValue reference when isRefineOnMapMove changes', () => {
+      const props = {
+        ...defaultProps,
+        isRefineOnMapMove: true,
+      };
+
+      const wrapper = shallow(<Provider {...props} />);
+      const mapValue = wrapper.instance().mapValue;
```
samouss: We heavily rely on implementation details here, would have been better to use a `Consumer` to test those cases. We also miss some cases e.g. when `isRefineOnMapMove` is equal. We can improve those tests later.

PR #2025, opened 2019-02-14
chore(website): use Webpack rather than CRA [PART-11]

samouss:
**Summary**

This PR drops CRA for Webpack to build the examples. Apologise for the huge PR but a lot of the changes are only files that have been removed. Here is the list of important changes:

- drop CRA for Webpack for each examples
- fix an issue with the URLSync of each examples (https://github.com/algolia/react-instantsearch/commit/dae64e8a8de02fe29f00377269335a1a739b6e40)
- Wepback configuration for the examples (https://github.com/algolia/react-instantsearch/commit/d5d99375b24b90530d5e52cda71ac4126893398d)

**Result**

The previous Netlify build was about 10 ~ 11 minutes we are now back to 6 ~ 7 minutes. We have a last change to make to boost the build but we have to wait for the next release of Storybook. The `quiet` option is not available for the production build. The logs slow down the build.


website/examples/default-theme/README.md
```diff
@@ -1,21 +0,0 @@
-# Algolia theme
```
Haroenv: This exact readme isn't necessary, but folders show up a lot nicer on GitHub if they have a readme.

samouss: The thing is that you cannot run the examples separately. What do you want to put in the README then?

Haroenv: A description of what this project is; similar to the little blurb on the website maybe? Since you can't run them individually, that's of course a bit of a problem. I wonder if we can get to a position where they can be run individually; but are bundled for the website at once

samouss: Yes it's possible, it's Webpack so we can do whatever we want. We can add it when we have the needs.

PR #2179, opened 2019-03-18
feat(context): migrate base connector [STEP 2]

Haroenv:
This is enough changes to be able to look at the storybook without errors again (only pagination, hits, searchbox of course, and nothing multi-index)

For full roadmap, see #2178

[→ stories](
https://deploy-preview-2179--react-instantsearch.netlify.com/storybook/?selectedKind=Configure&selectedStory=default&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs)

PR #2180, opened 2019-03-18
feat(context): migrate remaining connectors [STEP 3]

Haroenv:
This is like #2179, but for the remaining connectors

Here again, all multi-index tests are skipped

[→ stories](
https://deploy-preview-2180--react-instantsearch.netlify.com/storybook)

EDIT: looks like some connectors didn't migrate correctly, looking into it


packages/react-instantsearch-core/src/connectors/__tests__/connectHitsPerPage.js
```diff
@@ -56,7 +57,7 @@ describe('connectHitsPerPage', () => {
     });
 
     it("calling refine updates the widget's search state", () => {
-      const nextState = refine({}, { otherKey: 'val' }, 30);
+      const nextState = connect.refine({}, { otherKey: 'val' }, 30);
```
samouss: We don't provide the `contextValue` but we use it inside the `refine` function.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -5,39 +5,35 @@ jest.mock('../../core/createConnector', () => x => x);
 describe('connectInfiniteHits', () => {
   describe('single index', () => {
     const createSingleIndexContext = () => ({
-      context: {
-        ais: {
-          mainTargetedIndex: 'index',
-        },
-      },
+      mainTargetedIndex: 'index',
```
samouss: We removed the function in the other tests we should stay consistent, either we keep it or we remove it.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -5,39 +5,35 @@ jest.mock('../../core/createConnector', () => x => x);
...
 
     it('accumulate hits internally', () => {
-      const context = createSingleIndexContext();
-      const getProvidedProps = connect.getProvidedProps.bind(context);
+      const contextValue = createSingleIndexContext();
 
       const hits = [{}, {}];
       const hits2 = [{}, {}];
-      const res1 = getProvidedProps(null, null, {
+      const ctx = {};
```
samouss: `ctx` is misleading since we speak about `context` inside the props. `instance` should be ok since it's actually a fake component instance. We have to apply this change in the all file.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -51,21 +47,21 @@ describe('connectInfiniteHits', () => {
     });
 
     it('accumulate hits internally while changing hitsPerPage configuration', () => {
-      const context = createSingleIndexContext();
-      const getProvidedProps = connect.getProvidedProps.bind(context);
```
samouss: We can keep this line, at least we bind it once for the test case.

packages/react-instantsearch-core/src/connectors/__tests__/connectRefinementList.js
```diff
@@ -388,16 +412,25 @@ describe('connectRefinementList', () => {
...
     it('calling searchForItems return the right searchForItems parameters with limit', () => {
-      const parameters = searchForFacetValues(
-        { attribute: 'ok', limit: 15, showMoreLimit: 25, showMore: false },
+      const parameters = connect.searchForFacetValues(
+        {
+          attribute: 'ok',
+          limit: 15,
+          showMoreLimit: 25,
+          showMore: false,
+          contextValue,
```
samouss: It's useless in that case, we don't use the context value inside the `searchForFacetValues`.

packages/react-instantsearch-core/src/connectors/__tests__/connectRefinementList.js
```diff
@@ -410,8 +443,14 @@ describe('connectRefinementList', () => {
...
     it('calling searchForItems return the right searchForItems parameters with showMoreLimit', () => {
-      const parameters = searchForFacetValues(
-        { attribute: 'ok', limit: 15, showMoreLimit: 25, showMore: true },
+      const parameters = connect.searchForFacetValues(
+        {
+          attribute: 'ok',
+          limit: 15,
+          showMoreLimit: 25,
+          showMore: true,
+          contextValue,
```
samouss: It's useless in that case, we don't use the context value inside the `searchForFacetValues`.

packages/react-instantsearch-core/src/connectors/__tests__/connectScrollTo.js
```diff
@@ -5,32 +5,35 @@ jest.mock('../../core/createConnector', () => x => x);
 let props;
 describe('connectScrollTo', () => {
   describe('single index', () => {
-    const context = { context: { ais: { mainTargetedIndex: 'index' } } };
-    const getProvidedProps = connect.getProvidedProps.bind(context);
+    const contextValue = { mainTargetedIndex: 'index' };
     it('provides the correct props to the component', () => {
-      props = getProvidedProps(
-        { scrollOn: 'p' },
+      props = connect.getProvidedProps(
```
samouss: This component relies on `this` we have to create a fake instance.

packages/react-instantsearch-core/src/connectors/__tests__/connectSortBy.js
```diff
@@ -8,15 +8,12 @@ let params;
...
-    const refine = connect.refine.bind(context);
-    const getSP = connect.getSearchParameters.bind(context);
-    const getMetadata = connect.getMetadata.bind(context);
-    const cleanUp = connect.cleanUp.bind(context);
+    const contextValue = { mainTargetedIndex: 'index' };
+
     it('provides the correct props to the component', () => {
-      props = getProvidedProps(
-        { items: [{ value: 'yep' }, { value: 'yop' }] },
+      props = connect.getProvidedProps.call(
```
samouss: We don't have to use `.call` we don't rely on `this` inside the connector. We have to apply this change in the all file.

packages/react-instantsearch-core/src/connectors/__tests__/connectSortBy.js
```diff
@@ -58,18 +66,26 @@ describe('connectSortBy', () => {
...
+        new SearchParameters(),
+        { contextValue },
+        { sortBy: 'yep' }
+      );
       expect(params.index).toBe('yep');
     });
 
     it('registers its id in metadata', () => {
-      const metadata = getMetadata({});
+      const metadata = connect.getMetadata.call({}, { contextValue });
```
samouss: It's useless in that case, we don't use the context value inside the `getMetadata`.

packages/react-instantsearch-core/src/connectors/connectMenu.js
```diff
@@ -112,21 +112,20 @@ export default createConnector({
     );
 
     // Search For Facet Values is not available with derived helper (used for multi index search)
-    if (searchable && this.context.multiIndexContext) {
-      throw new Error(
-        'react-instantsearch: searching in *List is not available when used inside a' +
-          ' multi index context'
-      );
-    }
+    // @TODO: re-enable this error
```
samouss: Can we already re-enable it?

packages/react-instantsearch-core/src/connectors/connectRefinementList.js
```diff
@@ -134,22 +134,21 @@ export default createConnector({
...
 
+    // @TODO: re-enable this error based on index context
     // Search For Facet Values is not available with derived helper (used for multi index search)
-    if (searchable && this.context.multiIndexContext) {
-      throw new Error(
-        'react-instantsearch: searching in *List is not available when used inside a' +
-          ' multi index context'
-      );
-    }
+    // if (searchable && this.context.multiIndexContext) {
```
samouss: Can we already re-enable it?

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
Haroenv: that might be inconsistent, since we no longer bind in the other tests

packages/react-instantsearch-core/src/connectors/__tests__/connectSortBy.js
Haroenv: I just realised I confused SortBy & ScrollTo

packages/react-instantsearch-core/src/connectors/connectMenu.js
Haroenv: The error only makes sense once multi-index is enabled, so I did in #2185

packages/react-instantsearch-core/src/connectors/connectRefinementList.js
Haroenv: see #2185

packages/react-instantsearch-core/src/connectors/connectMenu.js
samouss: Yep so it with my review of the PR.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
samouss: Yes but you can do it for the full test suite.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
Haroenv: the reason I kept it here is because the object gets modified in these tests, but not in others

packages/react-instantsearch-core/src/connectors/__tests__/connectQueryRules.ts
```diff
@@ -20,29 +23,31 @@ describe('connectQueryRules', () => {
...
-    const getProvidedProps = connect.getProvidedProps.bind(context);
-    const getSearchParameters = connect.getSearchParameters.bind(context);
+    const contextValue: any = { mainTargetedIndex: indexName };
 
     describe('default', () => {
       it('without userData provides the correct props to the component', () => {
-        const props: QueryRulesProps = {
+        const props: ConnectedProps<QueryRulesProps> = {
           ...defaultProps,
+          contextValue,
```
samouss: We can move the value inside the `defaultProps` that would avoid to repeat it N times in the file.

PR #2181, opened 2019-03-18
feat(context): multi-index [STEP 4]

Haroenv:
- [x] expose new context from Index
- [x] read new context from createConnector
- [x] test new meaning of Index widget


packages/react-instantsearch-core/src/components/Index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  unregisterWidget?: () => void;
+
+  state = {
+    indexContext: {
+      targetedIndex: this.props.indexId,
+    },
+  };
+
+  componentWillMount() {
+    this.unregisterWidget = this.props.contextValue.widgetsManager.registerWidget(
```
Haroenv: I wonder if this behaviour is the same or not

packages/react-instantsearch-core/src/components/Index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+
+const IndexWrapper = (props: Props) => (
+  <InstantSearchConsumer>
+    {contextValue => <Index contextValue={contextValue} {...props} />}
+  </InstantSearchConsumer>
+);
+
+export const IndexComponentWithoutContext = Index;
```
Haroenv: This is similar to the split made in createConnector

packages/react-instantsearch-core/src/components/__tests__/Index.js
```diff
@@ -104,30 +90,26 @@ describe.skip('Index', () => {
...
-      <Index {...requiredProps}>
-        <div />
-      </Index>,
-      {
-        context,
-      }
+    const wrapper = mount(
+      <IndexComponentWithoutContext {...requiredProps} contextValue={context}>
+        <IndexConsumer>
+          {childContext => (
```
samouss:
```suggestion
          {multiIndexContext => (
```

packages/react-instantsearch-core/src/components/__tests__/Index.js
```diff
@@ -157,4 +136,72 @@ describe.skip('Index', () => {
...
+      getProvidedProps: props => ({ providedProps: props }),
+    })(Dummy);
+
+    const props = {
+      message: 'hello',
+    };
+
+    const wrapper = mount(
+      <InstantSearchProvider value={context}>
+        <Index root={{ Root: 'div' }} indexId="test" indexName="test">
```
samouss: We can use the `requiredProps` defined above. Here there is a confusion between `indexName` & `indexId` since both values are the same. The snapshots display the value but we don't know which one is it. We don't test this particular behaviour in the test but it brings a bit of confusion.

packages/react-instantsearch-core/src/connectors/connectConfigure.js
```diff
@@ -16,12 +16,12 @@ export default createConnector({
     return {};
   },
   getSearchParameters(searchParameters, props) {
-    const items = omit(props, 'children', 'contextValue');
+    const items = omit(props, 'children', 'contextValue', 'indexContextValue');
```
samouss: We have to add a test for this behaviour.

packages/react-instantsearch-core/src/connectors/connectConfigure.js
```diff
@@ -16,12 +16,12 @@ export default createConnector({
...
     return searchParameters.setQueryParameters(items);
   },
   transitionState(props, prevSearchState, nextSearchState) {
     const id = getId();
-    const items = omit(props, 'children', 'contextValue');
+    const items = omit(props, 'children', 'contextValue', 'indexContextValue');
```
samouss: We have to add a test for this behaviour.

packages/react-instantsearch-core/src/core/createConnector.tsx
```diff
@@ -346,7 +355,7 @@ export function createConnectorWithoutContext(
       };
 
       render() {
-        const { contextValue, ...props } = this.props;
+        const { contextValue, indexContextValue, ...props } = this.props;
```
samouss: We'll probably have to rename `contextValue` to avoid name collision. Later.

packages/react-instantsearch-core/src/core/createConnector.tsx
```diff
@@ -72,6 +77,15 @@ type ConnectorDescription = {
   defaultProps?: {};
 };
 
+type ConnectorProps = {
+  contextValue: InstantSearchContext;
+  indexContextValue: IndexContext;
```
samouss: This value is optional. When the component is not outside of  the multi-index context we don't have a value.

packages/react-instantsearch-core/src/core/context.ts
```diff
@@ -27,3 +27,14 @@ export const {
...
+export type IndexContext =
+  | {
+      targetedIndex: string;
+    }
+  | undefined;
+
+export const {
+  Consumer: IndexConsumer,
+  Provider: IndexProvider,
+} = createContext<IndexContext>(undefined);
```
samouss: Is it really necessary? `createContext<IndexContext>()` should be equivalent?

Haroenv: the TS definitions require a value there `Expected 1-2 arguments, but got 0. ts(2554)` if I remove the `undefined`

packages/react-instantsearch-core/src/connectors/connectConfigure.js
Haroenv: added a test in 8fa0d684

packages/react-instantsearch-core/src/components/Index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  IndexProvider,
+  IndexContext,
+} from '../core/context';
+
+type Props = {
+  indexName: string;
+  indexId: string;
+  root: {
+    Root: ReactType;
+    props: {}; // I think this is correct
```
samouss:
```suggestion
    props: {};
```

packages/react-instantsearch-core/src/components/Index.tsx
```diff
@@ -0,0 +1,144 @@
...
+      <Root {...props}>
+        <IndexProvider value={this.state.indexContext}>
+          {this.props.children}
+        </IndexProvider>
+      </Root>
+    );
+  }
+}
+
+const IndexWrapper = (props: Props) => (
```
samouss: We should use `React.FC<Props>`.

This is not correct the `IndexWrapper` does not accept a prop `contextValue`.

packages/react-instantsearch-core/src/core/createConnector.tsx
```diff
@@ -379,7 +388,17 @@ const createConnectorWithContext = (connectorDesc: ConnectorDescription) => (
   // @TODO: does this need a display name?
   return props => (
```
samouss: We could create a `ConnectorWrapper` like we did for `Index`.

packages/react-instantsearch-core/src/components/Index.tsx
```diff
@@ -0,0 +1,145 @@
...
+type Props = {
+  indexName: string;
+  indexId: string;
+  root: {
+    Root: ReactType;
+    props: {};
+  };
+};
+
+type InnerProps = Props & { contextValue: InstantSearchContext };
```
samouss:
```suggestion
type IndexProps = Props & { contextValue: InstantSearchContext };
```

PR #2185, opened 2019-03-18
feat(context): apply multi-index [STEP 5]

Haroenv:
All connectors need to have multiIndex added and those tests re-enabled

- [x] connectAutoComplete
- [x] connectBreadcrumb
- [x] connectConfigure
- [x] connectCurrentRefinements
- [x] connectGeoSearch
- [x] connectHierarchicalMenu
- [x] connectHighlight
- [x] connectHits
- [x] connectHitsPerPage
- [x] connectInfiniteHits
- [x] connectMenu
- [x] connectNumericMenu
- [x] connectPagination
- [x] connectPoweredBy
- [x] connectRange
- [x] connectRefinementList
- [x] connectScrollTo
- [x] connectSearchBox
- [x] connectSortBy
- [x] connectStateResults
- [x] connectStats
- [x] connectToggleRefinement


packages/react-instantsearch-core/src/connectors/connectMenu.js
```diff
@@ -99,8 +99,11 @@ export default createConnector({
     meta,
     searchForFacetValuesResults
   ) {
-    const { attribute, searchable } = props;
-    const results = getResults(searchResults, { ais: props.contextValue });
+    const { attribute, searchable, indexContextValue } = props;
+    const results = getResults(searchResults, {
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
```
samouss: Since we destruct the value above we can re-use it. We can apply the change in the all function.

packages/react-instantsearch-core/src/connectors/connectRefinementList.js
```diff
@@ -122,8 +122,11 @@ export default createConnector({
     metadata,
     searchForFacetValuesResults
   ) {
-    const { attribute, searchable } = props;
-    const results = getResults(searchResults, { ais: props.contextValue });
+    const { attribute, searchable, indexContextValue } = props;
+    const results = getResults(searchResults, {
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
```
samouss: Since we destruct the value above we can re-use it. We can apply the change in the all function.

packages/react-instantsearch-core/src/connectors/connectMenu.js
Haroenv: I didn't do that, because grepping the code for this pattern would become harder, since we plan on removing this in favour of a nicer API

packages/react-instantsearch-core/src/connectors/connectConfigure.js
```diff
@@ -16,27 +16,36 @@ export default createConnector({
...
     const nonPresentKeys = this._props
       ? difference(Object.keys(this._props), Object.keys(props))
       : [];
     this._props = props;
     const nextValue = {
       [id]: { ...omit(nextSearchState[id], nonPresentKeys), ...items },
     };
-    return refineValue(nextSearchState, nextValue, { ais: props.contextValue });
+    return refineValue(nextSearchState, nextValue, {
+      ais: props.contextValue,
```
samouss: We can use `contextValue` & `indexContextValue` that have been defined above.

packages/react-instantsearch-core/src/connectors/connectGeoSearch.js
```diff
@@ -128,21 +128,24 @@ export default createConnector({
   displayName: 'AlgoliaGeoSearch',
 
   getProvidedProps(props, searchState, searchResults) {
-    const results = getResults(searchResults, { ais: props.contextValue });
+    const results = getResults(searchResults, {
+      ais: props.contextValue,
```
samouss: We can store the context inside a value rather the create the same object three times.

PR #2192, opened 2019-03-19
feat(context): re-enable SSR [STEP 6]

Haroenv:
- [x] fix onSearchParameters (maybe needs to be done in a separate pr? can rebase later
- [x] re-enable tests on ssr
- [x] fix tests on ssr


Haroenv:
looks like I undid some changes :(

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -103,6 +103,7 @@ describe.skip('createInstantSearchServer', () => {
     });
 
     it('uses the provided algoliaClient', () => {
+      global.console.warn = jest.fn();
```
samouss: We have to restore the original function after the test.

PR #2210, opened 2019-03-21
feat(queryRules): add connectQueryRules connector [part 1]

francoischalifour:
_This is the 1st PR in the series "Merchandized Query Rules"._

## Summary

This creates the `connectQueryRules` connector, responsible for handling Algolia Query Rules features.

## Motivation

[Read RFC →](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/merchandized-query-rules.md)

## Implementation

The connector passes `userData` array to the widget via a prop named `items`.

For this React implementation, I decided not to keep the name `userData` because at second thought, it makes more sense to apply `transformItems` to something called `items`. I will rename this data in the InstantSearch.js implementation.

## Connector usage

Two widgets will be based on this connector:

- `QueryRuleCustomData`
- `QueryRuleContext`

## Related

- InstantSearch.js
  - [`connectQueryRules`](https://github.com/algolia/instantsearch.js/pull/3597)
  - [`queryRuleCustomData`](https://github.com/algolia/instantsearch.js/pull/3600)
  - [`queryRuleContext`](https://github.com/algolia/instantsearch.js/pull/3602)
- [Merchandized Query Rules RFC](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/merchandized-query-rules.md)

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectQueryRules.ts
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectQueryRules.ts
@@ -1,8 +1,6 @@
-import connect from '../connectQueryRules';
+import connect, { QueryRulesProps } from '../connectQueryRules';
 
-jest.mock('../../core/createConnector', () => (component: React.Component) =>
-  component
-);
+jest.mock('../../core/createConnector', () => (connector: any) => connector);
 
 describe('connectQueryRules', () => {
   describe('single index', () => {
@@ -12,7 +10,9 @@ describe('connectQueryRules', () => {
 
     describe('without userData', () => {
       it('provides the correct props to the component', () => {
-        const props = {};
+        const props: QueryRulesProps = {
+          transformItems: items => items,
+        };
         const searchState = {};
         const searchResults = {
           results: { [indexName]: { userData: undefined } },
@@ -27,7 +27,9 @@ describe('connectQueryRules', () => {
 
     describe('with userData', () => {
       it('provides the correct props to the component', () => {
-        const props = {};
+        const props: QueryRulesProps = {
+          transformItems: items => items,
+        };
         const searchState = {};
         const searchResults = {
           results: {
@@ -42,8 +44,11 @@ describe('connectQueryRules', () => {
       });
 
       it('transforms items before passing the props to the component', () => {
-        const props = {
-          transformItems: () => [{ banner: 'image-transformed.png' }],
+        const transformItemsSpy = jest.fn(() => [
+          { banner: 'image-transformed.png' },
+        ]);
+        const props: QueryRulesProps = {
+          transformItems: transformItemsSpy,
         };
         const searchState = {};
         const searchResults = {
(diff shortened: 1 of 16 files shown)


packages/react-instantsearch-core/src/connectors/__tests__/connectQueryRules.ts
```diff
@@ -0,0 +1,122 @@
...
+
+        expect(getProvidedProps(props, searchState, searchResults)).toEqual({
+          items: [{ banner: 'image-transformed.png' }],
+          canRefine: true,
+        });
+      });
+    });
+  });
+
+  describe('multi index', () => {
```
francoischalifour: I'm not sure if the "multi index" test suit makes sense for this connector. This is the same as the "single index" test suit, with a different context.

Haroenv: Still makes sense to test that it's actually calling `getResults` correctly no?

packages/react-instantsearch-core/src/connectors/__tests__/connectQueryRules.ts
```diff
@@ -0,0 +1,122 @@
+import connect from '../connectQueryRules';
+
+jest.mock('../../core/createConnector', () => (component: React.Component) =>
```
samouss: It not a `React.component` but the connector definition.

packages/react-instantsearch-core/src/connectors/__tests__/connectQueryRules.ts
```diff
@@ -0,0 +1,122 @@
...
+
+        expect(getProvidedProps(props, searchState, searchResults)).toEqual({
+          items: [{ banner: 'image.png' }],
+          canRefine: true,
+        });
+      });
+
+      it('transforms items before passing the props to the component', () => {
+        const props = {
+          transformItems: () => [{ banner: 'image-transformed.png' }],
```
samouss: We should also test that the `transformItems` function is called with the `items` as argument.

packages/react-instantsearch-core/src/connectors/__tests__/connectQueryRules.ts
samouss: We can keep it, the multi-index is a feature that does a lot of work behind the hood. But you can test on a different index that the main one, otherwise yes it's a bit useless.

packages/react-instantsearch-core/src/connectors/connectQueryRules.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (results === null) {
+      return {
+        items: [],
+        canRefine: false,
+      };
+    }
+
+    const { userData = [] } = results;
+    const { transformItems } = props;
+    const transformedItems = transformItems
```
samouss: We have a default value, we don't have to make the assertions. You can fake the default props inside the test.

packages/react-instantsearch-core/src/index.js
```diff
@@ -48,3 +48,4 @@ export { default as connectStats } from './connectors/connectStats';
 export {
   default as connectToggleRefinement,
 } from './connectors/connectToggleRefinement';
+export { default as connectQueryRules } from './connectors/connectQueryRules';
```
samouss: I think the export are sorted for the connectors, easier to navigate inside them.

PR #2215, opened 2019-03-22
 feat(insights): compute insights data in connectInfiniteHits [PART-0]

tkrugg:
**Summary**

This PR adds what's need to add **absolute position** and **queryID** to every hit of connectInfiniteHits

**Result**

Not much to demonstrate yet but from now on every hit will look like this
```js
{
    objectID: "theObjectID",
    __position: 42,
    __queryID: "theQueryID"
}
```

**Global Outline** 
- [x] PART-0 add __positions and __query ID inside connectInfiniteHits [#current](https://github.com/algolia/react-instantsearch/pull/2215)
- [x] PART-1 add __positions and __query ID inside connectHits https://github.com/algolia/react-instantsearch/pull/2216
- [x] PART-2 create connectInsights https://github.com/algolia/react-instantsearch/pull/2217


packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -6,6 +6,21 @@ export function getIndexId(context) {
     : context.ais.mainTargetedIndex;
 }
 
+export function addAbsolutePositions(hits, hitsPerPage, page) {
```
samouss: It should be more inside `utils` rather than `indexUtils`. This file mainly deals with `context` and `searchState`.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -53,7 +55,10 @@ export default createConnector({
       };
     }
 
-    const { hits, page, nbPages } = results;
+    const { page, nbPages } = results;
+
+    let hits = addAbsolutePositions(results.hits, results.hitsPerPage, page);
```
samouss: I would rather use two different variables to actually inform that they've changed e.g. `hitsWithPositions`.

packages/react-instantsearch-core/src/core/indexUtils.js
tkrugg: Thanks for advising on that, I initially has a FIXME: find theses functions a better place but linting got to remove it :)
Will do

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
tkrugg: I agree, let do it!

packages/react-instantsearch-core/src/core/utils.js
```diff
@@ -44,3 +44,18 @@ export const removeEmptyKey = obj => {
...
+
+export function addAbsolutePositions(hits, hitsPerPage, page) {
+  return hits.map((hit, index) => ({
+    ...hit,
+    __position: hitsPerPage * page + index + 1,
+  }));
+}
+
+export function addQueryID(hits, queryID) {
+  if (!queryID) return hits;
```
francoischalifour: Can we avoid inline returns?

tkrugg: I know this is a pretty common coding style preference but it doesn't seem we follow it in the codebase: 
eg. https://github.com/algolia/react-instantsearch/blob/ec9e0fbd6106d1c3e47f1dbfa6eaac3e20af6bd5/packages/react-instantsearch-core/src/components/Index.js#L82

If we decide this is a coding style we want to adopt, we should just add it to linting. 
https://eslint.org/docs/rules/curly#require-following-curly-brace-conventions-curly

francoischalifour: It's mentioned in our [ESLint config](https://github.com/algolia/eslint-config-algolia/blob/e6661db323a69d4304a4bc84571e2263e572a6a4/rules/base.js#L45) but for some reason not picked up.

We can still apply this rule for new code.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
samouss: Yes.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -59,10 +64,25 @@ describe('connectInfiniteHits', () => {
...
+          page: 0,
+          hitsPerPage: 6,
+          nbPages: 10,
+          queryID: 'theQueryID_0',
+        },
       });
 
-      expect(res1.hits).toEqual(hits);
+      expect(res1.hits).toEqual(hits.map(hit => expect.objectContaining(hit)));
+      expect(res1.hits.map(hit => hit.__position)).toEqual([1, 2, 3, 4, 5, 6]);
```
samouss: `_highlightResults` contains only one `_` but our `__position` contains two `__`. Is it on purpose?

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -1,4 +1,4 @@
-import { has, omit, get } from 'lodash';
+import { get, has, omit } from 'lodash';
```
samouss: We can revert this change, unless you really care about it.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
tkrugg: I had `_` initially in the RFC and someone mentioned that we only `__` is considered truly private
hence having `__hitIndex` currently, rather than `_hitIndex`.
Ready to change it if necessary (here and in vanilla)

packages/react-instantsearch-core/src/core/indexUtils.js
tkrugg: I don't, I was sure it was prettier at first I can't find a rule that does that. Reverting

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
samouss: You're right it's probably better at least the API won't override it by mistake.

packages/react-instantsearch-core/src/core/utils.js
tkrugg: of course, but it's best to leverage the tooling to move the focus away from style in reviews.

I looked a bit and it looks like it's disabled by prettier which we use in algolia/react
https://github.com/prettier/eslint-config-prettier/blob/master/index.js#L9

We should reenable it if we care about this curly rule

tkrugg: this is fixed btw, but looking forward to continue the discussion on adding this rule to our linter.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -21,7 +21,10 @@ describe('connectInfiniteHits', () => {
         results: { hits, page: 0, hitsPerPage: 2, nbPages: 3 },
       });
 
-      expect(props).toEqual({ hits, hasMore: true });
+      expect(props).toEqual({
+        hits: hits.map(hit => expect.objectContaining(hit)),
```
Haroenv: same comment as for Hits, this could be better something more precise

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -83,10 +138,63 @@ describe('connectInfiniteHits', () => {
...
+        5,
+        6,
+        // page: 1, hitsPerPage: 6
+        7,
+        8,
+        9,
+        10,
+        11,
+        12,
+        // page: 2, hitsPerPage: 8
```
Haroenv: I'd clarify after this that it's an existing limitation of the engine of not being able to correctly compute an offset

tkrugg: yes, I'll add the following
> hitsPerPage changed from 6 to 8, elements 13-16 are skipped

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
tkrugg: I added because now hits get `__position` and `__queryID`.
This tests is strictly about checking hits identity

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
tkrugg: fixed here: https://github.com/algolia/react-instantsearch/pull/2217/commits/45220e54ada5f9734405e7b0d971dc6f238243ee

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
Haroenv: `hits.map(hit => expect.objectContaining(hit))` will always pass, except if hit isn't an object though. I don't see how this asserts anything.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -53,14 +54,24 @@ export default createConnector({
       };
     }
 
-    const { hits, page, nbPages } = results;
+    const { page, nbPages } = results;
+
+    const hitsWithPositions = addAbsolutePositions(
+      results.hits,
+      results.hitsPerPage,
```
samouss: We should stay consistent either we destruct either we use the dot form.

tkrugg: fixed here 52ebc802201bbbbf7a6dbb3c719802bd1347c8ad

PR #2216, opened 2019-03-22
feat(insights): compute insights data in connectHits [PART-1] 

tkrugg:
**Summary**

This PR adds what's need to add **absolute position** and **queryID** to every hit of connectHits

**Result**

Not much to demonstrate yet but from now on every hit will look like this
```js
{
    objectID: "theObjectID",
    __position: 42,
    __queryID: "theQueryID"
}
```

**Global Outline** 
- [x] PART-0 add __positions and __query ID inside connectInfiniteHits https://github.com/algolia/react-instantsearch/pull/2215
- [x] PART-1 add __positions and __query ID inside connectHits [#current](https://github.com/algolia/react-instantsearch/pull/2216)
- [x] PART-2 create connectInsights https://github.com/algolia/react-instantsearch/pull/2217

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectHits.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectHits.js
@@ -11,9 +11,31 @@ describe('connectHits', () => {
     it('provides the current hits to the component', () => {
       const hits = [{}];
       const props = getProvidedProps(null, null, {
-        results: { hits },
+        results: { hits, hitsPerPage: 2, page: 2 },
       });
-      expect(props).toEqual({ hits });
+      expect(props).toEqual({
+        hits: hits.map(hit => expect.objectContaining(hit)),
+      });
+    });
+
+    it('adds positions to the hits provided to the component', () => {
+      const hits = [{}];
+      const props = getProvidedProps(null, null, {
+        results: { hits, hitsPerPage: 2, page: 2 },
+      });
+      expect(props).toEqual({
+        hits: [{ __position: 5 }],
+      });
+    });
+
+    it('adds queryID to the hits provided to the component', () => {
+      const hits = [{}];
+      const props = getProvidedProps(null, null, {
+        results: { hits, hitsPerPage: 2, page: 2, queryID: 'theQueryID' },
+      });
+      expect(props).toEqual({
+        hits: [expect.objectContaining({ __queryID: 'theQueryID' })],
+      });
     });
 
     it("doesn't render when no hits are available", () => {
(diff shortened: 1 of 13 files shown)


packages/react-instantsearch-core/src/connectors/connectHits.js
```diff
@@ -45,8 +46,16 @@ export default createConnector({
 
   getProvidedProps(props, searchState, searchResults) {
     const results = getResults(searchResults, this.context);
-    const hits = results ? results.hits : [];
+    let hits = [];
+    if (results) {
+      hits = addAbsolutePositions(
```
samouss: see https://github.com/algolia/react-instantsearch/pull/2215#discussion_r268181991

tkrugg: fixed!

packages/react-instantsearch-core/src/connectors/__tests__/connectHits.js
```diff
@@ -11,9 +11,31 @@ describe('connectHits', () => {
...
+      });
+    });
+
+    it('adds queryID to the hits provided to the component', () => {
+      const hits = [{}];
+      const props = getProvidedProps(null, null, {
+        results: { hits, hitsPerPage: 2, page: 2, queryID: 'theQueryID' },
+      });
+      expect(props).toEqual({
+        hits: [expect.objectContaining({ __queryID: 'theQueryID' })],
```
Haroenv: this only asserts on the first item of the list

packages/react-instantsearch-core/src/connectors/__tests__/connectHits.js
```diff
@@ -11,9 +11,31 @@ describe('connectHits', () => {
     it('provides the current hits to the component', () => {
       const hits = [{}];
       const props = getProvidedProps(null, null, {
-        results: { hits },
+        results: { hits, hitsPerPage: 2, page: 2 },
+      });
+      expect(props).toEqual({
+        hits: hits.map(hit => expect.objectContaining(hit)),
```
Haroenv: aren't you just asserting that Array#map works here?

packages/react-instantsearch-core/src/connectors/connectHits.js
```diff
@@ -45,9 +46,19 @@ export default createConnector({
...
-    return { hits };
+    if (!results) {
+      return { hits: [] };
+    }
+    const hitsWithPositions = addAbsolutePositions(
+      results.hits,
+      results.hitsPerPage,
+      results.page
+    );
+    const hitsWithPositionsAndQueryID = addQueryID(
```
Haroenv: these names don't seem very useful to me, what about composing it: 

```js
  return {
    hits: addQueryId(
      addAbsolutePositions(
        results.hits,
        results.hitsPerPage,
        results.page
      ),
      results.queryID
    )
  };
```

packages/react-instantsearch-core/src/connectors/__tests__/connectHits.js
tkrugg: 😆 You made me double check.
So apparently it works as expected
https://repl.it/repls/SupportiveEveryFirmware

To summarize this checks that `props.hits` and `hits`: 
- are the same length
- contains the same objectIDs as hits, in the same order
```
props.hits = [ { objectID: 1, __position: 1, __queryID: "x" }, { objectID: 2, __position: 1, __queryID: "x" },  ]
hits = [ {objectID: 1}, {objectID: 2} ]
```js

packages/react-instantsearch-core/src/connectors/__tests__/connectHits.js
tkrugg: yes but there is only one item (in the list, so I made it simple) 😄

packages/react-instantsearch-core/src/connectors/connectHits.js
tkrugg: @samouss advised to keep the names more explicit in his review. They do look at little verbose. I have no preference.

packages/react-instantsearch-core/src/connectors/__tests__/connectHits.js
Haroenv: ohh I'm seeing the difference now

PR #2288, opened 2019-04-08
fix(compat): upgrade SearchBox lifecycle

francoischalifour:
This migrates away from deprecated React lifecycles the `SearchBox` component.

I converted the `componentWillReceiveProps` to `componentDidMount`. This resolves in another render (from one to two) when hitting <kbd>Enter</kbd> in `{ searchAsYouType: false }` mode. As discussed IRL, this is a tradeoff that we're ready to accept until it's proven to be an issue.

[Try the stories →](https://deploy-preview-2288--react-instantsearch.netlify.com/storybook/?selectedKind=SearchBox&selectedStory=default)

### DIFF
--- a/packages/react-instantsearch-dom/src/components/SearchBox.js
+++ b/packages/react-instantsearch-dom/src/components/SearchBox.js
@@ -115,19 +115,13 @@ class SearchBox extends Component {
     document.removeEventListener('keydown', this.onKeyDown);
   }
 
-  componentWillReceiveProps(nextProps) {
-    // @TODO: should component maybe be controlled, otherwise Derived
-
-    // Reset query when the searchParameters query has changed.
-    // This is kind of an anti-pattern (props in state), but it works here
-    // since we know for sure that searchParameters having changed means a
-    // new search has been triggered.
+  componentDidUpdate(prevProps) {
     if (
-      !nextProps.searchAsYouType &&
-      nextProps.currentRefinement !== this.props.currentRefinement
+      !prevProps.searchAsYouType &&
+      prevProps.currentRefinement !== this.props.currentRefinement
     ) {
       this.setState({
-        query: nextProps.currentRefinement,
+        query: this.props.currentRefinement,
       });
     }
   }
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-dom/src/components/SearchBox.js
```diff
@@ -115,19 +115,13 @@ class SearchBox extends Component {
...
   }
 
-  componentWillReceiveProps(nextProps) {
-    // @TODO: should component maybe be controlled, otherwise Derived
-
-    // Reset query when the searchParameters query has changed.
-    // This is kind of an anti-pattern (props in state), but it works here
-    // since we know for sure that searchParameters having changed means a
-    // new search has been triggered.
+  componentDidUpdate(prevProps) {
```
Haroenv: I think this should be `getDerivedStateFromProps`, so it happens before render

Haroenv: (especially since we are settings state exclusively inside it)

francoischalifour: According to the unit tests and the manual tests on the Storybook, it works as previously. There might be more renders though. I need to check that more thoroughly.

Haroenv: might be the case that it's batched, but not sure how to count that

francoischalifour: It's now rendered twice when hitting <kbd>Enter</kbd> in `{ searchAsYouType: false }` mode. I added more info about it in the PR description.

Haroenv: >You may call setState() immediately in componentDidUpdate() but note that it must be wrapped in a condition like in the example above, or you’ll cause an infinite loop. It would also cause an extra re-rendering which, while not visible to the user, can affect the component performance. If you’re trying to “mirror” some state to a prop coming from above, consider using the prop directly instead. Read more about why copying props into state causes bugs.

https://reactjs.org/docs/react-component.html#componentdidupdate

PR #2289, opened 2019-04-08
fix(compat): upgrade RangeInput lifecycle

francoischalifour:
This migrates away from deprecated React lifecycles the `RangeInput` component.

### DIFF
--- a/packages/react-instantsearch-dom/src/components/RangeInput.js
+++ b/packages/react-instantsearch-dom/src/components/RangeInput.js
@@ -35,8 +35,7 @@ export class RawRangeInput extends Component {
   componentDidUpdate(prevProps) {
     if (
       this.props.canRefine &&
-      (prevProps.canRefine !== this.props.canRefine ||
-        prevProps.currentRefinement.min !== this.props.currentRefinement.min ||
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-dom/src/components/RangeInput.js
```diff
@@ -32,24 +32,13 @@ export class RawRangeInput extends Component {
...
-
-    // In react@16.0.0 the call to setState on the inputs trigger this lifecycle hook
-    // because the context has changed (for react). I don't think that the bug is related
-    // to react because I failed to reproduce it with a simple hierarchy of components.
-    // The workaround here is to check the differences between previous & next props in order
-    // to avoid to override current state when values are not yet refined. In the react documentation,
-    // they DON'T categorically say that setState never run componentWillReceiveProps.
-    // see: https://reactjs.org/docs/react-component.html#componentwillreceiveprops
-
+  componentDidUpdate(prevProps) {
```
Haroenv: same here, I don't think it will behave correctly if this logic happens after render

francoischalifour: See https://github.com/algolia/react-instantsearch/pull/2288#discussion_r273374037 for more information.

PR #2290, opened 2019-04-08
fix(compat): upgrade RangeSlider lifecycle

francoischalifour:
This migrates away from deprecated React lifecycles the `RangeSlider` component and its 3rd party integrations.

[See stories →](https://deploy-preview-2290--react-instantsearch.netlify.com/storybook/?selectedKind=Integration%20With%20Other%20Libraries&selectedStory=Airbnb%20Rheostat)

### DIFF
--- a/packages/react-instantsearch-dom/src/widgets/RangeSlider.js
+++ b/packages/react-instantsearch-dom/src/widgets/RangeSlider.js
@@ -29,14 +29,17 @@ class Range extends React.Component {
 
   state = { currentValues: { min: this.props.min, max: this.props.max } };
 
-  componentWillReceiveProps(sliderState) {
-    // @TODO: Derived State, maybe render
-    if (sliderState.canRefine) {
+  componentDidUpdate(prevProps) {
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-dom/src/widgets/RangeSlider.js
```diff
@@ -29,14 +29,17 @@ class Range extends React.Component {
 
   state = { currentValues: { min: this.props.min, max: this.props.max } };
 
-  componentWillReceiveProps(sliderState) {
-    // @TODO: Derived State, maybe render
-    if (sliderState.canRefine) {
+  componentDidUpdate(prevProps) {
```
Haroenv: I also think that this should be derived state, to make sure it happens before render

francoischalifour: See https://github.com/algolia/react-instantsearch/pull/2288#discussion_r273374037 for more information.

PR #2296, opened 2019-04-09
feat(infiniteHits): add previous button

yannickcr:
**Summary**

The [InfiniteHits widget](https://www.algolia.com/doc/api-reference/widgets/infinite-hits/react/) is used to display a list of results with a “Show more” button. The user can click on this button to display more results without leaving its current page context.

Each time the user click on the “Show more” button the user will navigate to a new page (provided that [you implmented URL routing in your React integration](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/react/)).

However, in the case the user navigates directly to a particular page (direct link, previous button, page refresh) then the widget only displays the last result page.

This change add a previous button to the Infinite Hits widget to give the possibility to the user to load the previous result pages.

For implementation details see [the full RFC](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/infinite-hits-previous-page.md).

**Result**

A "with previous button" story has been added for the Instant Hits widget.

https://deploy-preview-2296--react-instantsearch.netlify.com/storybook/?path=/story/infinitehits--with-previous-button


stories/util.js
```diff
@@ -62,97 +62,135 @@ Wrap.defaultProps = {
...
-        href={sourceCodeUrl}
-        className="source-code-url"
-        target="_blank"
-        rel="noopener noreferrer"
-      >
-        <div>View source code</div>
-      </a>
-    </div>
-  ) : null;
+  // defaultProps added so that they're displayed in the JSX addon
```
samouss:
```suggestion

```

We don't have it anymore.

stories/util.js
```diff
@@ -62,97 +62,135 @@ Wrap.defaultProps = {
...
-  const hits = hitsElement || <CustomHits />;
+    this.state = {
+      searchState: props.initialSearchState,
+    };
+  }
 
-  const searchParameters = {
-    hitsPerPage: 3,
-    ...askedSearchParameters,
+  onSearchStateChange = ({ configure, ...searchState }) => {
```
samouss: Why do we filter out `configure`?

stories/util.js
```diff
@@ -62,97 +62,135 @@ Wrap.defaultProps = {
...
+    this.setState({ searchState });
+    if (this.props.onSearchStateChange) {
```
samouss: We can use a default prop to avoid the `if`.

stories/util.js
```diff
@@ -62,97 +62,135 @@ Wrap.defaultProps = {
...
+    hasPlayground: false,
+    searchParameters: {},
+    initialSearchState: {},
+  };
+
+  constructor(props) {
+    super();
 
-  const hits = hitsElement || <CustomHits />;
+    this.state = {
```
samouss: It's our internal setup for Storybook we can use hooks inside.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -45,39 +47,69 @@ export default createConnector({
...
       };
     }
 
-    const { hits, hitsPerPage, page, nbPages } = results;
+    const {
+      page,
+      hits,
+      hitsPerPage,
+      nbPages,
+      _state: { page: p, ...currentState } = {},
```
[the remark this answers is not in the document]
yannickcr: I tried to use the `searchState` but I got some issues with it, since it is not totally in sync with the `SearchParameters`.

When I update the query, the `getProvidedProps` method get called 2 times:

#### Initial State
`searchState`
```json
{"page": 3}
```
`SearchParameters`
```json
{"page": 2, "query": ""}
```

#### First call
`searchState`
```json
{"page": 1, "query": "g"}
```
`SearchParameters`
```json
{"page": 2, "query": ""}
```
Additionally `results.hits` still contains the old hits

#### Second call
`searchState`
```json
{"page": 1, "query": "g"}
```
`SearchParameters`
```json
{"page": 0, "query": "g"}
```
Now `results.hits` is containing the new hits.

In my case I want to reset the widget on any query or refinement changes and when I have the new hits to display, so on the second render. But if I make my condition on `searchState` instead of `SearchParameters` then I will reset the widget on the first render.

Another solution I have is to put a flag during the first render to mark the widget as "to be reset"  and to reset it on the second render. It works, but that's assuming that we will always want to reset on the following render, is it the case?

Do you see a better solution to this issue using `searchState`?

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -45,39 +47,69 @@ export default createConnector({
...
       hitsWithPositions,
       results.queryID
     );
 
-    if (page === 0) {
-      this._allResults = hitsWithPositionsAndQueryID;
-    } else if (page > this._previousPage) {
+    if (
+      this._firstReceivedPage === undefined ||
+      !isEqual(currentState, this._prevState)
```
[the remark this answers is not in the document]
yannickcr: It's a re-implementation of [what was done on `instantsearch.js`](https://github.com/algolia/instantsearch.js/pull/3675/files#diff-4ca1cb93f270a17eee0204844ee91ed9). I'm agree with you, it is a bit risky but sadly I did not find a better way to properly reset the widget in our current case. I am open to any suggestion if you have a better idea for the implementation :hugs:

stories/util.js
yannickcr: No reason, I think it's a left-over from a previous implementation, I'll remove it.

stories/util.js
yannickcr: Oh, didn't saw we were on React 16.8, I'll update the component to use hooks instead :+1:

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -45,39 +47,69 @@ export default createConnector({
     const results = getResults(searchResults, this.context);
 
     this._allResults = this._allResults || [];
-    this._previousPage = this._previousPage || 0;
+    this._prevState = this._prevState || '';
```
francoischalifour: `this._prevState` is expected to be an object, right?

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -45,39 +47,69 @@ export default createConnector({
...
+      this.context.ais.onInternalStateUpdate(
+        refineValue(searchState, nextValue, this.context, resetPage)
+      );
+    };
 
     return {
       hits: this._allResults,
+      hasPrevious,
       hasMore,
+      refine: () => refine(this._lastReceivedPage + 2),
```
francoischalifour: Can you add a comment to explain why the offset is 2?

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -62,22 +65,37 @@ export default createConnector({
...
+      this._allResults = [...hitsWithPositionsAndQueryID, ...this._allResults];
+      this._firstReceivedPage = page;
     }
 
+    const hasPrevious = this._firstReceivedPage > 0;
     const lastPageIndex = nbPages - 1;
     const hasMore = page < lastPageIndex;
-
-    this._previousPage = page;
+    const refine = index => {
```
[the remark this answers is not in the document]
yannickcr: [Same question](https://github.com/algolia/vue-instantsearch/pull/659#discussion_r274967396) by @Haroenv on the Vue implementation.

Copy-past of my response:
> Talked about it with @samouss during the RFC review ([discussion](https://github.com/algolia/instantsearch-rfcs/pull/16#discussion_r270315733)). I had `refine('previous')` and `refine('next')` but we agreed for `refineNext()` and `refinePrevious()`.
>
> But we can re-discuss it together :wink: 

If we decide to change it it should not be too much work, but it should be done in every flavor to stay consistent.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -62,22 +65,37 @@ export default createConnector({
...
+      );
+    };
 
     return {
       hits: this._allResults,
+      hasPrevious,
       hasMore,
+      refine: () => refine(this._lastReceivedPage + 2),
+      refinePrevious: () => refine(this._firstReceivedPage),
+      refineNext: () => refine(this._lastReceivedPage + 2),
```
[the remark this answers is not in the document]
yannickcr: I'm not sure to understand :thinking: , do you have an example of what you're thinking of?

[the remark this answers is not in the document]
francoischalifour: I meant storing the `refine` function from line 96 and reusing it for `refineNext`.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
francoischalifour: Sorry, my question was different. It was about reusing the [`refine` function hook](https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-own-widgets/react#lifecyle-and-api) from the `createConnector` object instead of returning it in `getProvidedProps`.

It might make the code more complex though.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -373,29 +406,56 @@ describe('connectInfiniteHits', () => {
...
+            nbPages: 3,
+          },
+        }
+      );
+
+      props.refinePrevious.apply(context);
+
+      expect(
+        context.context.ais.onInternalStateUpdate.mock.calls[0][0]
+      ).toEqual({ page: 2 });
```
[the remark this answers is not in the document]
yannickcr: Yes, added a comment to explain why. Let me know if it is clear enough.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
yannickcr: Correct. It was previously serialized as a string and I forgot to change the initial value when I changed it's type.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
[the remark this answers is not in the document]
yannickcr: Yes, added a comment to explain why. Let me know if it is clear enough.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
yannickcr: Oh, my bad, I see what you means now. I'll try this.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
yannickcr: Ho, understood :+1:

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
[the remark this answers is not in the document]
yannickcr: Reused the `refine` function, so the code is clearer and the comment is not needed anymore :slightly_smiling_face:

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
samouss: Ok, I'm fine with this solution.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
samouss: I'm fine with this implementation, it's working. Worst case we'll fix issue if we have some.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -373,7 +416,55 @@ describe('connectInfiniteHits', () => {
...
+            page: 2,
+            hitsPerPage: 2,
+            nbPages: 3,
+          },
+        }
+      );
+
+      props.refineNext.apply(context);
+
+      expect(context.refine.mock.calls[0][0]).toEqual(3);
```
samouss: We could use [`toHaveBeenCalledTimes`](https://jestjs.io/docs/en/expect#tohavebeencalledtimesnumber) + [`toHaveBeenLastCalledWith`](https://jestjs.io/docs/en/expect#tohavebeenlastcalledwitharg1-arg2-) to have a clearer intent.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -373,7 +416,55 @@ describe('connectInfiniteHits', () => {
...
+            page: 2,
+            hitsPerPage: 2,
+            nbPages: 3,
+          },
+        }
+      );
+
+      props.refinePrevious.apply(context);
+
+      expect(context.refine.mock.calls[0][0]).toEqual(1);
```
samouss: We could use [`toHaveBeenCalledTimes`](https://jestjs.io/docs/en/expect#tohavebeencalledtimesnumber) + [`toHaveBeenLastCalledWith`](https://jestjs.io/docs/en/expect#tohavebeenlastcalledwitharg1-arg2-) to have a clearer intent.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -480,6 +592,32 @@ describe('connectInfiniteHits', () => {
...
     });
 
+    it('prepend hits internally', () => {
+      const context = createMultiIndexContext();
+      const getProvidedProps = connect.getProvidedProps.bind(context);
+
+      const hits = [{}, {}];
+      const hits0 = [{}, {}];
+
+      const res1 = getProvidedProps(null, null, {
```
samouss: We could re-use the same name used for the single index, it's easier to understand.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -45,39 +47,63 @@ export default createConnector({
...
-
-    this._previousPage = page;
+    const refinePrevious = () => this.refine(this._firstReceivedPage - 1);
+    const refineNext = () => this.refine(this._lastReceivedPage + 1);
 
     return {
       hits: this._allResults,
+      hasPrevious,
       hasMore,
+      refine: refineNext,
```
samouss: The `refine` function is overridden by [`createConnector`](https://github.com/algolia/react-instantsearch/blob/10ae0a0c31664cf941de9b6d9688cc6903678ca3/packages/react-instantsearch-core/src/core/createConnector.js#L336-L341). The outcome is that `refine` is different than `refineNext` for the connected component. `refine` uses the `currentRefinement + 1` but the `refineNext` function uses the `_lastReceivedPage + 1`. It depends on some condition but the call to `refine` is a noop if we previously click on `refinePrevious`.

packages/react-instantsearch-dom/src/components/InfiniteHits.js
```diff
@@ -11,14 +11,29 @@ class InfiniteHits extends Component {
...
+            className={cx(
+              'loadPrevious',
+              !hasPrevious && 'loadPrevious--disabled'
+            )}
+            onClick={() => refinePrevious()}
+            disabled={!hasPrevious}
+          >
+            {translate('loadPrevious')}
+          </button>
+        ) : null}
```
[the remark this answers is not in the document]
samouss: @yannickcr You forgot to apply the feedback.

packages/react-instantsearch-dom/src/components/__tests__/InfiniteHits.js
```diff
@@ -46,27 +50,108 @@ describe('InfiniteHits', () => {
...
+        hasPrevious={false}
       />
     );
     expect(mockedRefine.mock.calls).toHaveLength(0);
     wrapped.find('button').simulate('click');
     expect(mockedRefine.mock.calls).toHaveLength(1);
   });
 
-  it('Button is disabled when it is the last page', () => {
+  it('calls refineNext when the load more button is clicked', () => {
```
samouss: From what I've understood this test is exactly the same than the one above. 

We can get rid of the previous one since `refine` is not provided anymore.

packages/react-instantsearch-dom/src/components/__tests__/InfiniteHits.js
```diff
@@ -46,27 +50,108 @@ describe('InfiniteHits', () => {
...
+      <InfiniteHits
+        refinePrevious={() => undefined}
+        refineNext={mockedRefine}
+        hitComponent={Hit}
+        hits={hits}
+        hasMore={true}
+        hasPrevious={false}
+      />
+    );
+    expect(mockedRefine.mock.calls).toHaveLength(0);
```
samouss: Prefer the usage of `toHaveBeenCalledTimes(0)`. We can replace it in the whole file.

packages/react-instantsearch-dom/src/components/InfiniteHits.js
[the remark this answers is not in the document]
yannickcr: Oups, sorry.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
yannickcr: Oh, I though I managed this case but I missed some details. I pushed a fix here https://github.com/algolia/react-instantsearch/pull/2296/commits/c91b8c0dbd7af0289148d32612f945c0df008442#diff-c3b7a7fe9c467c6bbe02a87f5b8729c4. Let me know if it's ok for you.

samouss: The issue persist the two function does not have the same behaviour.

issue

yannickcr: Rah, should be ok now. Thanks for pointing that out.

I also added a "with custom rendering" story to ease testing.

yannickcr: Rah, should be ok now. Thanks for pointing that out.

I also added a "with custom rendering" story to ease testing.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -87,10 +113,16 @@ export default createConnector({
     });
   },
 
-  refine(props, searchState) {
+  refine(
+    props,
+    searchState,
+    event,
+    index = this._lastReceivedPage !== undefined
```
samouss: We could have inline this inside the function body it would be simpler to understand.

yannickcr: I inlined it, but I do not find it simpler to understand :innocent:

PR #2304, opened 2019-04-10
fix(compat): upgrade React Native example and its lifecycles

francoischalifour:
This updates the React Native example with the newest [Create InstantSearch App](https://github.com/algolia/create-instantsearch-app) React Native template and the newest React lifecycles.

The previous template didn't work anymore on Expo. I tested this implementation and it works.

You can find the new lifecycle in [`Price.js`](https://github.com/algolia/react-instantsearch/pull/2304/files#diff-eca6b9c51246c9ab0d19a11181c0e503R87). The other changes are the new React Native template.


francoischalifour:
Not entirely sure why ESLint is complaining.

yannickcr:
> Not entirely sure why ESLint is complaining.

Seems the `yarn install` in the `examples/*` directories is made after running ESLint on the whole repository, so when `eslint-plugin-import` is trying to resolve the imports it cannot find the modules.

Some solutions are to install the examples dependencies before running ESLint, or to disable the rule for the projects in examples:

**examples/react-native/.eslintrc.js**
```js
module.exports = {
  extends: 'algolia/react',
  rules: {
    'import/no-unresolved': 'off',
  },
};
```

francoischalifour:
Thank you @yannickcr. I'll try adding the ESLint rule in the example.

examples/react-native/README.md
```diff
@@ -1,25 +1,21 @@
...
-```
+## Get started
 
-## Start the example
+To run this project locally, install the dependencies and run the local server:
 
 ```sh
-yarn install --no-lockfile
-
-yarn start //to see the app on your phone using the expo client
```
Haroenv: this information seems to be gone

examples/react-native/README.md
```diff
@@ -1,25 +1,21 @@
...
+# example-react-native
 
-It was created using `create-react-native-app`.
+_This project was generated with [create-instantsearch-app](https://github.com/algolia/create-instantsearch-app) by [Algolia](https://algolia.com)._
 
-## Clone the example
+This examples shows how to use `react-instantsearch` along with `react-native`.
 
-```
-curl https://codeload.github.com/algolia/react-instantsearch/tar.gz/master | tar -xz --strip=2 react-instantsearch-master/examples/react-native
```
Haroenv: is it consistent with the other readmes to no longer have this comment?

examples/react-native/package.json
```diff
@@ -1,37 +1,44 @@
...
-    "prop-types": "15.6.0",
-    "react": "16.8.4",
-    "react-dom": "16.8.4",
-    "react-instantsearch": "5.4.0",
-    "react-native": "0.52.0",
+    "prop-types": "15.7.2",
+    "react": "16.5.0",
+    "react-instantsearch-core": "5.4.0",
+    "react-instantsearch-native": "5.4.0",
+    "react-native": "https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz",
```
Haroenv: I wonder if renovate will be able to update this

francoischalifour: That's part of CISA. We don't want Renovate to update this because the Expo SDK is not compatible with all the React Native versions.

examples/react-native/README.md
francoischalifour: Which one? We still have the `yarn start` command in the README coming from CISA.

examples/react-native/README.md
francoischalifour: Indeed, I'll add it back.

examples/react-native/README.md
Haroenv: The lower ones, like `start ios` and the comment about expo

examples/react-native/README.md
```diff
@@ -1,6 +1,4 @@
-This examples shows how to use `react-instantsearch` along with `react-native`.
-
-It was created using `create-react-native-app`.
+This examples shows how to use React InstantSearch along with React Native.
```
Haroenv:
```suggestion
This example shows how to use React InstantSearch along with React Native.
```

probably needs to be updated in the other ones too?

francoischalifour: Yes, we can review all the examples' READMEs in another pass.

examples/react-native/README.md
francoischalifour: I added them back!

examples/react-native/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
samouss: You can drop this `.eslintrc`, the folder already contains one that inherits from the top level configuration.

examples/react-native/src/Categories.js
```diff
@@ -10,15 +10,15 @@ import {
...
-import { InstantSearch } from 'react-instantsearch/native';
+import { InstantSearch } from 'react-instantsearch-native';
+import Icon from 'react-native-vector-icons/FontAwesome';
 import {
   connectMenu,
   connectRefinementList,
   connectSearchBox,
   connectRange,
-} from 'react-instantsearch/connectors';
+} from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/Filters.js
```diff
@@ -9,14 +9,14 @@ import {
...
-import { InstantSearch } from 'react-instantsearch/native';
+import { InstantSearch } from 'react-instantsearch-native';
 import {
   connectCurrentRefinements,
   connectMenu,
   connectRange,
   connectRefinementList,
   connectSearchBox,
-} from 'react-instantsearch/connectors';
+} from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/Home.js
```diff
@@ -22,14 +22,14 @@ import {
   connectSortBy,
   connectRange,
   connectCurrentRefinements,
-} from 'react-instantsearch/connectors';
-import Highlight from './components/Highlight';
-import Spinner from './components/Spinner';
+} from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/Price.js
```diff
@@ -1,15 +1,15 @@
...
-import { InstantSearch } from 'react-instantsearch/native';
+import { InstantSearch } from 'react-instantsearch-native';
 import {
   connectRefinementList,
   connectSearchBox,
   connectRange,
   connectMenu,
-} from 'react-instantsearch/connectors';
-import Stats from './components/Stats';
+} from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/Rating.js
```diff
@@ -8,15 +9,14 @@ import {
...
 } from 'react-native';
-import { InstantSearch } from 'react-instantsearch/native';
+import { InstantSearch } from 'react-instantsearch-native';
 import {
   connectRefinementList,
   connectSearchBox,
   connectRange,
   connectMenu,
-} from 'react-instantsearch/connectors';
+} from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/Type.js
```diff
@@ -10,13 +10,13 @@ import {
...
 } from 'react-native';
-import { InstantSearch } from 'react-instantsearch/native';
+import { InstantSearch } from 'react-instantsearch-native';
 import {
   connectRefinementList,
   connectSearchBox,
   connectRange,
   connectMenu,
-} from 'react-instantsearch/connectors';
+} from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/components/Highlight.js
```diff
@@ -1,19 +1,22 @@
 import React from 'react';
 import { Text } from 'react-native';
-import { connectHighlight } from 'react-instantsearch/connectors';
+import { connectHighlight } from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/components/Spinner.js
```diff
@@ -1,12 +1,13 @@
 import React from 'react';
 import { ActivityIndicator, View, Dimensions } from 'react-native';
-import { connectStateResults } from 'react-instantsearch/connectors';
+import { connectStateResults } from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/src/components/Stats.js
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import { Button, View, Platform, Dimensions } from 'react-native';
-import { connectStats } from 'react-instantsearch/connectors';
+import { connectStats } from 'react-instantsearch-core';
```
samouss: We should import from `react-instantsearch-native`. The `core` package is only used internally.

examples/react-native/.eslintrc.js
francoischalifour: It is necessary because https://github.com/algolia/react-instantsearch/pull/2304#issuecomment-483241937.

samouss: I don't think so. The issue was present because we added an `.eslintrc` during the migration to new template. The previous example didn't have an `.eslintrc` but it fallbacks on the one inside the parent folder.

PR #2316, opened 2019-04-12
feat(voiceSearch): add voice search widget

eunjae-lee:
**Summary**

This adds a new widget, connector and component for voice search.

**Usage**

```
stories.add('default', () => (
  <WrapWithHits
    searchBox={false}
    hasPlayground={true}
    linkedStoryGroup="VoiceSearch"
  >
    <VoiceSearch searchAsYouSpeak={boolean} />
  </WrapWithHits>
)).add('fully customized', () => (
  <WrapWithHits
    searchBox={false}
    hasPlayground={true}
    linkedStoryGroup="VoiceSearch"
  >
    <VoiceSearch
      searchAsYouSpeak={searchAsYouSpeak}
      buttonComponent={ButtonComponent}
      StatusComponent={StatusComponent}
    />
  </WrapWithHits>
));
```

**Checklist**

- [x] component
- [ ] ~connector~ (using `connectSearchBox`)
- [x] widget
- [x] tests
- [x] stories


eunjae-lee:
This is a draft pull request to check if the implementation is okay for `connectVoiceSearch` and `VoiceSearch.tsx` before going further. It is not property typed. Here's what's done so far. Please let me know what you think.

## VoiceSearch.tsx

This components receives props from connector like

* isBrowserSupported
* isListening
* toggleListening
* voiceListeningState
* searchAsYouSpeak

When user clicks the button, it will execute `toggleListening` which triggers `toggleListeneing` of `VoiceSearchHelper` inside the connector. The function will start a recognition and according to the lifecycle, its state will change. And it will be delivered to the component via `voiceListeneingState`.

This component also receives these two props.

* buttonComponent
* statusComponent

It is to let user customise the UI like the way we provide templates in InstantSearch.js.

## connectVoiceSearch

This is a tricky part. We have `VoiceSearchHelper` copy&pasted from InstantSearch.js. As we know, it handles `SpeechRecognition` api. To keep the current structure of widget, connector and component, the instance of `VoiceSearchHelper` should be inside the connector, so that the connector can deliver the states from the instance to the component.

To achieve this, `connectVoiceSearch` doesn't export `createConnector(...)` but a function wrapping it. So `VoiceSearchHelper` is instantiated inside the function.

For `VoiceSearchHelper` to notify its internal changes(query or state), two callbacks are being used. In the current draft, the callbacks are set in `getProvidedProps` which is invoked inside the constructor of the component. And when the state changes, it calls `this.context.ais.widgetsManager.update();` to trigger re-render in order to deliver new `voiceListeningState` to the component. I'm not sure if this is the good way (or aligned in the existing convention).

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.ts
```diff
@@ -0,0 +1,72 @@
...
+    searchState,
+    context,
+    id,
+    '',
+    refinementsCallback
+  );
+}
+
+export default Composed => {
+  const helper = voiceSearchHelper();
```
Haroenv: I wouldn't call this `helper`, it will be confusing with the algoliasearch helper

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <path d="M12 1a3 3 0 0 0-3 3v8a3 3 0 0 0 6 0V4a3 3 0 0 0-3-3z" />
+        <path d="M19 10v2a7 7 0 0 1-14 0v-2" />
+        <line x1="12" y1="19" x2="12" y2="23" />
+        <line x1="8" y1="23" x2="16" y2="23" />
+      </svg>
+    );
+  }
+};
+
+const DefaultStatusComponent: React.FunctionComponent<InnerComponentProps> = ({
```
Haroenv:
```suggestion
const DefaultStatusComponent: React.FC<InnerComponentProps> = ({
```

this is more consistent with the rest of the code base

Haroenv:
```suggestion
const DefaultStatus: React.FunctionComponent<InnerComponentProps> = ({
```

We usually don't postfix with Component

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+  toggleListening: ToggleListening;
+  voiceListeningState: VoiceListeningState;
+  searchAsYouSpeak?: boolean;
+
+  translate: Translate;
+  buttonComponent?: React.FunctionComponent<InnerComponentProps>;
+  statusComponent?: React.FunctionComponent<InnerComponentProps>;
+};
+
+const DefaultButtonComponent: React.FunctionComponent<InnerComponentProps> = ({
```
Haroenv:
```suggestion
const DefaultButton: React.FunctionComponent<InnerComponentProps> = ({
```

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+  return (
+    <div className={cx('')}>
+      <button
+        className={cx('button')}
+        type="button"
+        title={translate('buttonTitle')}
+        onClick={onClick}
+        disabled={!isBrowserSupported}
+      >
+        <ButtonComponent {...innerProps} />
```
Haroenv:
```suggestion
        <Button {...innerProps} />
```

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+      </button>
+      <div className={cx('div')}>
+        <StatusComponent {...innerProps} />
```
Haroenv:
```suggestion
        <Status {...innerProps} />
```

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+const VoiceSearch = ({
+  translate,
+  isBrowserSupported,
+  isListening,
+  toggleListening,
+  voiceListeningState,
+  searchAsYouSpeak = false,
+
+  buttonComponent: ButtonComponent = DefaultButtonComponent,
```
Haroenv:
```suggestion
  buttonComponent: Button = DefaultButtonComponent,
```

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+  statusComponent: StatusComponent = DefaultStatusComponent,
```
Haroenv:
```suggestion
  statusComponent: Status = DefaultStatusComponent,
```

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.ts
eunjae-lee: good point. I've renamed it to `voiceSearch`. Any other suggestion is welcomed.

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.ts
```diff
@@ -0,0 +1,74 @@
+import createConnector from '../core/createConnector';
+import voiceSearchHelper from '../lib/voiceSearchHelper';
+import { refineValue, getCurrentRefinementValue } from '../core/indexUtils';
+
+function getId() {
+  return 'query';
+}
+
+/* tslint:disable:variable-name */
```
francoischalifour: You can use `tslint:disable-next-line` to avoid re-enabling it on the next line.

```js
// tslint:disable-next-line:variable-name
```

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  const getDefaultState = (status: string): VoiceListeningState => ({
+    status,
+    transcript: undefined,
+    isSpeechFinal: undefined,
+    errorCode: undefined,
+  });
+  let state: VoiceListeningState = getDefaultState(STATUS_INITIAL);
+  let recognition: SpeechRecognition | undefined;
+  let searchAsYouSpeak;
+  // tslint:disable-next-line:no-empty
```
francoischalifour: This rule should be disabled at the project level: https://github.com/algolia/react-instantsearch/blob/a94e73a53dbefab5383cd67f2d959d75c51a3eb2/tslint.json#L15

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  let onQueryChange: QueryChangeListener = () => {};
+  // tslint:disable-next-line:no-empty
+  let onStateChange: StateChangeListener = () => {};
+
+  const setOnQueryChange = (queryChangeListener: QueryChangeListener) =>
+    (onQueryChange = queryChangeListener);
```
francoischalifour: We can use curly braces for better comprehension.

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  const setOnStateChange = (stateChangeListener: StateChangeListener) =>
+    (onStateChange = stateChangeListener);
```
francoischalifour: We can use curly braces for better comprehension.

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const setSearchAsYouSpeak = value => (searchAsYouSpeak = value);
```
francoischalifour: We can use curly braces for better comprehension.

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,131 @@
...
+    (window as any).SpeechRecognition;
+  const getDefaultState = (status: string): VoiceListeningState => ({
+    status,
+    transcript: undefined,
+    isSpeechFinal: undefined,
+    errorCode: undefined,
+  });
+  let state: VoiceListeningState = getDefaultState(STATUS_INITIAL);
+  let recognition: SpeechRecognition | undefined;
+  let searchAsYouSpeak;
```
francoischalifour:
```suggestion
  let searchAsYouSpeak: boolean;
```

francoischalifour: Should this have a default value?

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
francoischalifour:
```suggestion
  const setSearchAsYouSpeak = (value: boolean) => (searchAsYouSpeak = value);
```

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+};
+
+const DefaultButton: React.FC<InnerComponentProps> = ({
+  status,
+  errorCode,
+  isListening,
+}) => {
+  if (status === 'error' && errorCode === 'not-allowed') {
+    return (
+      <svg
```
francoischalifour: We should be able to not repeat the outer `svg` element, only the inner elements differ.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,158 @@
...
+      <button
+        className={cx('button')}
+        type="button"
+        title={translate('buttonTitle')}
+        onClick={onClick}
+        disabled={!isBrowserSupported}
+      >
+        <Button {...innerProps} />
+      </button>
+      <div className={cx('div')}>
```
francoischalifour: This should be `cx('status')`.

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
```diff
@@ -0,0 +1,116 @@
...
+const renderAndMatchSnapshot = (element: JSX.Element) => {
+  const instance = renderer.create(element);
+  expect(instance.toJSON()).toMatchSnapshot();
+  instance.unmount();
+};
+
+const defaultProps = {
+  isBrowserSupported: true,
+  isListening: false,
+  // tslint:disable-next-line:no-empty
```
francoischalifour: This should be disabled at project-level.

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Enzyme, { mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import VoiceSearch from '../VoiceSearch';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+const renderAndMatchSnapshot = (element: JSX.Element) => {
```
francoischalifour: The type `React.ReactNode` might be more accurate.

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
```diff
@@ -0,0 +1,116 @@
...
+  voiceListeningState: {
+    status: 'initial',
+    transcript: undefined,
+    isSpeechFinal: undefined,
+    errorCode: undefined,
+  },
+};
+
+describe('VoiceSearch', () => {
+  it('does nothing', () => {
```
francoischalifour: That's useless?

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
francoischalifour: Is this function wrapper necessary? In latest tests I think we use Enzyme for this. There's less boilerplate.

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
eunjae-lee: Good. I've merged master to this branch and now I have it :)

packages/react-instantsearch-core/src/lib/voiceSearchHelper/index.ts
eunjae-lee: It is meant to be set by the connector, but I will put a default value to keep consistency because I put default values for onQueryChange and onStateChange.

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
eunjae-lee: @francoischalifour 
Oh the latest way is just using Enzyme for matching snapshot. I must've followed some old way doing it. Anyway, I have a question on this.

```js
      const wrapper = shallow(<VoiceSearch {...defaultProps} />);
```

This produced

```js
exports[`VoiceSearch Rendering with default props 1`] = `
<VoiceSearch
  isBrowserSupported={true}
  isListening={false}
  toggleListening={[Function]}
  translate={[Function]}
  voiceListeningState={
    Object {
      "errorCode": undefined,
      "isSpeechFinal": undefined,
      "status": "initial",
      "transcript": undefined,
    }
  }
/>
`;
```

Do you have any idea why it's not actual dom elements?

francoischalifour: Probably because of the serializer: https://github.com/algolia/react-instantsearch/blob/c02cd328e80066c8d3799beed714bb7cd0553153/jest.config.js#L5

eunjae-lee: Hmm, but I don't understand why 

```js
const wrapper = shallow(<QueryRuleCustomData {...props} />);
...
```

generates 

```js
exports[`QueryRuleCustomData expects to render the empty container with empty items 1`] = `
<div
  className="ais-QueryRuleCustomData"
/>
`;
```

According to the serializer, the result should've been something like this

```js
exports[`QueryRuleCustomData expects to render the empty container with empty items 1`] = `
<QueryRuleCustomData />
`;
```

eunjae-lee:
The commit `a1808db` has an insufficient title. It also adds render the title of the button according to `isBrowserSupported` value.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,148 @@
...
+  errorCode?: string;
+  isListening: boolean;
+  transcript?: string;
+  isSpeechFinal?: boolean;
+  isBrowserSupported: boolean;
+};
+
+type VoiceSearchProps = {
+  searchAsYouSpeak?: boolean;
+
```
francoischalifour: Except if there's a reason:

```suggestion
```

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,148 @@
...
+};
+
+const getDefaultButtonInnerElement = (
+  status: string,
+  errorCode: string | undefined,
+  isListening: boolean
+) => {
+  if (status === 'error' && errorCode === 'not-allowed') {
+    return (
+      <>
```
francoischalifour: This is the first time we use Fragments in the codebase. According to our React peer dependencies, it should be fine. Can you confirm @samouss?

packages/react-instantsearch-dom/src/lib/voiceSearchHelper/__tests__/index.ts
```diff
@@ -0,0 +1,172 @@
+import VoiceSearchHelper, { VoiceSearchHelperParams } from '..';
+
+// copied from InstantSearch.js
```
francoischalifour: You can add a [permanent link](https://help.github.com/en/articles/getting-permanent-links-to-files) to the original file.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+  private voiceSearch: VoiceSearchHelper;
+
+  constructor(props: VoiceSearchProps) {
+    super(props);
+    const { searchAsYouSpeak = false, refine } = props;
+    this.voiceSearch = voiceSearchHelper({
+      searchAsYouSpeak,
+      onQueryChange: query => refine(query),
+      onStateChange: () => {
+        this.setState(this.voiceSearch.getState());
```
samouss: We never clean the listeners on unmount, which might throw an error e.g. the widget is currently listening but for some reasons, it's unmounted from the page. The widget still listens but once the event fires and trigger the listener React throws an error.

> Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.

Note that we might have the same issue with InstantSearch.js. With the Vanilla version is would probably not happen that often but with the Vue & Angular integration the chances are higher.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+  VoiceListeningState,
+} from '../lib/voiceSearchHelper';
+const cx = createClassNames('VoiceSearch');
+
+type InnerComponentProps = {
+  status: string;
+  errorCode?: string;
+  isListening: boolean;
+  transcript?: string;
+  isSpeechFinal?: boolean;
```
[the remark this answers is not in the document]
eunjae-lee: My reason behind this is here: https://github.com/algolia/angular-instantsearch/pull/530#discussion_r283756680
What do you think?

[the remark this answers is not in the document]
samouss: I would have stick with only the boolean. If the value have to carry another meaning it probably means that the boolean is not the appropriate type. The information that the recognition didn't start could be provided by another value (actually `isListening` could probably used for that), or use an enum rather than a boolean. Anyway it doesn't have a huge impact.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+    });
+    this.state = this.voiceSearch.getState();
+  }
+
+  public render() {
+    const { status, transcript, isSpeechFinal, errorCode } = this.state;
+    const { isListening, isBrowserSupported } = this.voiceSearch;
+    const {
+      translate,
+      buttonComponent: Button = DefaultButton,
```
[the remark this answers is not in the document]
eunjae-lee: 100e0c0

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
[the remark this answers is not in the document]
eunjae-lee: It makes sense. A boolean variable containing three meanings shouldn't be a boolean. Now that there's no strong reason to store three meanings, I'll just change it to non-optional boolean. I'm going to do this change to IS.js and all the flavours.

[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/5fa8fa0163ac752cf78171118e6f0de1e66942f4

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import Enzyme, { mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import VoiceSearch from '../VoiceSearch';
+
+const mockGetState = jest.fn().mockImplementation(() => ({}));
+const mockIsBrowserSupported = jest.fn().mockImplementation(() => true);
+const mockIsListening = jest.fn();
+const mockToggleListening = jest.fn();
```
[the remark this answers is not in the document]
eunjae-lee: `../../lib/voiceSearchHelper` exports a function `voiceSearchHelper` and `voiceSearchHelper()` returns an object containing `getState, isBrowserSupported, isListening, toggleListening`.
Do you know how I can achieve this with inlining them?

```js
      const wrapper = mount(<VoiceSearch />);
      wrapper.find('button').simulate('click');
      expect(mockToggleListening).toHaveBeenCalledTimes(1);
```

Apparently the instance of `voiceSearchHelper` is not visible here. So I don't know how to get the mocked version of `toggleListening` here with inlining way.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,147 @@
...
+          type="button"
+          title={
+            isBrowserSupported()
+              ? translate('buttonTitle')
+              : translate('disabledButtonTitle')
+          }
+          onClick={this.onClick}
+          disabled={!isBrowserSupported()}
+        >
+          <Button {...innerProps} />
```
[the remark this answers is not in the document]
eunjae-lee: Does the name `buttonTextComponent` make sense to you?

packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export type VoiceSearchHelperParams = {
+  searchAsYouSpeak: boolean;
+  onQueryChange: (query: string) => void;
+  onStateChange: () => void;
+};
+
+export type VoiceListeningState = {
+  status: string;
+  transcript?: string;
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/commit/5fa8fa0163ac752cf78171118e6f0de1e66942f4

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
eunjae-lee: Okay, trying to fix it [here](https://github.com/algolia/instantsearch.js/pull/3779) first.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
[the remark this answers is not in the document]
samouss: Yep, it does.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -24,41 +24,30 @@ type VoiceSearchProps = {
...
-};
-
 const DefaultButton: React.FC<InnerComponentProps> = ({
   status,
   errorCode,
   isListening,
 }) => {
-  return (
+  return status === 'error' && errorCode === 'not-allowed' ? (
+    <svg
```
samouss: You could create a component for both `svg` element, it would help to understand what is rendered in which condition.

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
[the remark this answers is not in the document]
samouss: You're correct it's not visible. We could access it through the component instance but it's a terrible idea. A solution would have been to decoupled a bit the `VoiceSearch` component from the `VoiceSearchHelper`. We could have create the service outside of the component (at the widget level for example) to pass it as prop. It would have been a lot easier to test. Anyway no problem to keep it as is.

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    it('with custom component for button with isListening: true', () => {
+      const customButton = ({ isListening }) =>
+        isListening ? 'Stop' : 'Start';
+      mockIsListening.mockImplementation(() => true);
+
+      const wrapper = mount(<VoiceSearch buttonComponent={customButton} />);
+      expect(wrapper.find('button').text()).toBe('Stop');
+      mockIsListening.mockClear();
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/de79d8db76a484b3cd3e405af16151fcc15e9ab0

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/8f8e971714ec61e24640311ee4b76a4ebdc6b5df

https://github.com/algolia/react-instantsearch/pull/2316/commits/92cf94f9bd44939dda23a94dc86a263de80e7e0b

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/48d6dc4e4b479ad8fc418183112e284f91c9484d

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/dbca103abb105654c51992516a0970281f1d5b9f

stories/VoiceSearch.stories.tsx
```diff
@@ -0,0 +1,175 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { VoiceSearch, SearchBox } from 'react-instantsearch-dom';
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/82d5e6f0cd59fa1451fc104f4c9f14532e0c62f3

stories/VoiceSearch.stories.tsx
```diff
@@ -0,0 +1,175 @@
...
+    </WrapWithHits>
+  ))
+  .add('with a custom button text', () => {
+    const style = window.document.createElement('style');
+    window.document.head.appendChild(style);
+    [
+      `.custom-button-story .ais-VoiceSearch-button:hover {
+        background: inherit;
+      }`,
+    ].forEach(rule => (style.sheet as CSSStyleSheet).insertRule(rule));
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/53db919c7949237596de842671c9416defe29ab6

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    it('renders a specific title when it is disabled', () => {
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2316/commits/4d17a689890d12580ea52acd0e251c882c147953

packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+    if (state.status !== STATUS_ERROR) {
+      setState({ status: STATUS_FINISHED });
+    }
+  };
+
+  const stop = (): void => {
+    if (recognition) {
+      recognition.stop();
+      recognition = undefined;
```
samouss: We attach the listeners on `start` but we don't clear them on `stop`. Is it expected?

packages/react-instantsearch-dom/src/components/__tests__/VoiceSearch.tsx
```diff
@@ -22,6 +22,13 @@ jest.mock('../../lib/voiceSearchHelper', () => {
 Enzyme.configure({ adapter: new Adapter() });
 
 describe('VoiceSearch', () => {
+  afterEach(() => {
+    mockGetState.mockClear();
+    mockIsBrowserSupported.mockImplementation(() => true);
```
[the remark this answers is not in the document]
eunjae-lee: I just tested it and `mockClear` didn't reset it to the default implementation which was `() => true`.

> Resets all information stored in the mockFn.mock.calls and mockFn.mock.instances arrays.
> Often this is useful when you want to clean up a mock's usage data between two assertions.

https://jestjs.io/docs/en/mock-function-api.html#mockfnmockclear

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,146 @@
...
+  protected static defaultProps = {
+    buttonTextComponent: DefaultButtonText,
+    statusComponent: DefaultStatus,
+  };
+  private voiceSearchHelper: VoiceSearchHelper;
+
+  constructor(props: VoiceSearchProps) {
+    super(props);
+    const { searchAsYouSpeak = false, refine } = props;
+    this.voiceSearchHelper = createVoiceSearchHelper({
```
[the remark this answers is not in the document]
eunjae-lee: Thanks. I just read it and it makes sense to move the initalization to `componentDidMount`.

packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
eunjae-lee: Thanks for pointing it out. We need to clear them.

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -0,0 +1,146 @@
...
+  status: string;
+  errorCode?: string;
+  isListening: boolean;
+  transcript: string;
+  isSpeechFinal: boolean;
+  isBrowserSupported: boolean;
+};
+
+type VoiceSearchProps = {
+  searchAsYouSpeak?: boolean;
```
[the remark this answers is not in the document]
eunjae-lee: b93918a

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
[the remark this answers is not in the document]
eunjae-lee: 085d58f

packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
eunjae-lee: e5bb1ad

packages/react-instantsearch-core/src/index.js
```diff
@@ -51,3 +51,6 @@ export {
   default as connectToggleRefinement,
 } from './connectors/connectToggleRefinement';
 export { default as connectHitInsights } from './connectors/connectHitInsights';
+
+// Types
+export * from './types';
```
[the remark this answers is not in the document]
eunjae-lee: Right, I didn't notice it. It's fixed at 6509e7c.
To make sure of the change in the rollup config, I compared the `dist` of react-instantsearch, react-instantsearch-core, react-instantsearch-dom before and after this commit. And they're the same.

Haroenv:
I can't see the custom search box

eunjae-lee:
> I can't see the custom button example

Right. That's fixed at https://github.com/algolia/react-instantsearch/pull/2316/commits/9656c5a3f5b33b42de82b8add8a24b0504740ef6

packages/react-instantsearch-dom/src/index.js
```diff
@@ -65,3 +65,6 @@ export { default as QueryRuleCustomData } from './widgets/QueryRuleCustomData';
 
 // Utils
 export { createClassNames } from './core/utils';
+
+// voiceSearchHelper
+export { default as createVoiceSearchHelper } from './lib/voiceSearchHelper';
```
eunjae-lee: Unlike other flavours, in React-IS, this helper is directly used inside the VoiceSearch component. So we need to export this for users to create their own component.

PR #2326, opened 2019-04-15
chore(build): remove React propTypes from build

yannickcr:
**Summary**

The propTypes were still present in the production build, making the UMD bundles a bit heavier than needed.

This change adds the [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) plugin to the Babel configuration to remove the propTypes at build time.

With this change we gain a few KB on the final bundles.

**Result**

No visible changes.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -7,6 +7,7 @@ const clean = x => x.filter(Boolean);
 
 module.exports = api => {
   const isTest = api.env('test');
+  const isStorybook = api.env('storybook');
   const targets = {};
 
   if (!isTest) {
@@ -26,7 +27,11 @@ module.exports = api => {
       ],
       '@babel/preset-react',
     ],
-    plugins: ['@babel/plugin-proposal-class-properties', 'babel-plugin-lodash'],
+    plugins: clean([
+      '@babel/plugin-proposal-class-properties',
+      'babel-plugin-lodash',
+      !isStorybook && 'babel-plugin-transform-react-remove-prop-types',
+    ]),
     overrides: [
       {
         test: 'packages/*',
--- a/babel.config.js
+++ b/babel.config.js
@@ -7,7 +7,6 @@ const clean = x => x.filter(Boolean);
 
 module.exports = api => {
   const isTest = api.env('test');
-  const isStorybook = api.env('storybook');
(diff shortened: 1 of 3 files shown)


babel.config.js
```diff
@@ -26,7 +27,11 @@ module.exports = api => {
       ],
       '@babel/preset-react',
     ],
-    plugins: ['@babel/plugin-proposal-class-properties', 'babel-plugin-lodash'],
+    plugins: clean([
+      '@babel/plugin-proposal-class-properties',
+      'babel-plugin-lodash',
+      !isStorybook && 'babel-plugin-transform-react-remove-prop-types',
```
francoischalifour: We've been relying on PropTypes for development errors.

Example:

```jsx
<RefinementList />

// Warning: Failed prop type: The prop `attribute` is marked as required in 
// `AlgoliaRefinementList(RefinementListWidget)`, but its value is `undefined`.
```

We will lose this information with this config.

Can we benefit from this change only in the production bundle?

yannickcr: Yes, my intent was only to touch the UMD bundles but seems the CJS and ES builds are also affected.

I will update my change to only target the UMD bundles

yannickcr:
@samouss Seems it detected ours, at least I saw no differences in final bundles with and without the annotation comments.

package.json
```diff
@@ -125,27 +126,27 @@
...
       "maxSize": "40.50 kB"
     },
     {
       "path": "packages/react-instantsearch/dist/umd/Dom.min.js",
-      "maxSize": "63.75 kB"
+      "maxSize": "63.00 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
       "maxSize": "41.50 kB"
```
Haroenv: there's also proptypes in this package no?

samouss: Yes but those packages use the already transpiled version of the `react-instantsearch-core` & `react-instantsearch-dom` and I don't think Rollup runs Babel on them. I don't think it's an issue since this package is not documented anymore we only advocate for the platform specific package.

PR #2328, opened 2019-04-16
chore(typescript): transform typescript files in stories

eunjae-lee:
**Summary**

This enables us to write stories in TypeScript.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -33,7 +33,7 @@ module.exports = api => {
     ]),
     overrides: [
       {
-        test: 'packages/*',
+        test: ['packages/*', 'stories/*'],
         presets: ['@babel/preset-typescript'],
         plugins: [
           [


Haroenv:
I'm still not convinced of having public API in TS before it can easily be used by consumers

francoischalifour:
What didn't work before?

The [Query Rules stories](https://github.com/algolia/react-instantsearch/blob/c02cd328e80066c8d3799beed714bb7cd0553153/stories/QueryRuleCustomData.stories.tsx) are written in TypeScript.

eunjae-lee:
@francoischalifour okay let's forget and close this pr.
It failed to build with `yarn start` but now it does. 
I'm not sure if it's because I merged develop into this a few hours ago, or because I removed node_modules and re-installed.

@Haroenv By "public API", do you mean something like widget and connector?

Haroenv:
I meant having things that people look at (storybook source) in a way that they can't consume easily (TS)

PR #2339, opened 2019-04-17
refactor(root): remove createIndex & createInstantSearch

Haroenv:
- remove root div (replaced by just the provider)
- move logic for user agents to instantSearchManager

BREAKING CHANGE: removal of the factories (createInstantSearch & createIndex → InstantSearch & Index)
BREAKING CHANGE: DOM no longer has a div around root


Haroenv:
tests will likely fail

Haroenv:
I moved `addAlgoliaAgents` back from `utils.ts` → `createInstantSearchManager.js`, but still there's a lot more of lodash included in `umd/Connectors` for no clear reason :(

packages/react-instantsearch-core/src/components/Index.tsx
```diff
@@ -29,7 +31,6 @@ type State = {
  * @kind widget
  * @name <Index>
  * @propType {string} indexName - index in which to search.
```
samouss: They are outdated + the file is written in TypeScript. It's probably better to remove them.

packages/react-instantsearch-core/src/components/Index.tsx
```diff
@@ -58,36 +59,31 @@ type State = {
...
+
   constructor(props: InnerProps) {
     super(props);
 
     this.props.contextValue.onSearchParameters(
       this.getSearchParameters.bind(this),
       {
         ais: this.props.contextValue,
-        multiIndexContext: this.state.indexContext,
+        multiIndexContext: getIndexContext(props),
```
samouss: We can directly read the value stored inside the state.

packages/react-instantsearch-core/src/index.js
```diff
@@ -11,6 +9,8 @@ export { default as translatable } from './core/translatable';
 // Widgets
 export { default as Configure } from './widgets/Configure';
 export { default as QueryRuleContext } from './widgets/QueryRuleContext';
+export { default as Index } from './components/Index';
```
samouss: The two components have to move to widgets then.

packages/react-instantsearch-native/src/widgets/InstantSearch.tsx
```diff
@@ -0,0 +1,9 @@
+import { InstantSearch as CreateInstantSearch } from 'react-instantsearch-core';
+
+// @TODO: add React Native version as a user agent
+// @TODO: import & export prop type
+const InstantSearch: React.FC<any> = props => (
```
samouss: What is the purpose of this component at the moment?

packages/react-instantsearch-native/src/widgets/InstantSearch.js
```diff
@@ -1,8 +0,0 @@
-import { View } from 'react-native';
-import { createInstantSearch } from 'react-instantsearch-core';
-
-const InstantSearch = createInstantSearch({
```
samouss: We should probably have kept the pattern with the factory for the User Agent.

packages/react-instantsearch-core/src/components/__tests__/Index.js
```diff
@@ -55,7 +55,8 @@ describe('Index', () => {
   it('calls update if indexName prop changes', () => {
     const context = createContext();
 
-    const wrapper = shallow(
+    // componentDidUpdate wasn't called on `shallow`
```
samouss: That's unexpected though. It was working fine with the previous implementation?

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -3,6 +3,15 @@ import createWidgetsManager from './createWidgetsManager';
 import createStore from './createStore';
 import { HIGHLIGHT_TAGS } from './highlight';
 import { hasMultipleIndices } from './indexUtils';
+import { version as ReactVersion } from 'react';
+import version from './version';
+
+export function addAlgoliaAgents(searchClient) {
```
samouss: Why do we export the function?

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -3,6 +3,15 @@ import createWidgetsManager from './createWidgetsManager';
...
+  if (typeof searchClient.addAlgoliaAgent === 'function') {
+    searchClient.addAlgoliaAgent(`react (${ReactVersion})`);
+    searchClient.addAlgoliaAgent(`react-instantsearch (${version})`);
```
samouss: We should be able to add additional agent from the environment, DOM, Native, Server.

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -69,9 +65,9 @@ describe('createInstantSearchServer', () => {
         searchClient,
       };
 
-      shallow(<InstantSearch {...props} />);
+      mount(<InstantSearch {...props} />);
 
-      expect(searchClient.addAlgoliaAgent).toHaveBeenCalledTimes(2);
+      expect(searchClient.addAlgoliaAgent).toHaveBeenCalledTimes(4);
```
samouss: Why is it called four times?

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
Haroenv: This was because it was in utils first, good catch

packages/react-instantsearch-core/src/components/__tests__/Index.js
yannickcr: Seems to be a regression on Enzyme 3.9.0 https://github.com/airbnb/enzyme/issues/2020

Fixed in Enzyme 3.10.0 https://github.com/airbnb/enzyme/commit/ebdd4e782d25e7ecd1f5d71cc06ee693d8aedf7c

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
yannickcr: `<InstantSearch>`'s constructor will call `createInstantSearchManager` and `createInstantSearchManager`'s constructor will call `algoliasearchHelper`.

* `algoliasearchHelper`'s constructor calls it **one time**  with `JS Helper {version}`
* `algoliasearchHelper`'s constructor calls it a **second time**  (through `this.setClient`) with `JS Helper {version}` (not sure if this is wanted but it is the way it is done in the helper)
* `createInstantSearchManager`'s constructor calls `addAlgoliaAgents` that will call it **two more times**  with `react ({version})` and `react-instantsearch ({version})`

Before this change only the calls by `algoliasearchHelper` were made.

I agree that this is counter intuitive, a comment could be added here to explain why it has been called 4 times

packages/react-instantsearch-core/src/components/__tests__/Index.js
samouss: Ok then once it's rebased it will work.

packages/react-instantsearch-native/src/index.ts
```diff
@@ -6,6 +6,7 @@ export { translatable } from 'react-instantsearch-core';
 // Widget
 export { Configure } from 'react-instantsearch-core';
 export { QueryRuleContext } from 'react-instantsearch-core';
+export { Index } from 'react-instantsearch-core';
```
samouss: We have to export the widget from `react-instantsearch-core` otherwise can't use the `native` package.

```js
import { InstantSearch } from 'react-instantsearch-native';

// <InstantSearch /> is not exported anymore
```

packages/react-instantsearch-core/src/widgets/Index.tsx
```diff
@@ -99,18 +92,10 @@ class Index extends Component<InnerProps, State> {
     );
   }
 
-  componentWillReceiveProps(nextProps: InnerProps) {
-    // @TODO: DidUpdate
-    if (this.props.indexName !== nextProps.indexName) {
+  componentDidUpdate(prevProps: InnerProps) {
+    if (this.props.indexName !== prevProps.indexName) {
```
samouss: We might have the same issue raised inside the PR #2349 about the duplicate request. It's not easy to test right now because the multi-index API does not work anymore. We should fix it first, then test that we do have the correct behaviour.

yannickcr: Fixed in https://github.com/algolia/react-instantsearch/pull/2552 , should be ok now.

PR #2349, opened 2019-04-19
refactor(InstantSearch): migrate life cycles

Haroenv:
TODO: rebase once #2339 is merged in v6


packages/react-instantsearch-core/src/components/InstantSearch.tsx
```diff
@@ -137,64 +130,63 @@ class InstantSearch extends Component<Props, State> {
...
+    }
+
+    if (prevState.isControlled && !nextIsControlled) {
+      throw new Error(
+        "You can't switch <InstantSearch> from being controlled to uncontrolled"
+      );
+    }
+
+    return {
+      isControlled: nextIsControlled,
```
samouss: We can't switch from controlled to uncontrolled, it means that we can compute the value once for all. It was the case inside the previous implementation: the value was computed inside the constructor that's it. The check was performed inside cWRP. We can't do it there anymore but we can move the error inside cDU. It will throw an error anyway, sooner or later it doesn't matter.

packages/react-instantsearch-core/src/components/InstantSearch.tsx
```diff
@@ -137,64 +130,63 @@ class InstantSearch extends Component<Props, State> {
...
-      }));
-    }
-
-    if (this.props.refresh !== nextProps.refresh) {
-      if (nextProps.refresh) {
-        this.aisManager.clearCache();
-      }
-    }
+  isUnmounting: boolean = false;
+  aisManager: InstantSearchManager = createInstantSearchManager({
```
samouss: With the `StrictMode` the manager could be created [more than once](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects), it sounds incorrect. We'll create the Helper + listeners multiple times, even though they won't collide with each other we spend time for nothing. A  solution would be to move it inside cDU but we've to investigate how feasible it is. Worst case we at least have to check how it behaves inside `StrictMode`  with the current implementation.

packages/react-instantsearch-core/src/components/InstantSearch.tsx
```diff
@@ -137,64 +130,63 @@ class InstantSearch extends Component<Props, State> {
...
+      store: this.aisManager.store,
+      widgetsManager: this.aisManager.widgetsManager,
+      mainTargetedIndex: this.props.indexName,
+    },
+  };
 
-    if (this.props.searchClient !== nextProps.searchClient) {
-      this.aisManager.updateClient(nextProps.searchClient);
+  componentDidUpdate() {
+    if (this.props.refresh) {
```
samouss: The previous implementation was also responsible to detect whether or not the value was already `true`. It means that two consecutive render with `true` only clear the cache once. It prevents user that "forget" to revert the value to _always_  clear the cache on future render.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -72,12 +72,20 @@ export default function createInstantSearchManager({
...
+      return;
+    }
     addAlgoliaAgents(client);
     helper.setClient(client);
     search();
   }
 
-  function clearCache() {
+  let lastRefreshValue;
+  function clearCache(refreshValue) {
```
samouss: We never use the `clearCache` argument at the call site.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -72,12 +72,20 @@ export default function createInstantSearchManager({
   }
 
   function updateClient(client) {
+    if (helper.getClient() === client) {
```
samouss: Depends on the implementation of `InstantSearch` but I would rather branch inside the component rather than inside the function. It's easier to follow at the component level, it's common to have condition around such "side effects" inside the component.

packages/react-instantsearch-core/src/components/__tests__/InstantSearch.js
```diff
@@ -7,17 +7,31 @@ import { InstantSearchConsumer } from '../../core/context';
...
+const createFakeSearchClient = (rest = {}) => ({
+  search: jest.fn(() => {}),
+  ...rest,
+});
+
 jest.mock('../../core/createInstantSearchManager', () =>
   jest.fn(() => ({
     context: {},
+    updateIndex: () => {},
+    updateClient: () => {},
```
samouss: It would be much simpler to test with the manager as prop. It might be an improvement for a latter PR.

packages/react-instantsearch-core/src/components/InstantSearch.tsx
Haroenv: ah, I wasn't aware constructor also can be called multiple times. cDU happens after render though, that means we render a first time empty for no reason, no?

samouss: Yes that's why we have to check how it behaves with different solution.

yannickcr: If `createInstantSearchManager` does not have any side effect I'm more to keep it here. If we move it to cDU we'll delay the first render each time. Here, at worse, we make some extra work but it is not very significant imho.

packages/react-instantsearch-core/src/components/InstantSearch.tsx
```diff
@@ -180,7 +180,11 @@ class InstantSearch extends Component<Props, State> {
       );
     }
 
-    if (this.props.refresh) {
+    /*
+     * Clear cache when `refresh` changes to `true`.
+     * Prevents users to always clear the cache on render if they forget to revert it to `false`.
+     */
+    if (this.props.refresh !== prevProps.refresh && this.props.refresh) {
```
samouss: We should cover this use case with a test.

packages/react-instantsearch-core/src/components/InstantSearch.tsx
```diff
@@ -65,6 +57,7 @@ type Props = {
 };
 
 type State = {
+  isControlled: boolean;
```
samouss: We don't have to store it inside the state anymore.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -354,6 +354,9 @@ export default function createInstantSearchManager({
   }
 
   function updateIndex(newIndex) {
+    if (initialSearchParameters.index === newIndex) {
```
samouss: This check is already performed inside the component.

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -137,64 +129,74 @@ class InstantSearch extends Component<Props, State> {
...
+      this.aisManager.clearCache();
     }
 
     if (this.isControlled) {
-      this.aisManager.onExternalStateUpdate(nextProps.searchState);
+      this.aisManager.onExternalStateUpdate(this.props.searchState);
+    }
+
+    if (prevProps.indexName !== this.props.indexName) {
+      this.aisManager.updateIndex(this.props.indexName);
```
samouss: This update of the `indexName` now triggers two requests. It might be because gDSP & cDU take effect into two different mico-tasks (the searches are batched into a single mico-task). The first request is triggered by the call to `updateIndex`. The second one is triggered by the cDU of one of the widgets. The widgets perform a deep-equal on the props to determine whether or not we have to schedule a new request. It was fine with the previous implementation but it's not the case anymore. The `contextValue` which provided inside the props change because of the `mainTargetedIndex`.

yannickcr: I saw this, but this already present on the `v6` branch. I will try to fix this, but maybe not in this PR then.

samouss: Ok then we can merge with the issues.

PR #2350, opened 2019-04-19
fix(indexUtils): allow index with dots in it

Haroenv:
How it's done:
- remove usage of lodash.get & lodash.has
- refactor `getCurrentRefinementValue` to remove the callback argument
- split up body of `getCurrentRefinementValue` in three smaller functions

In the mean time, I did a small refactor, removing the callback to getCurrentRefinementValue, because it's useless.


packages/react-instantsearch-core/src/connectors/connectRange.js
```diff
@@ -65,37 +65,32 @@ function getCurrentRange(boundaries, stats, precision) {
 }
 
 function getCurrentRefinement(props, searchState, currentRange, context) {
-  const refinement = getCurrentRefinementValue(
+  let refinement = getCurrentRefinementValue(
```
samouss: We can inline the destruct assignment that we do [L76](https://github.com/algolia/react-instantsearch/pull/2350/files#diff-588750eca1a5cfabf9bda1e9070350ebR76).

packages/react-instantsearch-core/src/connectors/connectRange.js
```diff
@@ -65,37 +65,32 @@ function getCurrentRange(boundaries, stats, precision) {
...
+  if (typeof nextMin === 'string') {
+    nextMin = isFloatPrecision ? parseFloat(nextMin) : parseInt(nextMin, 10);
+  }
+
+  let nextMax = max;
+  if (typeof nextMax === 'string') {
+    nextMax = isFloatPrecision ? parseFloat(nextMax) : parseInt(nextMax, 10);
+  }
+
+  refinement = {
```
samouss: We can now use a `const` for the refinement.

packages/react-instantsearch-core/src/connectors/connectRefinementList.js
```diff
@@ -15,25 +15,24 @@ function getId(props) {
...
-          return [];
-        }
+    []
+  );
 
-        // Only one item was in the searchState but we know it should be an array
-        return [currentRefinement];
-      }
-      return currentRefinement;
+  if (typeof currentRefinement === 'string') {
```
samouss: We can invert the condition to avoid nested condition:

```js
function getCurrentRefinement(props, searchState, context) {
  const currentRefinement = getCurrentRefinementValue();

  if (typeof currentRefinement !== 'string') {
    return currentRefinement;
  }

  if (currentRefinement) {
    return [currentRefinement];
  }

  return [];
}
```

packages/react-instantsearch-core/src/core/__tests__/indexUtils.js
```diff
@@ -419,13 +405,32 @@ describe('utility method for manipulating the search state', () => {
...
         'refinement',
-        null,
-        () => {}
+        null
       );
 
       expect(value).toEqual('defaultRefinement');
     });
 
+    it('retrieve the current refinement value when impossible', () => {
```
samouss: I didn't understood the purpose of the test at first. We should mention the default value, WDYT?

packages/react-instantsearch-core/src/core/__tests__/indexUtils.js
```diff
@@ -419,13 +405,32 @@ describe('utility method for manipulating the search state', () => {
...
+
+      const value = getCurrentRefinementValue(
+        {},
+        searchState,
+        context,
+        'refinement',
+        defaultRefinement
+      );
+
+      expect(value).toEqual(defaultRefinement);
```
samouss:
```suggestion
      expect(value).toBe(defaultRefinement);
```

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -69,24 +69,27 @@ export function refineValue(
...
-          ...page,
-        },
-      }
-    : {
-        ...searchState.indices,
-        ...{
+  const state =
+    searchState &&
+    searchState.indices &&
+    searchState.indices.hasOwnProperty(indexId)
```
samouss:
```suggestion
    searchState.indices[indexId]
```

The only value possible for `indices` is an object it should be safe to call it with the bracket notation.

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -69,24 +69,27 @@ export function refineValue(
...
-          ...searchState.indices[indexId],
-          ...nextRefinement,
-          ...page,
-        },
-      }
-    : {
-        ...searchState.indices,
-        ...{
+  const state =
+    searchState &&
```
samouss: We can drop this first condition because in both close of the ternary we use `searchState.indices`.

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -69,24 +69,27 @@ export function refineValue(
...
+            ...searchState.indices[indexId],
             ...nextRefinement,
             ...page,
           },
-        },
-      };
+        }
+      : {
+          ...searchState.indices,
+          ...{
```
samouss: This part is useless we can directly use the `[indexId]`:

```js
{
  ...searchState.indices,
  [indexId]: {
    ...nextRefinement,
    ...page
  }
}
```

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -109,29 +112,32 @@ function refineMultiIndexWithNamespace(
...
-    : {
-        ...searchState.indices,
-        ...{
-          [indexId]: {
-            [namespace]: nextRefinement,
-            ...page,
+        }
+      : {
+          ...searchState.indices,
+          ...{
```
samouss: SEE ABOVE

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -109,29 +112,32 @@ function refineMultiIndexWithNamespace(
...
-    ? {
-        ...searchState.indices,
-        [indexId]: {
-          ...searchState.indices[indexId],
-          ...{
-            [namespace]: {
-              ...searchState.indices[indexId][namespace],
-              ...nextRefinement,
+  const state =
+    searchState &&
```
samouss: SEE ABOVE

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -109,29 +112,32 @@ function refineMultiIndexWithNamespace(
...
+    searchState.indices &&
+    searchState.indices.hasOwnProperty(indexId)
+      ? {
+          ...searchState.indices,
+          [indexId]: {
+            ...searchState.indices[indexId],
+            ...{
+              [namespace]: {
```
samouss: SEE ABOVE

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -161,44 +167,99 @@ function getNamespaceAndAttributeName(id) {
...
-      currentRefinement = namespace
-        ? get(searchState.indices[`${indexId}`][namespace], attributeName)
-        : get(searchState.indices[indexId], id);
-    } else {
-      currentRefinement = namespace
-        ? get(searchState[namespace], attributeName)
-        : get(searchState, id);
-    }
+  const hasRefinementsValue = hasRefinements({
+    context,
```
samouss: Inside both functions, the `context` is only used to detect whether or not we are in a multi-index context. We can hoist the call at this level and provide the value directly to the functions rather than delegate the computation.

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -161,44 +167,99 @@ function getNamespaceAndAttributeName(id) {
...
+  searchState,
+  indexId,
+  attributeName,
+  id,
+}) {
+  const multiIndex = hasMultipleIndices(context);
+
+  if (multiIndex && namespace) {
+    return (
+      searchState.indices[indexId][namespace] &&
```
samouss: We already perform those checks inside the `hasRefinements` function. It should be safe to directly access the value. The comment applies to the rest of the function.

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -109,29 +112,32 @@ function refineMultiIndexWithNamespace(
...
-        [indexId]: {
-          ...searchState.indices[indexId],
-          ...{
-            [namespace]: {
-              ...searchState.indices[indexId][namespace],
-              ...nextRefinement,
+  const state =
+    searchState &&
+    searchState.indices &&
+    searchState.indices.hasOwnProperty(indexId)
```
samouss: SEE ABOVE

packages/react-instantsearch-core/src/core/indexUtils.js
Haroenv: I agree, although it would mean the value can no longer be falsy

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -161,44 +156,84 @@ function getNamespaceAndAttributeName(id) {
...
-      currentRefinement = namespace
-        ? get(searchState.indices[`${indexId}`][namespace], attributeName)
-        : get(searchState.indices[indexId], id);
-    } else {
-      currentRefinement = namespace
-        ? get(searchState[namespace], attributeName)
-        : get(searchState, id);
-    }
+  const multiIndex = hasMultipleIndices(context);
+  const args = {
```
samouss: IMO it was easier with the value inlined inside the call.

Haroenv: was mainly  to keep them in the same order

PR #2357, opened 2019-04-23
fix(createConnector): new life cycles

Haroenv:
IFW-671

**Summary**

- move willMount -> constructor
- move willReceiveProps -> didUpdate

This didn't have an impact on more requests, but did cause one more render if you update a prop

**Result**

there's no more old life cycles!

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createConnector.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createConnector.js
@@ -686,7 +686,7 @@ describe('createConnector', () => {
       expect(update).toHaveBeenCalledTimes(0);
     });
 
-    it('triggers an onSearchStateChange on props change with transitationState', () => {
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-core/src/core/__tests__/createConnector.js
```diff
@@ -723,7 +723,7 @@ describe('createConnector', () => {
 
       expect(onSearchStateChange).toHaveBeenCalledTimes(1);
       expect(onSearchStateChange).toHaveBeenCalledWith({
-        hello: 'there', // the instance didn't have the next props yet
+        hello: 'again', // the instance didn't have the next props yet
```
Haroenv:
```suggestion
        hello: 'again',
```

This comment was necessary, since this used to be stale state, but is now up-to-date state, which is correct

Haroenv:
the bundlesize change here is unexpected, I'm not sure we can rely on it

PR #2366, opened 2019-04-26
feat(voice): add additionalQueryParameters

### LINKED ISSUE #2874: Set Language in Voice Search Component
Hi, is possible adding for VoiceSearch component a parameter for the language? I see that in InstantSearch JS library is possible.

marieglr:
Two new props:
- `additionalQueryParameters: () => Partial<SearchParameters>` (only applied when voice search is mounted)
- `language: string` (iso 639-1), the parameter sent to Algolia will be always the short version

New connector: `connectVoiceSearch`

known limitation: additional query parameters will stay applied as long as the Voice Search widget is mounted, meaning they can cause stale values if you switch input method without unmounting (limitation of the architecture of React InstantSearch)

<small>The connector is written in JS instead of TS, because it's mostly copy-pasted from searchBox, and that would be needless extra work otherwise #hackathon</small>

fixes #2874

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
+++ b/packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
@@ -1,0 +1,156 @@
+import PropTypes from 'prop-types';
+import createConnector from '../core/createConnector';
+import {
+  cleanUpValue,
+  refineValue,
+  getCurrentRefinementValue,
+  getIndexId,
+} from '../core/indexUtils';
+
+function getId() {
+  return 'query';
+}
+
+function getAdditionalId() {
+  return 'additionalVoiceParameters';
+}
+
+function getCurrentRefinementQuery(props, searchState, context) {
+  const id = getId();
+  const currentRefinement = getCurrentRefinementValue(
+    props,
+    searchState,
+    context,
+    id,
+    ''
+  );
+
+  if (currentRefinement) {
+    return currentRefinement;
+  }
+  return '';
+}
+
+function getCurrentRefinementAdditional(props, searchState, context) {
+  const id = getAdditionalId();
+  const currentRefinement = getCurrentRefinementValue(
+    props,
+    searchState,
+    context,
+    id,
+    ''
+  );
+
+  if (currentRefinement) {
+    return currentRefinement;
+  }
+  return {};
+}
+
+function refine(props, searchState, nextRefinement, context) {
+  const id = getId();
+  const voiceParams = getAdditionalId();
+
+  const nextValue = {
+    [id]: nextRefinement,
+    [voiceParams]:
+      typeof props.additionalQueryParameters === 'function'
+        ? {
+            queryLanguages: props.language
+              ? [props.language.split('-')[0]]
+              : undefined,
+            ignorePlurals: true,
+            removeStopWords: true,
+            optionalWords: nextRefinement,
+            ...props.additionalQueryParameters({ query: nextRefinement }),
+          }
+        : {},
+  };
+  const resetPage = true;
+  return refineValue(searchState, nextValue, context, resetPage);
+}
+
+function cleanUp(props, searchState, context) {
+  const interimState = cleanUpValue(searchState, context, getId());
+  return cleanUpValue(interimState, context, getAdditionalId());
+}
+
+export default createConnector({
+  displayName: 'AlgoliaVoiceSearch',
+
+  propTypes: {
+    defaultRefinement: PropTypes.string,
+  },
+
+  getProvidedProps(props, searchState, searchResults) {
+    return {
+      currentRefinement: getCurrentRefinementQuery(props, searchState, {
+        ais: props.contextValue,
+        multiIndexContext: props.indexContextValue,
+      }),
+      isSearchStalled: searchResults.isSearchStalled,
+    };
+  },
+
+  refine(props, searchState, nextRefinement) {
+    return refine(props, searchState, nextRefinement, {
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
+    });
+  },
+
+  cleanUp(props, searchState) {
+    return cleanUp(props, searchState, {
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -142,6 +145,8 @@ class VoiceSearch extends Component<VoiceSearchProps, VoiceListeningState> {
   };
 }
 
+<VoiceSearch additionalQueryParameters={({nose: bone}) => {}} />
```
Haroenv:
```suggestion
```

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -142,6 +145,8 @@ class VoiceSearch extends Component<VoiceSearchProps, VoiceListeningState> {
...
+
```
Haroenv:
```suggestion
```

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
```diff
@@ -0,0 +1,156 @@
...
+function refine(props, searchState, nextRefinement, context) {
+  const id = getId();
+  const voiceParams = getAdditionalId();
+
+  const nextValue = {
+    [id]: nextRefinement,
+    [voiceParams]:
+      typeof props.additionalQueryParameters === 'function'
+        ? {
+            queryLanguages: props.language
```
Haroenv: queryLanguages should be set based on language, the other ones when you pass the argument. We changed this in InstantSearch.js

eunjae-lee:
@Haroenv could you review the recent commits? 
ab3bd42 c31b590 
Is this what you meant by https://github.com/algolia/react-instantsearch/pull/2366#discussion_r334365481 ?

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
```diff
@@ -0,0 +1,159 @@
...
+      ? {
+          ignorePlurals: true,
+          removeStopWords: true,
+          optionalWords: nextRefinement,
+          ...props.additionalQueryParameters({ query: nextRefinement }),
+        }
+      : {};
+  const nextValue = {
+    [id]: nextRefinement,
+    [voiceParams]: {
```
Haroenv: I think this part isn't tested yet (the id itself)

eunjae-lee: here it is 8ef899d

eunjae-lee:
> looks ready to me now @eunjae-lee, just maybe tests for multi-index (as far as I can tell it will just work)

added some tests for multi index 349b03e

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
```diff
@@ -0,0 +1,159 @@
...
+  refineValue,
+  getCurrentRefinementValue,
+  getIndexId,
+} from '../core/indexUtils';
+
+function getId() {
+  return 'query';
+}
+
+function getAdditionalId() {
```
yannickcr: Why not using constants? (same for `getId`)

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
```diff
@@ -0,0 +1,159 @@
...
+    searchState,
+    context,
+    id,
+    ''
+  );
+
+  if (currentRefinement) {
+    return currentRefinement;
+  }
+  return {};
```
yannickcr: Not sure about the expected return value here.

In `getCurrentRefinementQuery` the return value can be `currentRefinement` or a string `''`.
In `getCurrentRefinementAdditional` the return value can be `currentRefinement` or an object `{}`.

I don't know what type `currentRefinement` is returning but it seems to me that the return type is at least inconsistent is one of the 2 functions.

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
```diff
@@ -0,0 +1,159 @@
...
+      }
+    );
+
+    return searchParameters
+      .setQuery(query)
+      .setQueryParameters(additionalParams);
+  },
+
+  getMetadata(props, searchState) {
+    const id = getId(props);
```
yannickcr: `getId` does not takes any parameters

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
Haroenv: consistent with the other connectors.

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
Haroenv: currentRefinement is a string for the query, and an object for the additional, so it's consistent, no?

Haroenv:
> Would it also be possible to do the connector in TypeScript?

I wrote this in the original description, but since it's mostly the same as search box, and the patterns for connectors aren't yet clear in React InstantSearch, I didn't do this in this PR

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
Haroenv:
```suggestion
    const id = getId();
```

packages/react-instantsearch-core/src/connectors/connectVoiceSearch.js
yannickcr: Ok, as seen together `getCurrentRefinementValue` always return a string when id is `getId()` return value, and always an object with id is `getAdditionalId()` return value. So no inconsistency here :+1:

Haroenv:
accidentally merged into next instead of master, but fixed in 3a45b2c4

PR #2391, opened 2019-05-02
Small accessibility fixes

catamphetamine:
**Summary**

Adds things like `type="button"`, `aria-label`, `tabIndex={-1}` on `aria-hidden`, also makes `<form/>` optional for `<SearchBox/>`.

**Result**

We're using that on our website.


samouss:
Thanks a lot for the initiative! Accessibility is something that we postponed for a long time but we'll most likely tackle those issue in the coming months. React InstantSearch is not the only flavor available, we also have the Angular, Vanilla, and Vue version. Those libraries follow the same DOM "specification" to easily apply a common theme. We'll revamp all of them at the same time after our accessibility research. We'll take your inputs into account for the changes we'll apply to the libraries. 

Thanks again! Hope you understand.

PR #2427, opened 2019-05-12
Add an onChange prop to the Pagination widget

ValentinH:
**Summary**

Because of #164, I'd like to manage the scroll to top logic manually. Also, Thanks to this, I'll be able to animate the scroll to top when the page change.

**Result**

This adds a `onChange(page)` prop to the `<Pagination>` widget so the developer can implements custom logic when the user changes the page.

### DIFF
--- a/packages/react-instantsearch-dom/src/components/Pagination.js
+++ b/packages/react-instantsearch-dom/src/components/Pagination.js
@@ -62,6 +62,7 @@ class Pagination extends Component {
     padding: PropTypes.number,
     totalPages: PropTypes.number,
     className: PropTypes.string,
+    onChange: PropTypes.func,
   };
 
   static defaultProps = {
@@ -87,6 +88,14 @@ class Pagination extends Component {
     };
   }
 
+  onSelect = page => {
+    const { currentRefinement, onChange, refine } = this.props;
+    refine(page);
+    if (page !== currentRefinement && onChange) {
+      this.props.onChange(page);
+    }
+  };
+
   render() {
     const {
       listComponent: ListComponent,
@@ -98,7 +107,6 @@ class Pagination extends Component {
       showPrevious,
       showNext,
       showLast,
-      refine,
       createURL,
       canRefine,
       translate,
@@ -170,7 +178,7 @@ class Pagination extends Component {
           {...otherProps}
           cx={cx}
           items={items}
-          onSelect={refine}
+          onSelect={this.onSelect}
           createURL={createURL}
           canRefine={canRefine}
         />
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-dom/src/components/Pagination.js
```diff
@@ -87,6 +88,14 @@ class Pagination extends Component {
     };
   }
 
+  onSelect = page => {
+    const { currentRefinement, onChange, refine } = this.props;
+    refine(page);
+    if (page !== currentRefinement && onChange) {
+      this.props.onChange(page);
```
Haroenv: I wonder if it would make more sense to call onChange inside every refine itself to solve this generically

ValentinH: Wouldn't this imply that it will also be called when a refinement is made via another widget? This is basically what I'd like to avoid as reported in  #164.

Haroenv: No, I meant the onChange prop of each widget could be called when its own refine function is called. This can not be called from another widget

ValentinH: OK I see 🙂. Indeed, it would be more generic. Do I have a way to add this on each widget without modifying each of them?

Haroenv: It has a quite big impact on the library as a whole, so we will discuss what this means for the whole library as a  team. 

Because there's a generic solution for this (in createConnector's refine method) this is a change we'd better do in a major version. 

One complicated thing about calling onChange ourselves, without making the components "controllable" is that we would also need to accept a value, which will be harder to implement

ValentinH: Any chance we could still add it only here for now to see if there is really a need before generalizing?

Haroenv:
Your issue can be solved by using `connectSortBy` and doesn't require listening to events on the pagination component. https://www.algolia.com/doc/api-reference/widgets/scroll-to/react/#connector-param-provided-hasnotchanged

We will take your feedback into account, and if there's more use cases for a real onChange / value combo that would make sense. I don't exactly have a time line for this, since only very infrequently we get feedback about use cases like this. I hope you understand.

Have a nice day!

ValentinH:
Hey @Haroenv, thanks for this answer. However, this does not solve my issue since the `hasNotChanged` is also true when changing an other refinement than `page`. This is basically what was explained in #164.

In my case, I'm trying to implement a scroll-to-top behavior when changing the page of the search result of CodeSandbox. However, I don't want to scroll to the top when changing another refinement like the filters on the right. See demo below (especially when adding a Tag refinement):

algolia-scrollto

Haroenv:
Is that with `scrollOn="page"` ([ref](https://www.algolia.com/doc/api-reference/widgets/scroll-to/react/?language=vue#widget-param-scrollon)? Can you make a sandbox with that behaviour, I can't seem to be able to reproduce it on my end :)

PR #2432, opened 2019-05-13
refactor(lodash): pick

Haroenv:
replaced by destructuring

### DIFF
--- a/packages/react-instantsearch-dom/src/components/HierarchicalMenu.js
+++ b/packages/react-instantsearch-dom/src/components/HierarchicalMenu.js
@@ -1,4 +1,3 @@
-import { pick } from 'lodash';
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { translatable } from 'react-instantsearch-core';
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-dom/src/components/HierarchicalMenu.js
```diff
@@ -28,6 +27,7 @@ class HierarchicalMenu extends Component {
     className: PropTypes.string,
     limit: PropTypes.number,
     showMoreLimit: PropTypes.number,
+    isEmpty: PropTypes.bool,
```
eunjae-lee: Does this exist? I can't find it from [the doc](https://www.algolia.com/doc/api-reference/widgets/hierarchical-menu/react/) and `connectHierarchicalMenu.js`.

tkrugg: yes, I guess we should remove it here and in other parts of the code that assume there is a isEmpty prop.

Haroenv: there's no more instances of the isEmpty prop, so this removal removed them all

PR #2434, opened 2019-05-13
refactor(lodash): has

Haroenv:
replace mostly by item[key] === undefined or hasOwnProperty where rendering `undefined`  could be an option

remaining usages of `has` are removed in #2350

### DIFF
--- a/packages/react-instantsearch-core/src/core/translatable.js
+++ b/packages/react-instantsearch-core/src/core/translatable.js
@@ -22,7 +22,7 @@ export default function translatable(defaultTranslations) {
         const { translations } = this.props;
 
         const translation =
-          translations && translations[key] !== undefined
+          translations && translations.hasOwnProperty(key)
             ? translations[key]
             : defaultTranslations[key];
 
--- a/packages/react-instantsearch-core/src/core/translatable.js
+++ b/packages/react-instantsearch-core/src/core/translatable.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react';
-import { has } from 'lodash';
 
 const withKeysPropType = keys => (props, propName, componentName) => {
   const prop = props[propName];
--- a/packages/react-instantsearch-core/src/core/translatable.js
+++ b/packages/react-instantsearch-core/src/core/translatable.js
@@ -23,7 +23,7 @@ export default function translatable(defaultTranslations) {
         const { translations } = this.props;
 
         const translation =
-          translations && has(translations, key)
+          translations && translations[key] !== undefined
             ? translations[key]
             : defaultTranslations[key];
 
--- a/packages/react-instantsearch-dom/src/components/LinkList.js
+++ b/packages/react-instantsearch-dom/src/components/LinkList.js
@@ -35,7 +35,7 @@ export default class LinkList extends Component {
       <ul className={cx('list', !canRefine && 'list--noRefinement')}>
         {items.map(item => (
           <li
-            key={has(item, 'key') ? item.key : item.value}
+            key={item.key === undefined ? item.value : item.key}
             className={cx(
               'item',
               item.selected && !item.disabled && 'item--selected',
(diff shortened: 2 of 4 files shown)


packages/react-instantsearch-core/src/core/translatable.js
```diff
@@ -23,7 +23,7 @@ export default function translatable(defaultTranslations) {
         const { translations } = this.props;
 
         const translation =
-          translations && has(translations, key)
+          translations && translations[key] !== undefined
```
Haroenv: this is what potentially is a breaking change, let's decide if it's important enough to not break / leave as is

francoischalifour: If I understand correctly, the previous version considered `undefined` as a value (because the key exists), and therefore displayed "undefined", right? IMO this wasn't expected, then it's not a breaking change.

If the behavior changed, we should add a test to keep track of it.

Haroenv: That’s what I thought too, should cover it with a test to be sure

Haroenv: decided not to have this behaviour in ab81d0b

packages/react-instantsearch-dom/src/components/LinkList.js
```diff
@@ -35,7 +34,7 @@ export default class LinkList extends Component {
       <ul className={cx('list', !canRefine && 'list--noRefinement')}>
         {items.map(item => (
           <li
-            key={has(item, 'key') ? item.key : item.value}
+            key={item.key === undefined ? item.value : item.key}
```
francoischalifour: Why not use `hasOwnProperty` everywhere?

Haroenv: Because you can only put undefined on purpose here with transformItems, which I can’t see having a reason. There’s already many more places we check for undefined indiscriminately. 

With translate I used hasOwnProperty because it’s a more dynamic piece of code, where passing undefined might be expected. 

However, I just realised that undefined doesn’t render nothing, meaning it can’t really have a reason to be used on purpose. Maybe that means we should revert the hasOwnProperty, and do use the default in that case. It can’t simply be a truthy check though

francoischalifour: IMO having the default check with `undefined` was fine even though it could be "breaking" (although the behavior would probably never happen).

We should be consistent across this file anyway. Strictly speaking, `hasOwnProperty` seems to better replace `lodash.has`, but it's fine using an `undefined` comparison here.

samouss: I think you can even simplify the check to `Boolean(...)`. It was overkill to use the `has` function.

packages/react-instantsearch-core/src/core/translatable.js
```diff
@@ -23,7 +22,7 @@ export default function translatable(defaultTranslations) {
         const { translations } = this.props;
 
         const translation =
-          translations && has(translations, key)
+          translations && translations.hasOwnProperty(key)
```
Haroenv: So, should this be reverted to a regular boolean check  (which technically would be breaking) or not?

francoischalifour: I would say _yes_.

samouss: Nope, not this one.

francoischalifour: True!

Haroenv: perfect, so I'll leave this as-is

packages/react-instantsearch-dom/src/components/LinkList.js
```diff
@@ -45,7 +44,7 @@ export default class LinkList extends Component {
           >
             {item.disabled ? (
               <span className={cx('link')}>
-                {has(item, 'label') ? item.label : item.value}
+                {item.label === undefined ? item.value : item.label}
```
samouss: IMO the `=== undefined` is overkill. We can stick with a `Boolean` check.

PR #2442, opened 2019-05-15
refactor(lodash): isEmpty

Haroenv:
reusing objectHasKeys strategy from the helper

IFW-736

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
+++ b/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
@@ -1,5 +1,6 @@
-import { find, isEmpty } from 'lodash';
+import { find } from 'lodash';
 import PropTypes from 'prop-types';
+import { objectHasKeys } from '../core/utils';
 import createConnector from '../core/createConnector';
 import {
   cleanUpValue,
@@ -156,7 +157,7 @@ export default createConnector({
     if (!items.some(item => item.value === '')) {
       items.push({
         value: '',
-        isRefined: isEmpty(refinedItem),
+        isRefined: !objectHasKeys(refinedItem),
         noRefinement: !stats,
         label: 'All',
       });
--- a/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
+++ b/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
@@ -156,7 +156,7 @@ export default createConnector({
     if (!items.some(item => item.value === '')) {
       items.push({
         value: '',
-        isRefined: refinedItem && refinedItem.length > 0,
+        isRefined: refinedItem && refinedItem.length === 0,
         noRefinement: !stats,
         label: 'All',
       });
--- a/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
+++ b/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import createConnector from '../core/createConnector';
-import { objectHasKeys, find } from '../core/utils';
+import { find } from '../core/utils';
 import {
   cleanUpValue,
   refineValue,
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch-core/src/core/utils.ts
```diff
@@ -62,3 +62,7 @@ export function addQueryID(hits, queryID) {
     __queryID: queryID,
   }));
 }
+
+export function objectHasKeys(object) {
```
francoischalifour: We should type the newly added utils.

Haroenv: I wasn't sure to type it as `any` or `undefined`

samouss: `unknown` would be better.

packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
```diff
@@ -156,7 +157,7 @@ export default createConnector({
     if (!items.some(item => item.value === '')) {
       items.push({
         value: '',
-        isRefined: isEmpty(refinedItem),
+        isRefined: !objectHasKeys(refinedItem),
```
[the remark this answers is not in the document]
Haroenv: we also need to check the existence of `refinedItem` thoug; it can be `undefined`

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -116,7 +115,7 @@ const createInstantSearchServer = algoliasearch => {
 
     const helper = algoliasearchHelper(client, sharedParameters.index);
 
-    if (isEmpty(derivedParameters)) {
+    if (Object.keys(derivedParameters).length === 0) {
```
samouss: We can simply the condition to `!Object.keys(derivedParameters).length` the return value is always a number.

packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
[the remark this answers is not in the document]
samouss: It looks like you didn't apply the feedback.

packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
```diff
@@ -157,7 +156,7 @@ export default createConnector({
     if (!items.some(item => item.value === '')) {
       items.push({
         value: '',
-        isRefined: isEmpty(refinedItem),
+        isRefined: refinedItem && refinedItem.length > 0,
```
Haroenv:
```suggestion
        isRefined: refinedItem && refinedItem.length === 0,
```

packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
```diff
@@ -157,7 +156,7 @@ export default createConnector({
     if (!items.some(item => item.value === '')) {
       items.push({
         value: '',
-        isRefined: isEmpty(refinedItem),
+        isRefined: refinedItem && refinedItem.length === 0,
```
samouss: Actually is it really an array? The return value of `find` is either `T | undefined`.

Haroenv: very good point, need to investigate this

packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
```diff
@@ -157,7 +156,7 @@ export default createConnector({
     if (!items.some(item => item.value === '')) {
       items.push({
         value: '',
-        isRefined: isEmpty(refinedItem),
+        isRefined: refinedItem === undefined,
```
samouss:
```suggestion
        isRefined: !refinedItem,
```

Haroenv: Might be a bit much of a micro-optimization, but at least there’s less branches that wag

PR #2446, opened 2019-05-15
refactor(lodash): replace find

Haroenv:
Chosen to duplicate the funciton in core & dom, because we have no good way to share functionality otherwise

IFW-740

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
+++ b/packages/react-instantsearch-core/src/connectors/connectNumericMenu.js
@@ -1,6 +1,7 @@
-import { find, isEmpty } from 'lodash';
+import { isEmpty } from 'lodash';
 import PropTypes from 'prop-types';
 import createConnector from '../core/createConnector';
+import { find } from '../core/utils';
 import {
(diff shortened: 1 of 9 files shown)


yannickcr:
This is not specific to this PR but one point that bother me is that we're adding some helper functions to replace native methods that are not available in older browsers (IE in that case) while it is not always needed.

It makes sense for AMD bundles as our polyfill should be lighter than core-js one (not as spec compliant but should be enough for our use-case) but CJS or ES bundles will already be recompiled with polyfills if they need to be executed in an environment that do not support those methods. So with CJS and ES the consumer will likely end with 2 find functions in its final bundle (our helper + babel polyfill) in old environments or with an extraneous helper in newer environment, making it's bundle heavier.

Not blocking for the current lodash migration (since it's already improving original the bundle size a lot :tada: ) but I think it could be good to have a reflection on this subject afterward.

Haroenv:
Yes, I think we can reflect on this later and have maybe some more size change if we manage to deduplicate or remove these helper methods

PR #2455, opened 2019-05-16
refactor(lodash): remove difference

Haroenv:
Both are arrays & always defined, so it's really enough to find all props in this._props (prevProps), but not in `props`

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectConfigure.js
+++ b/packages/react-instantsearch-core/src/connectors/connectConfigure.js
@@ -1,4 +1,4 @@
-import { omit, difference } from 'lodash';
+import { omit } from 'lodash';
 import createConnector from '../core/createConnector';
 import {
   refineValue,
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-core/src/connectors/connectConfigure.js
```diff
@@ -22,8 +22,9 @@ export default createConnector({
   transitionState(props, prevSearchState, nextSearchState) {
     const id = getId();
     const { children, contextValue, indexContextValue, ...items } = props;
+    const propKeys = Object.keys(props);
     const nonPresentKeys = this._props
```
samouss: The name of the variable is confusing, it looks like we already have the non present keys. `previousPropsKeys` sounds a better name. We can also perform the call to `Object.keys` at this stage to be sync with `propsKeys` (which should be renamed to `propsKeys` notice the `S`).

PR #2458, opened 2019-05-16
refactor(lodash): replace isFinite

Haroenv:
lodash.isFinite is just `typeof value == 'number' && nativeIsFinite(value);`, so that's what I implemented

IFW-745

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectRange.js
+++ b/packages/react-instantsearch-core/src/connectors/connectRange.js
@@ -150,12 +150,8 @@ function refine(props, searchState, nextRefinement, currentRange, context) {
   const nextMinAsNumber = !isMinReset ? parseFloat(nextMin) : undefined;
   const nextMaxAsNumber = !isMaxReset ? parseFloat(nextMax) : undefined;
 
-  const isNextMinValid =
-    isMinReset ||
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-core/src/connectors/connectRange.js
```diff
@@ -151,8 +150,8 @@ function refine(props, searchState, nextRefinement, currentRange, context) {
   const nextMinAsNumber = !isMinReset ? parseFloat(nextMin) : undefined;
   const nextMaxAsNumber = !isMaxReset ? parseFloat(nextMax) : undefined;
 
-  const isNextMinValid = isMinReset || _isFinite(nextMinAsNumber);
-  const isNextMaxValid = isMaxReset || _isFinite(nextMaxAsNumber);
+  const isNextMinValid = isMinReset || isFinite(nextMinAsNumber);
```
francoischalifour: This is not exactly the same as before since we don't check for the `number` type before, although it should work here.

Haroenv:
```
> isFinite(NaN)
< false
> isFinite(undefined)
< false

> Number.isFinite(NaN)
< false
> Number.isFinite(undefined)
< false
```

AFAICT the behaviour is the same for any value that can be passed in here, since the real danger of isFinite is passing it strings of numbers

PR #2460, opened 2019-05-16
refactor(lodash): isPlainObject

Haroenv:
The values this function was used for seemed like they can only be arrays or objects to me.

We will only want to empty out "breadth-first", so empty objects will stay in the output if they became empty because of the function itself.

isEmpty here is removed in #2442, and still works the same (but will need a rebase)

The newly added tests also pass with the previous implementation

The function is only used in `onSearchStateChange` as a response to a `cleanUp` (caused by unmount) on a widget

IFW-743

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/utils.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/utils.js
@@ -38,7 +38,7 @@ describe('utils', () => {
     });
   });
 
-  describe.only('remove empty key', () => {
+  describe('remove empty key', () => {
     it('empty key should be removed', () => {
       const state = {
         query: '',
--- a/packages/react-instantsearch-core/src/core/__tests__/utils.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/utils.js
@@ -38,7 +38,7 @@ describe('utils', () => {
     });
   });
 
-  describe('remove empty key', () => {
+  describe.only('remove empty key', () => {
     it('empty key should be removed', () => {
       const state = {
         query: '',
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-core/src/core/__tests__/utils.js
```diff
@@ -38,7 +38,7 @@ describe('utils', () => {
     });
   });
 
-  describe('remove empty key', () => {
+  describe.only('remove empty key', () => {
```
Haroenv:
```suggestion
  describe('remove empty key', () => {
```

packages/react-instantsearch-core/src/core/utils.ts
```diff
@@ -33,13 +33,19 @@ export const defer = f => {
   resolved.then(f);
 };
 
-export const removeEmptyKey = obj => {
+const isPlainObject = (value: any): value is object =>
```
Haroenv: I'm not sure if this type assertion can / needs to be any better

yannickcr: I think this is the correct way to do this, TypeScript seems to do the same thing in [the documentation](https://www.typescriptlang.org/docs/handbook/advanced-types.html#typeof-type-guards)

francoischalifour: I think this is were you're supposed to use `unknown` ([see resource I sent yesterday](https://mariusschulz.com/blog/typescript-3-0-the-unknown-type#narrowing-the-unknown-type)):

```suggestion
const isPlainObject = (value: unknown): value is object =>
```

PR #2461, opened 2019-05-16
refactor(lodash): get

Haroenv:
IFW-742

Now accepts (just like lodash.get): 

- `a.b.c`
- `a[5].c`
- `a.5.c` (even if it would be invalid JS, we just won't document this)

does _not_ accept (lodash.get does)

- `a['b.c']`

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/utils.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/utils.js
@@ -157,6 +157,98 @@ describe('utils', () => {
     });
   });
 
+  describe('getPropertyByPath', () => {
+    it('returns undefined on non-object root', () => {
+      expect(utils.getPropertyByPath(false, 'fake')).toBe(undefined);
+      expect(utils.getPropertyByPath(undefined, 'fake')).toBe(undefined);
+      expect(utils.getPropertyByPath(null, 'fake.nested')).toBe(undefined);
+    });
+
+    it('returns path if exists', () => {
+      expect(utils.getPropertyByPath({ dog: true }, 'dog')).toBe(true);
+      expect(
+        utils.getPropertyByPath(
+          { i: { like: { properties: false } } },
+          'i.like.properties'
+        )
+      ).toBe(false);
+      expect(
+        utils.getPropertyByPath({ true: { nested: 'ok' } }, 'true.nested')
+      ).toBe('ok');
+    });
+
+    it('accepts a pre-split path as array', () => {
+      expect(utils.getPropertyByPath({ dog: true }, ['dog'])).toBe(true);
+      expect(
+        utils.getPropertyByPath({ i: { like: { properties: false } } }, [
+          'i',
+          'like',
+          'properties',
+        ])
+      ).toBe(false);
+      expect(
+        utils.getPropertyByPath({ true: { nested: 'ok' } }, ['true', 'nested'])
+      ).toBe('ok');
+    });
+
+    it('returns undefined if does not exist', () => {
+      expect(
+        utils.getPropertyByPath(
+          { name: { known: { value: '' } } },
+          'name.unkown'
+        )
+      ).toBe(undefined);
+
+      expect(utils.getPropertyByPath({ name: false }, 'name.unkown')).toBe(
+        undefined
+      );
+    });
+
+    it('returns indexed path if exists', () => {
+      expect(
+        utils.getPropertyByPath({ array: ['a', 'b', 'c'] }, 'array.2')
+      ).toBe('c');
+      expect(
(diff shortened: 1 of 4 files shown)


Haroenv:
now that I think of it, maybe we should enforce the array form instead of the string form, it would allow trivially for `['lvl1', 0, 'lvl2']` without the string parsing problems. 

I initially didn't think of this because it was a breaking change, but now that it's breaking anyway, maybe it's best to go for something correct

eunjae-lee:
As a point of a user, a string `lvl0[0].lvl2` looks more natural than an array `['lvl1', 0, 'lvl2']`.

> without the string parsing problems.

What kind of problem do we have if we have to parse string?

Haroenv:
the problem would be that we would be parsing strings, which is error-prone, as well as more code than necessary.

eunjae-lee:
Since it's a breaking change anyway, we could support two ways: a simple string concatenated with ".", and a fuction to get the attribute from a hit. WDYT?

samouss:
IMO we can stick with this behavior. This is what we have in other implementation IIRC. We can think about it once we have a user that complain or a real use case for it. WDYT?

Haroenv:
seems good to me, if people have more advanced use cases I think we might be able to solve them another way. I don't see `[0]` etc. used since we handle arrays

packages/react-instantsearch-core/src/core/highlight.js
```diff
@@ -1,4 +1,5 @@
-import { get } from 'lodash';
+const getPropertyByPath = (object, path) =>
+  path.split('.').reduce((current, key) => current && current[key], object);
```
samouss: We can add a test to cover this function.

packages/react-instantsearch-core/src/core/__tests__/utils.js
```diff
@@ -137,4 +137,103 @@ describe('utils', () => {
...
+        )
+      ).toBe(undefined);
+
+      expect(
+        utils.getPropertyByPath({ array: ['a', 'b', 'c'] }, 'array[4]')
+      ).toBe(undefined);
+      expect(
+        utils.getPropertyByPath(
+          { array: [{ letter: 'a' }, { letter: 'b' }, { letter: 'c' }] },
+          'array[5]letter'
```
eunjae-lee: not important, but you wanted to say `array[5].letter` ?

Haroenv: good point, both would fail, which is why I missed the error

packages/react-instantsearch-core/src/core/__tests__/utils.js
```diff
@@ -137,4 +137,96 @@ describe('utils', () => {
       ]);
     });
   });
+
+  describe('getPropertyByPath', () => {
+    it('returns undefined on non-object root', () => {
+      expect(utils.getPropertyByPath(false, 'fake')).toBe(undefined);
```
samouss:
```suggestion
      expect(utils.getPropertyByPath(false, 'fake')).toBeUndefined();
```

packages/react-instantsearch-core/src/core/__tests__/utils.js
```diff
@@ -137,4 +137,96 @@ describe('utils', () => {
...
+          { i: { like: { properties: false } } },
+          'i.like.properties'
+        )
+      ).toBe(false);
+      expect(
+        utils.getPropertyByPath({ true: { nested: 'ok' } }, 'true.nested')
+      ).toBe('ok');
+    });
+
+    it('accepts a pre-split path as array', () => {
```
samouss: We should add a test for this kind of input: `['i', 'like.properties']`.

packages/react-instantsearch-core/src/core/utils.ts
```diff
@@ -62,3 +62,26 @@ export function addQueryID(hits, queryID) {
...
+ * ); // true
+ *
+ * @param object Source object to query
+ * @param path either an array of properties, or a string form of the properties, separated by .
+ */
+export const getPropertyByPath = (object: object, path: string[] | string) =>
+  (Array.isArray(path)
+    ? path
+    : path.replace(/\[(\d+)]/g, '.$1').split('.')
+  ).reduce((current, key) => (current ? current[key] : undefined), object);
```
samouss: We loop over the full path even if the one of the member returns `undefined`. With recursion we might be able to avoid this unnecessary work. Worst case I'm fine with this implementation it's simple enough.

Haroenv: True, however it’s very rare to see deep paths, so there’s no need to early exit here IMO

PR #2462, opened 2019-05-16
refactor(lodash): orderBy

Haroenv:
The utility is copied from the helper (which passed all lodash tests), I put it inline in the connector, since it's not used in any other spot.

IFW-746


packages/react-instantsearch-core/src/connectors/connectMenu.js
```diff
@@ -31,6 +30,82 @@ function getCurrentRefinement(props, searchState, context) {
...
+    }
+
+    // This ensures a stable sort in V8 and other engines.
+    // See https://bugs.chromium.org/p/v8/issues/detail?id=90 for more details.
+    return object.index - other.index;
+  });
+
+  return result.map(res => res.value);
+}
+
```
tkrugg: could we copy the tests too?
it's complex enough that it might be worth it to "privately export" it and throughly test it.

Haroenv: It’s an option, since I don’t think anyone will try to look for this in this particular file

samouss: The call to the `orderBy` is not that useful I think. We use the function to sort only when the SFFV is allowed **AND** the facet values are **NOT** from the SFFV. Which means that we can use the `sortBy` option of `getFacetValues` if we really have to. WDYT?

tkrugg: That sounds good as long as it's not a breaking change. The less code the better. 😄

Haroenv: changed in adbe458

packages/react-instantsearch-core/src/connectors/__tests__/connectMenu.js
```diff
@@ -424,16 +424,16 @@ describe('connectMenu', () => {
...
         {
           name: 'oy',
           isRefined: true,
           count: 10,
         },
+        {
+          name: 'wat',
+          isRefined: false,
+          count: 20,
+        },
```
tkrugg: I guess since now we no longer "order the facets by `isRefined`" but rather let getFacetValues handle it, the description of the test should change to something that "uses the getFacetValue default order.
And check that we're passing it `undefined` when `searchable && !isFromSearch`.

Haroenv: Makes sense, I’ll change the test description

packages/react-instantsearch-core/src/core/indexUtils.js
```diff
@@ -8,13 +8,20 @@ export function getIndexId(context) {
...
+  if (searchResults.results.hits) {
+    return searchResults.results;
   }
+
+  const indexId = getIndexId(context);
+  if (searchResults.results[indexId]) {
+    return searchResults.results[indexId];
+  }
+
+  return null;
```
samouss: We could avoid the double `null` return with the condition `searchResults.results` inline: 

```js
export function getResults(searchResults, context) {
  if (searchResults.results && searchResults.results.hits) {
    return searchResults.results;
  }

  const indexId = getIndexId(context);
  if (searchResults.results && searchResults.results[indexId]) {
    return searchResults.results[indexId];
  }

  return null;
}
```

IMO it's easier to read but I let you chose which one you want to implement.

packages/react-instantsearch-core/src/connectors/__tests__/connectMenu.js
samouss: You forget to change the name of the test?

packages/react-instantsearch-core/src/core/indexUtils.js
Haroenv: yes, thought of that as an alternative, went for this style because it's less comparisons

Haroenv:
```js
export function getResults(searchResults, context) {
  if (searchResults.results) {
    if (searchResults.results.hits) {
      return searchResults.results;
    }

    const indexId = getIndexId(context);
    if (searchResults.results[indexId]) {
      return searchResults.results[indexId];
    }
  }

  return null;
}
```

this is also somewhat easier since it has a return less, but all three styles suggested are fine I think

packages/react-instantsearch-core/src/connectors/connectMenu.js
```diff
@@ -146,28 +145,23 @@ export default createConnector({
...
-
-    const sortedItems =
-      searchable && !isFromSearch
-        ? orderBy(
-            items,
-            ['isRefined', 'count', 'label'],
-            ['desc', 'desc', 'asc']
-          )
-        : items;
+      : results
```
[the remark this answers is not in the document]
Haroenv:
```js
    let items;
    if (isFromSearch) {
      items = searchForFacetValuesResults[attribute].map(v => ({
        label: v.value,
        value: getValue(v.value, props, searchState, {
          ais: props.contextValue,
          multiIndexContext: props.indexContextValue,
        }),
        _highlightResult: { label: { value: v.highlighted } },
        count: v.count,
        isRefined: v.isRefined,
      }));
    } else {
      items = results
        .getFacetValues(attribute, {
          sortBy: searchable ? undefined : defaultSortBy,
        })
        .map(v => ({
          label: v.name,
          value: getValue(v.name, props, searchState, {
            ais: props.contextValue,
            multiIndexContext: props.indexContextValue,
          }),
          count: v.count,
          isRefined: v.isRefined,
        }));
    }
```

would this be your suggestion then?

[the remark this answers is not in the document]
samouss: Yes, ternaries on multiple lines are hard to follow.

packages/react-instantsearch-core/src/connectors/__tests__/connectMenu.js
```diff
@@ -416,7 +416,53 @@ describe('connectMenu', () => {
...
     });
 
-    it('when search for facets values is activated order the item by isRefined first', () => {
+    it('if not searchable: uses a static sortBy order', () => {
+      const results = {
+        getFacetValues: jest.fn(() => []),
+        getFacetByName: () => true,
+        hits: [],
+      };
+      results.getFacetValues.mockClear();
```
samouss: This call is useless since the implementation is defined above.

packages/react-instantsearch-core/src/connectors/__tests__/connectMenu.js
```diff
@@ -416,7 +416,53 @@ describe('connectMenu', () => {
...
+      results.getFacetValues.mockImplementation(() => [
```
samouss: Why not inlining the mock implementation inside the declaration?

Haroenv: I didn't write this, only edited it. I tried out using a real helper, but it was a lot of boilerplate and much harder to set up by the way

packages/react-instantsearch-core/src/connectors/__tests__/connectMenu.js
```diff
@@ -416,32 +416,80 @@ describe('connectMenu', () => {
...
     });
 
-    it('when search for facets values is activated order the item by isRefined first', () => {
+    it('if not searchable: uses a static sortBy order', () => {
       const results = {
         getFacetValues: jest.fn(() => []),
         getFacetByName: () => true,
         hits: [],
       };
       results.getFacetValues.mockClear();
```
samouss:
```suggestion

```

packages/react-instantsearch-core/src/core/indexUtils.js
samouss: @Haroenv Take one or the other suggestion is clearer that what we have right now.

packages/react-instantsearch-core/src/connectors/connectMenu.js
```diff
@@ -135,18 +134,24 @@ export default createConnector({
...
+          multiIndexContext: props.indexContextValue,
+        }),
+        _highlightResult: { label: { value: v.highlighted } },
+        count: v.count,
+        isRefined: v.isRefined,
+      }));
+    } else {
+      items = results
+        .getFacetValues(attribute, {
+          sortBy: searchable ? undefined : defaultSortBy,
```
samouss: Rather than relying on the default one from the Helper (which might change). We could specify the sort that we want to apply in this case.

Haroenv: we will keep this as undefined, because the other widgets also pass undefined

PR #2466, opened 2019-05-17
refactor(lodash): replace omit

Haroenv:
Replaced with two things:

- omit utility taken from babel, which is used when you spread destructure
- replacing the path version of Lodash' omit with more spreading & omit


packages/react-instantsearch-core/src/connectors/connectScrollTo.js
```diff
@@ -67,10 +66,10 @@ export default createConnector({
     // using ScrollTo in combination of Pagination. As pagination can be change
     // by every widget, we want to scroll only if it cames from the pagination
     // widget itself. We also remove the configure key from the search state to
-    // do this comparaison because for now configure values are not present in the
+    // do this comparison because for now configure values are not present in the
     // search state before a first refinement has been made and will false the results.
     // See: https://github.com/algolia/react-instantsearch/issues/164
-    const cleanedSearchState = omit(omit(searchState, 'configure'), id);
```
Haroenv: I don't know why this used to call `omit` twice before, it seemed useless there

packages/react-instantsearch-dom/src/components/Link.js
```diff
@@ -17,6 +16,6 @@ export default class Link extends Component {
   };
 
   render() {
-    return <a {...omit(this.props, 'onClick')} onClick={this.onClick} />;
+    return <a {...this.props} onClick={this.onClick} />;
```
Haroenv: here I just chose to go for overriding the prop, since there isn't much wrong with having that argument passed twice

packages/react-instantsearch-core/src/connectors/connectScrollTo.js
eunjae-lee: Right, like you did, `['configure', id]` will do the job.

samouss: We don't have to use `omit` we can leverage the spread operator. It might be easier to read with the function though.

PR #2467, opened 2019-05-17
 refactor(lodash): replace isEqual

Haroenv:
1. specified function for maps
2. `fast-deep-equal` for createConnector, because it is generic enough for our use case & in a hot path

IFW-735

### DIFF
--- a/packages/react-instantsearch-dom-maps/src/Connector.js
+++ b/packages/react-instantsearch-dom-maps/src/Connector.js
@@ -1,9 +1,25 @@
-import { isEqual } from 'lodash';
 import { Component } from 'react';
 import { polyfill } from 'react-lifecycles-compat';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-dom-maps/src/Connector.js
```diff
@@ -1,10 +1,26 @@
...
 import { LatLngPropType, BoundingBoxPropType } from './propTypes';
 
+function isEqualPosition(a, b) {
+  if (a === b) {
+    return true;
+  }
+  return a.lat === b.lat && a.lng === b.lng;
+}
+
+function isEqualCurrentRefinement(a, b) {
```
samouss: We have `fast-deep-equal` inside the package since it's already used in `core`. I would have leverage it rather than this two custom functions. The simpler, the better.

PR #2499, opened 2019-06-03
chore(examples): Update next.js to use React hooks

notrab:
**Summary**
This PR updates the Next.js example to use React Hooks. I end up doing this with every implementation of React InstantSearch + Next.js.

**Result**
Run the Next.js example and there should be no output differences. Significantly removes unnecessary boilerplate from the class based approach.

### DIFF
--- a/examples/next/pages/index.js
+++ b/examples/next/pages/index.js
@@ -1,6 +1,6 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import Router from 'next/router';
+import Router, { withRouter } from 'next/router';
 import qs from 'qs';
 import { Head, App, findResultsState } from '../components';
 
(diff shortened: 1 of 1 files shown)


samouss:
Thanks a lot for the suggestions! First of all sorry about the delayed answer the PR has been lost into my notification... 

We didn't make the switch to hooks inside the examples because we aim to address a large public. I agree that the release of hooks is not new but we don't have questions about their usage. We also want to be consistent across our usage of React. Which means that we have to make the change here but also inside all the other code samples that we have inside the doc, it takes time. It's not our top priorities at the moment. Once it will be, for sure we'll re-use your work!

Thanks again for the contribution!

francisrod01:
@samouss  In fact, we should get a hooks example to the community, since it's the most popular feature of React now.

@notrab Could you provide more information about how you solve it? I'm facing a problem with the Algolia's code for Next.js (I'm using it for hooks).

```Error serializing `.resultsState.state` returned from `getServerSideProps` in "/search". Reason: `object` ("[object Object]") cannot be serialized as JSON. Please only return JSON serializable data types.
```

PR #2512, opened 2019-06-05
fix(RangeInput): add form label

francoischalifour:
> This targets the `v6` branch because it can technically be a breaking change.

The `RangeInput` component was missing the form label `.ais-RangeInput-label` ([see `RangeInput` spec](https://instantsearch-css.netlify.com/widgets/range-input/)).

### DIFF
--- a/packages/react-instantsearch-dom/src/components/RangeInput.js
+++ b/packages/react-instantsearch-dom/src/components/RangeInput.js
@@ -97,29 +97,32 @@ export class RawRangeInput extends Component {
         className={classNames(cx('', !canRefine && '-noRefinement'), className)}
       >
         <form className={cx('form')} onSubmit={this.onSubmit}>
-          <input
-            className={cx('input', 'input--min')}
-            type="number"
-            min={min}
-            max={max}
-            value={from}
-            step={step}
-            placeholder={min}
-            disabled={!canRefine}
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-dom/src/components/RangeInput.js
```diff
@@ -97,29 +97,32 @@ export class RawRangeInput extends Component {
...
-            type="number"
-            min={min}
-            max={max}
-            value={to}
-            step={step}
-            placeholder={max}
-            disabled={!canRefine}
-            onChange={e => this.setState({ to: e.currentTarget.value })}
-          />
+          <label className={cx('label')}>
```
samouss: This is incorrect, the `label` should not wrap both inputs. We should have two labels: one for each input. I don't understand why we put a `label` around the input in the first place,  we don't provide that possibility to add a text.

> https://instantsearch-css.netlify.com/widgets/range-input/

francoischalifour: Do you think we should rather remove the label from the spec?

samouss: If it does not bring any value, yes.

francoischalifour: Right – closing this PR then.

PR #2531, opened 2019-06-10
chore(examples): update e-commerce demo

francoischalifour:
This updates the examples' webpack configuration to allow images and web manifests.


website/webpack.config.js
```diff
@@ -39,24 +39,60 @@ module.exports = {
...
         new HTMLWebpackPlugin({
           template: path.join(__dirname, example, 'index.html'),
           filename: path.join(outputPath, example, 'index.html'),
           chunks: [example],
         })
-    )
-    .concat([
-      new CopyWebpackPlugin([
-        { from: path.join(__dirname, 'assets'), to: 'assets/' },
-        path.join(__dirname, '_redirects'),
```
samouss: It looks like we forgot this line inside the change. The redirects don't work anymore.

> https://deploy-preview-2531--react-instantsearch.netlify.com/

website/webpack.config.js
```diff
@@ -39,24 +39,60 @@ module.exports = {
...
+              outputPath(_url, resourcePath, context) {
+                return path.relative(context, resourcePath);
+              },
+              name: '[name].[ext]',
+            },
+          },
+        ],
+      },
+      {
+        test: /\.webmanifest$/,
```
samouss: The images don't use the hash for the name, so what is the purpose of the loaders for the manifest? It doesn't work with a plain copy? Nothing seems to be dynamic inside the manifest.

francoischalifour: That was the first solution but we then changed that with @yannickcr.

I updated to use `CopyWebpackPlugin` which seems to be easier to use in this case.

website/webpack.config.js
francoischalifour: Put back in 816ec028ef46b7d54040d51f602e1f82da1f39a1.

website/webpack.config.js
```diff
@@ -88,6 +69,15 @@ module.exports = {
           chunks: [example],
         })
     ),
+    ...examples.map(
+      example =>
+        new CopyWebpackPlugin([
```
samouss: I don't think we have to call the plugin again, we can provide an array of paths as an argument.

francoischalifour: I tried that but it was a nightmare. Having a regex in the `from` and using it in the `to` doesn't work as expected.

samouss: Not sure to follow, we can't do this?

```js
[
  new CopyWebpackPlugin([
    ...examples.map(example => ({
      from: path.join(__dirname, example, "assets"),
      to: path.join(outputPath, example, "assets")
    })),
    {
      from: path.join(__dirname, "assets"),
      to: "assets/"
    },
    path.join(__dirname, "_redirects")
  ])
]
```

francoischalifour: Ah right! Yes, this works!

PR #2532, opened 2019-06-10
chore(examples): update e-commerce demo [part 2]

francoischalifour:
## Description

This updates the e-commerce demo with the new design.

## Notes

- `connectRange` was used for the `Ratings` widget
- [`react-compound-slider`](https://github.com/sghall/react-compound-slider) was used for the range slider
- The magnifier for the submit button in the search box of the `RefinementList` remains the one by default because React InstantSearch doesn't provide an easy way to customize it

## Preview

[See preview →](https://deploy-preview-2532--react-instantsearch.netlify.com/examples/e-commerce/)

## Related

- [E-commerce demo design](https://app.zeplin.io/project/5acc8becfc0c980068fc12cc/screen/5cd1a7992e4f02039b5a8336)


website/examples/e-commerce/App.js
```diff
@@ -4,212 +4,275 @@ import {
...
+    </header>
 
-const Facets = () => (
-  <aside>
-    <ClearRefinements
-      translations={{
-        reset: 'Clear all filters',
-      }}
+    <Configure
+      attributesToSnippet={['name:10', 'description:10']}
```
samouss: Why do we use snippet on the `name`?

website/examples/e-commerce/App.js
```diff
@@ -4,212 +4,275 @@ import {
...
-      <button className="btn btn-default">
-        <i className="fa fa-search" />
-      </button>
-    </span>
-  </div>
-);
+      <section className="container-results">
+        <header className="container-header container-options">
+          <SortBy
+            attribute="price"
```
samouss:
```suggestion

```

website/examples/e-commerce/App.js
```diff
@@ -4,212 +4,275 @@ import {
...
-      </div>
-    );
-  }
-});
+        </header>
 
-const ConnectedSearchBox = connectSearchBox(CustomSearchBox);
-const ConnectedHits = connectHits(CustomHits);
+        <Hits
+          hitComponent={({ hit }) => (
```
samouss: The component should not be inlined inside the render function. The reference of the function change on each render. It triggers an unmount/mount of the component on every render of the component.

website/examples/e-commerce/App.js
```diff
@@ -4,212 +4,275 @@ import {
...
+            <article className="hit">
+              <header className="hit-image-container">
+                <img src={hit.image} alt={hit.name} className="hit-image" />
+              </header>
+
+              <p className="hit-category">{hit.categories[0]}</p>
+              <h1>
+                <Snippet attribute="name" tagName="mark" hit={hit} />
```
samouss:
```suggestion
                <Highlight attribute="name" tagName="mark" hit={hit} />
```

website/examples/e-commerce/widgets/PriceSlider.js
```diff
@@ -0,0 +1,47 @@
+import React, { useState } from 'react';
+import { Range, createSliderWithTooltip } from 'rc-slider';
+import { connectRange } from 'react-instantsearch-dom';
+
+const RangeWithTooltip = createSliderWithTooltip(Range);
+
+const PriceSlider = ({ min, max, refine, currentRefinement, canRefine }) => {
```
samouss: The component does not work properly with the min & max value. The slider does not stick to the min & max when those values gets lower and bigger again.  Here is an example:

Untitled

website/examples/e-commerce/App.js
```diff
@@ -4,212 +4,275 @@ import {
...
-        </footer>
-      </div>
-    );
-  }
-});
+        </header>
 
-const ConnectedSearchBox = connectSearchBox(CustomSearchBox);
-const ConnectedHits = connectHits(CustomHits);
+        <Hits
```
samouss: We should use the `ScrollTo` component to reset the scroll position to the top of the list.

website/examples/e-commerce/widgets/PriceSlider.js
francoischalifour: Indeed, that comes from `rc-slider` triggering the [`onChange`](https://github.com/react-component/slider/#common-api) function whenever a value changes, even without being dragged. That's problematic.

francoischalifour:
@samouss A single request is triggered now that we don't inline the `Hit` component.

website/examples/e-commerce/Theme.css
```diff
@@ -0,0 +1,535 @@
...
+.ais-RefinementList-checkbox {
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  appearance: none;
+  background-color: rgba(65, 66, 71, 0.08);
+  border-radius: 2px;
+  height: 1rem;
+  margin: 0;
+  margin-right: 1rem;
+  outline: none;
```
samouss: I didin't realise that we removed the outline. Indeed it looks nicer without it, but it's a nightmare to navigate with the keyboard without this indication. We should keep it, but not the default color.

francoischalifour: That's something that I mentioned in the first PR in InstantSearch.js. There's a lot of accessibility improvements to do, but they should be shared across all the e-commerce demos so let's not treat them as blockers.

I'll ask Nico if he can include this in the next design additions he plans to do.

samouss: Fair enough I wasn’t sure that we already talked about it.

francoischalifour:
@samouss I tried another implementation with [`react-compound-slider`](https://github.com/sghall/react-compound-slider). From my manual tests, it works as expected. Let me know if you find any edge cases.

francoischalifour:
@samouss This is the default behavior that we provide in InstantSearch.js, you can reproduce on the [vanilla e-commerce demo](https://chore--examples-e-commerce--instantsearchjs.netlify.com/examples/e-commerce/).

PR #2535, opened 2019-06-10
fix(voiceSearch): fix incorrect status on stop

eunjae-lee:
**Summary**

At `stop()` of voiceSearchHelper, it calls `dispose()` and `resetState()`.
Because `dispose()` removes event listeners, `end` listener is not called.
So we're setting the `status` as `finished` here.
If we don't do it, it will be still `waiting` or `recognizing`.

Related: https://github.com/algolia/instantsearch.js/commit/84cc9f9cf01b6c550dace5091c8ca770a1d7afb1 from https://github.com/algolia/instantsearch.js/pull/3845


eunjae-lee:
I guess I made a mistake while managing this PR. So I cleaned up and force-pushed to get only one meaningful commit to be merged to master. The change is the same.

PR #2536, opened 2019-06-10
feat(ssr): only expose findResultsState (PART-1)

samouss:
This PR changes the exposed API of SSR. The previous implementation had some issues. The main one is that `react-instantsearch-dom/server` is used on both sides: client and server. The downside of this solution is that have a larger footprint for the client with the usage of SSR.

With this implementation, the code on the server never reaches the client. We use the exact same component if we are client-side only or both server & client side. With the previous implementation, we had two API exposed: `createInstantSearch` and `findResultsState`. We've removed the needs for the first one and only kept the second. The API remains the same, the only impact is that now two props are required: `searchClient` and `indexName`.

With the example, you can notice that we have some duplication inside the module `server` and `browser`. We have duplicated the declaration of `indexName` and `searchClient`. We can avoid this easily with a `createApp` function that returns the `App` and the `props` that we have to provide. I didn't put it inside the example because it's not a requirement to use the SSR. We'll see that in practice with the update of the SSR example in a later PR.

**Before**

```js
// server.js

import React from 'react';
import { renderToString } from "react-dom/server";
import { App, findResultsState } from "./App";

server.get("/", async (req, res) => {
  const initialState = {
    resultsState: await findResultsState(App)
  };

  const plainHTML = renderToString(<App {...initialState} />);

  // ...
});
```

```js
// browser.js

import React from 'react';
import { hydrate } from 'react-dom';
import { App } from './App';

hydrate(
  <App {...window.__APP_INITIAL_STATE__} />,
  document.getElementById('root')
);
```

```js
// App.js

import React, { Component } from "react";
import algoliasearch from "algoliasearch/lite";
import { createInstantSearch } from "react-instantsearch-dom/server";
import { SearchBox } from "react-instantsearch-dom";

const { InstantSearch, findResultsState } = createInstantSearch();

const searchClient = algoliasearch(
  "latency",
  "6be0576ff61c053d5f9a3225e2a90f76"
);

const App = ({ resultsState }) => (
  <InstantSearch
    searchClient={searchClient}
    indexName="instant_search"
    resultsState={resultsState}
  >
    <SearchBox />
  </InstantSearch>
);

export { App, findResultsState };
```

**After**

```js
// server.js

import { renderToString } from "react-dom/server";
import { findResultsState } from "react-instantsearch-dom/server";
import App from "./App";

const indexName = "instant_search";
const searchClient = algoliasearch(
  "latency",
  "6be0576ff61c053d5f9a3225e2a90f76"
);

server.get("/", async (req, res) => {
  const initialState = {
    resultsState: await findResultsState(App, {
      indexName,
      searchClient
    })
  };

  const plainHTML = renderToString(<App {...initialState} />);

  // ...
});
```

```js
// browser.js

import { hydrate } from "react-dom";
import { App } from "./App";

const indexName = "instant_search";
const searchClient = algoliasearch(
  "latency",
  "6be0576ff61c053d5f9a3225e2a90f76"
);

hydrate(
  <App
    {...window.__APP_INITIAL_STATE__}
    indexName={indexName}
    searchClient={searchClient}
  />,
  document.getElementById("root")
);
```

```js
// App.js

import React, { Component } from "react";
import { InstantSearch, SearchBox } from "react-instantsearch-dom";

const App = ({ indexName, searchClient, resultsState }) => (
  <InstantSearch
    searchClient={searchClient}
    indexName={indexName}
    resultsState={resultsState}
  >
    <SearchBox />
  </InstantSearch>
);

export { App };
```


packages/react-instantsearch-dom/src/server.js
```diff
@@ -1,3 +1 @@
-import createInstantSearchServer from './core/createInstantSearchServer';
-
-export const createInstantSearch = createInstantSearchServer;
+export { findResultsState } from './core/createInstantSearchServer';
```
francoischalifour: Is prefixing the method `findResultsState` with `find` a convention? I've always found it weird in this context.

samouss: Yes it's a convention for our SSR API. It's maybe a bit counter-intuitive but we have the same inside Vue InstantSearch. Until we have a dedicated API inside InstantSearch we should stay with this one IMO.

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -55,559 +56,451 @@ describe('createInstantSearchServer', () => {
...
-      expect(searchClient.addAlgoliaAgent).toHaveBeenCalledTimes(2);
-      expect(searchClient.addAlgoliaAgent).toHaveBeenCalledWith(
-        `react (${React.version})`
-      );
-      expect(searchClient.addAlgoliaAgent).toHaveBeenCalledWith(
-        `react-instantsearch (${version})`
-      );
-    });
+    const props = {
+      ...requiredProps,
```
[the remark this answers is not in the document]
samouss: https://github.com/algolia/react-instantsearch/pull/2536/commits/f408c9eddd0efb7728726412ee6b58086e292145

PR #2537, opened 2019-06-10
feat(ssr): hydrate client [PART-2]

samouss:
This PR introduces the possibility to hydrate the `searchClient` with the results of the request made on the server. It avoids to perform an initial request on the browser! At the moment the Algolia client does not have a proper API to populate the cache. The only solution is to create it by hand with the `.cache` attribute. It means that React InstantSearch shares the knowledge of the cache with the client for the `key` generation. For the moment it's fine, worst case we create a key that is useless. We will make sure to provide a proper API to hydrate the results for the next version fo the client.

**Before**

before

**After**

after


packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -257,6 +259,68 @@ export default function createInstantSearchManager({
     }
   }
 
+  function hydrateSearchClient(client, results) {
+    if (!results) {
+      return;
+    }
+
+    if (!client._useCache) {
```
francoischalifour: We need to make sure that the API client squad is aware that we rely on the `_useCache` property. This will likely change in JS client v4.

Wouldn't it be safer to rely on the existence of `addAlgoliaAgent`?

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -257,6 +259,68 @@ export default function createInstantSearchManager({
...
+      // This condition avoid hydrating a `searchClient` different from the
```
```suggestion
      // This condition avoids hydrating a `searchClient` different from the
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -257,6 +259,68 @@ export default function createInstantSearchManager({
...
+      // Algolia one. Unless the implementation uses the exact same approach
+      // for the cache (the property `_useCache`). The implementation is brittle
+      // but we don't have a proper way to detect the Algolia client at the moment.
+      return;
+    }
+
+    if (Array.isArray(results)) {
+      hydrateSearchClientWithMultiIndexRequest(client, results);
+      return;
```
francoischalifour: In that case we could use `if/else` because it's be easy to miss some instructions since the multi-index path ends here.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -257,6 +259,68 @@ export default function createInstantSearchManager({
...
+    }
+
+    hydrateSearchClientWithSingleIndexRequest(client, results);
+  }
+
+  function hydrateSearchClientWithMultiIndexRequest(client, results) {
+    // https://github.com/algolia/algoliasearch-client-javascript/blob/c27e89ff92b2a854ae6f40dc524bffe0f0cbc169/src/AlgoliaSearchCore.js#L232-L240
```
francoischalifour: Are you referring to the way we need to differentiate the POST requests with this comment? It could be clearer with prior explanations.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
samouss: Yes it will definitely change inside the V4. We'll ask to create an API to correctly hydrate the cache of the client, rather than those hacks. We could do both.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
samouss: Nope, it's referring to the computation of the cache key inside the client.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
samouss: It  depends a lot on your habits, I don't have a strong opinion about that `if/else` is equivalent.

francoischalifour: I meant that the early return is quite misleading here since you don't necessarily want to end the path here, you rather don't want to call `hydrateSearchClientWithSingleIndexRequest`.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
francoischalifour: Right, an explanation would help to understand!

francoischalifour: Right, an explanation would help to understand!

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
samouss: Actually I do want to stop the path at this point.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
samouss: https://github.com/algolia/react-instantsearch/pull/2537/commits/f622ed82e49e4a77e964ee68e793afd83e0dbcd8

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
samouss: https://github.com/algolia/react-instantsearch/pull/2537/commits/a96de76af0d50a184308bc661e16303046d2bed7

PR #2178, opened 2019-03-18
feat(context): migrate core [STEP 1]

Haroenv:
This includes:

1. change to createConnector
2. change to createStore
3. skip failing tests
4. small nits found while coding

- [x] **STEP 1**: migrate createConnector (#2178, this PR)
- [x] **STEP 2**: first PR with connectors (searchbox, configure, hits, pagination) (#2179)
- [x] **STEP 3**: other connectors (#2180)
- [x] **STEP 4**: Multi index core (#2181)
- [x] **STEP 5**: Multi index connectors (#2185)
- [x] **STEP 6**: SSR (#2192)
- [ ] **STEP 7**: TypeScript completion (#2189)
- [x] **STEP 8**: documentation change (#2190)

This can be reviewed on its own, but not released (obviously)

PR #2570, opened 2019-06-20
 fix(createInstantSearchManager): fix multi-index search 

yannickcr:
**Summary**

The multi-index support was broken since the migration to the new React context (https://github.com/algolia/react-instantsearch/pull/2178).

This issue is due to multiple small changes:

* The context is now passed though the widget props but some functions in `createInstantSearchManager.js` were still trying to read it from `widget.context`.
* The `indexName` prop is no longer mapped to `indexId` in the `Index` widget if `indexId` is not defined, so the `Index` widget detection was not working correctly anymore.
* The widgets are now registered during the `componentDidMount` (before the change they were registered in the `constructor`), so the registration order is different. And since we're using a `reduce` on the widget list to calculate the `SearchParameters` the end result was different.

This PR makes the following fixes:

* The functions that were relying on on `widget.context` were updated to use `widget.props` instead.
* We re-added the mapping between `indexId` and `indexName`.
* Added a new step during the `mainParameters` and `derivedIndices` calculations to sort the `Index` widgets first so the following `reduce` result is correct.

**Result**

All Storybook `<Index>` stories are now working.
I'm not sure how to test this using unit tests (the existing tests are a bit odd as we register some "dummy" widgets and use them to calculate the `mainParameters` and `derivedParameters`).

Any help to reproduce storybook cases as real tests would be very helpful here :wink:

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
@@ -656,45 +656,41 @@ describe('createInstantSearchManager', () => {
         })
       );
 
-      expect(derivedParameters).toHaveLength(5);
-
-      expect(derivedParameters).toEqual(
-        expect.arrayContaining([
-          expect.objectContaining({
-            indexId: 'bestbuy',
-            parameters: expect.objectContaining({
-              index: 'bestbuy',
-            }),
+      expect(derivedParameters).toEqual([
+        expect.objectContaining({
+          indexId: 'bestbuy',
+          parameters: expect.objectContaining({
+            index: 'bestbuy',
           }),
-          expect.objectContaining({
-            indexId: 'instant_search',
-            parameters: expect.objectContaining({
-              index: 'instant_search',
-            }),
+        }),
+        expect.objectContaining({
+          indexId: 'instant_search',
+          parameters: expect.objectContaining({
+            index: 'instant_search',
           }),
-          expect.objectContaining({
-            indexId: 'instant_search_apple',
-            parameters: expect.objectContaining({
-              index: 'instant_search',
-              filters: 'brand:Apple',
-            }),
+        }),
+        expect.objectContaining({
+          indexId: 'instant_search_apple',
+          parameters: expect.objectContaining({
+            index: 'instant_search',
+            filters: 'brand:Apple',
           }),
-          expect.objectContaining({
-            indexId: 'instant_search_samsung',
-            parameters: expect.objectContaining({
-              index: 'instant_search',
-              filters: 'brand:Samsung',
-            }),
+        }),
+        expect.objectContaining({
+          indexId: 'instant_search_samsung',
+          parameters: expect.objectContaining({
+            index: 'instant_search',
+            filters: 'brand:Samsung',
           }),
-          expect.objectContaining({
-            indexId: 'instant_search_microsoft',
-            parameters: expect.objectContaining({
-              index: 'instant_search',
-              filters: 'brand:Microsoft',
-            }),
+        }),
+        expect.objectContaining({
+          indexId: 'instant_search_microsoft',
+          parameters: expect.objectContaining({
+            index: 'instant_search',
+            filters: 'brand:Microsoft',
           }),
-        ])
-      );
+        }),
+      ]);
     });
 
     it('expects widgets main parameters and derived parameters to be correctly calculated with SortBy within a multi index context', () => {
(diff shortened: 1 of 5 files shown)


packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -13,16 +13,31 @@ function addAlgoliaAgents(searchClient) {
...
 const isTargetedIndexEqualIndex = (widget, indexId) =>
-  widget.context.multiIndexContext.targetedIndex === indexId;
+  widget.props.indexContextValue.targetedIndex === indexId;
 
 // Relying on the `indexId` is a bit brittle to detect the `Index` widget.
 // Since it's a class we could rely on `instanceof` or similar. We never
 // had an issue though. Works for now.
-const isIndexWidget = widget => Boolean(widget.props.indexId);
+const isIndexWidget = widget =>
+  Boolean(widget.props.indexId || widget.props.indexName);
```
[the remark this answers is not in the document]
yannickcr: You're right :+1: , made it in https://github.com/algolia/react-instantsearch/pull/2570/commits/9330e8a6182ddbd2933fcc62b99ee9e0890282e7 and it feels cleaner that way.

packages/react-instantsearch-core/src/widgets/Index.tsx
```diff
@@ -125,7 +125,9 @@ class Index extends Component<InnerProps, State> {
 
 const IndexWrapper: React.FC<Props> = props => (
   <InstantSearchConsumer>
-    {contextValue => <Index contextValue={contextValue} {...props} />}
+    {contextValue => (
+      <Index contextValue={contextValue} indexId={props.indexName} {...props} />
```
Haroenv: This automatic overriding seems logical to me, but I was still confused seeing this written first, what about something like this:

```ts
const IndexWrapper: React.FC<Props> = props => {
  const inferredIndexId = props.indexName;
  return (
    <InstantSearchConsumer>
      {contextValue => (
        <Index contextValue={contextValue} indexId={inferredIndexId} {...props} />
      )}
    </InstantSearchConsumer>
  );
};
```

This makes it somewhat clearer to me it gets overriden

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -613,6 +624,134 @@ describe('createInstantSearchManager', () => {
...
+      expect(mainParameters).toEqual(
+        expect.objectContaining({
+          index: 'index1',
+        })
+      );
+
+      expect(derivedParameters).toHaveLength(5);
+
+      expect(derivedParameters).toEqual(
+        expect.arrayContaining([
```
samouss: I don't think this part is useful, we can directly use `expect.objectContaining`.

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -613,6 +624,134 @@ describe('createInstantSearchManager', () => {
...
+        derivedParameters,
+      } = wrapper.instance().aisManager.getSearchParameters();
+
+      expect(mainParameters).toEqual(
+        expect.objectContaining({
+          index: 'index1',
+        })
+      );
+
+      expect(derivedParameters).toHaveLength(5);
```
samouss: We don't have to assert the length since we strictly assert its equality below.

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -613,6 +624,134 @@ describe('createInstantSearchManager', () => {
...
       ]);
     });
+
+    it('expects widgets main parameters and derived parameters to be correctly calculated within a multi index context', () => {
+      const wrapper = mount(
+        <InstantSearch indexName="index1" searchClient={createSearchClient()}>
+          <Index indexName="bestbuy" />
+          <Index indexName="instant_search" />
+
+          <Index indexId="instant_search_apple" indexName="instant_search">
```
samouss: This is way simpler to test with this!

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -615,6 +624,130 @@ describe('createInstantSearchManager', () => {
...
+      expect(mainParameters).toEqual(
+        expect.objectContaining({
+          index: 'index1',
+        })
+      );
+
+      expect(derivedParameters).toHaveLength(2);
+
+      expect(derivedParameters).toEqual(
+        expect.arrayContaining([
```
samouss: > https://github.com/algolia/react-instantsearch/pull/2570/commits/8f46d590f8a08948894b90b8f7d6dc76b2934118

It was not the only place where it was used.

yannickcr: Oups, forgot the other one.

PR #2591, opened 2019-06-25
chore(tools): replace tslint by typescript-eslint

yannickcr:
**Summary**

(Same as https://github.com/algolia/instantsearch.js/pull/3711)

TypeScript support was added to `eslint-config-algolia@13.3.0` ([changelog](https://github.com/algolia/eslint-config-algolia/blob/master/CHANGELOG.md#1330-2019-04-16)), allowing us to drop TSLint and use ESLint to lint both JavaScript and TypeScript files.

* Enable `algolia/typescript` configuration
* Remove TSLint
* Update `lint` and `lint:fix` npm scripts to only use ESLint on JavaScript and TypeScript files
* Update files to match new rules:
  * Update ignore comments to target the new ESLint rules
  * Update naming of generic type variables (new `@typescript-eslint/generic-type-naming` rule)
  * Fix errors in TypeScript files (not validated by ESLint before)

**Result**

Tooling only, no changes are visible to the users.
For developers the development environment is simplified with one less tool to setup.


packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
```diff
@@ -91,7 +95,7 @@ class VoiceSearch extends Component<VoiceSearchProps, VoiceListeningState> {
     this.setState(this.voiceSearchHelper.getState());
   }
 
-  public render() {
+  public render(): React.ReactNode {
```
francoischalifour: This is quite verbose and I've got the feeling this is gonna create some friction quickly.

packages/react-instantsearch-core/src/connectors/__tests__/connectHitInsights.ts
```diff
@@ -1,11 +1,27 @@
...
+function setup(): {
+  insightsClient: jest.Mock<any>;
+  props: {
+    insights: InsightsClient;
+  };
+} {
   const insightsClient = jest.fn();
 
-  const createMultiIndexContext = () => ({
+  const createMultiIndexContext = (): MultiIndexContext => ({
```
Haroenv: I see no value in having such precise types in tests, because it's much more work to test something which doesn't pass the types, wdyt?

packages/react-instantsearch-core/src/connectors/connectQueryRules.ts
```diff
@@ -82,7 +82,7 @@ function getRuleContextsFromTrackedFilters({
 }: {
   searchState: SearchState;
   trackedFilters: QueryRulesProps['trackedFilters'];
-}) {
+}): string[] {
```
Haroenv: still think this rule can easily break inferred types 🤔

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
yannickcr: If everyone agree on this we could remove the rule from the ESLint config, WDYT?

packages/react-instantsearch-core/src/connectors/__tests__/connectHitInsights.ts
yannickcr: Agree.

But if we remove the `@typescript-eslint/explicit-function-return-type` it should not be an issue anymore :innocent:

packages/react-instantsearch-dom/src/components/VoiceSearch.tsx
yannickcr: Updated to the new eslint-config-algolia release without this rule https://github.com/algolia/react-instantsearch/pull/2591/commits/2ab2fd3414eccfc1ba26e5c0a5329c3a256ddac6

Haroenv:
Screenshot 2019-07-03 at 11 27 23

If the languages overflow, the design breaks for the other elements too, probably because it's a grid?

PR #2600, opened 2019-06-27
chore(lodash): remove dependencies

Haroenv:
- lodash from packages
- babel-plugin-lodash from root
- swap out isEqual from connectInfiniteHits

IFW-778

closes #2468

**Summary**

**Result**


Haroenv:
the error in examples is in v6 as well, an artefact of my rebase I assume, but I haven't found why yet

PR #2626, opened 2019-07-03
fix(peerDependencies): update React

Haroenv:
We are now relying on React 16.3 features like `static getDerivedStateFromProps` & `createContext` in the library, which thus won't work with lower versions.

The lowest version of React Native which requires React 16.3 (albeit an alpha) is 0.54.0 (released 5th of March 2018).

### DIFF
--- a/packages/react-instantsearch-native/package.json
+++ b/packages/react-instantsearch-native/package.json
@@ -43,7 +43,7 @@
     "react-instantsearch-core": "^5.7.0"
   },
   "peerDependencies": {
-    "react": ">= 15.3.0 < 17",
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch-native/package.json
```diff
@@ -43,7 +43,7 @@
     "react-instantsearch-core": "^5.7.0"
```
Haroenv: this should become 6 I guess but we haven't released that yet

samouss: Yep those are managed by the release script via `lerna`.

PR #2621, opened 2019-07-02
refactor(helper): replace Helper usage

yannickcr:
**Summary**

A new `algoliasearch-helper` major version will be released with a few breaking changes ([Migration guide](https://gist.github.com/Haroenv/87b32d4c6dca5deb940ba030bd7a38d8)).

This PR updates `algoliasearch-helper` dependency to use the new major version (released as canary for now) and fix `react-instantsearch` (and its tests) to handle the changes:

* `getQueryParameter` removal (https://github.com/algolia/algoliasearch-helper-js/pull/713)
* Default values removal (https://github.com/algolia/algoliasearch-helper-js/pull/670)
* Emit object instead of an argument list (https://github.com/algolia/algoliasearch-helper-js/pull/699, https://github.com/algolia/algoliasearch-helper-js/pull/700)

PR #2627, opened 2019-07-03
docs(MIGRATION): migrate to v6

samouss:


MIGRATION.md
```diff
@@ -0,0 +1,126 @@
+## Upgrade to 6.x.x
+
+This new major version of React InstantSearch leverages the [`Context`](https://reactjs.org/docs/context.html) API that has been released part of React 16.3.0. It means that now the **minimal required version for React InstantSearch is React 16.3.0**. We took this opportunity to make React InstantSearch compliant with the [`Strict`](https://reactjs.org/docs/strict-mode.html) mode too. Prior to this version, the lib was not working with those constraints. Here is the list of APIs that have breaking changes:
```
Haroenv:
```suggestion
This new major version of React InstantSearch leverages the [`Context`](https://reactjs.org/docs/context.html) API that has been released part of React 16.3.0. It means that now the **minimal required version for React InstantSearch is React 16.3.0**. We took this opportunity to make React InstantSearch compliant with the [`Strict`](https://reactjs.org/docs/strict-mode.html) mode too. Prior to this version, the library was not working with those constraints. Here is the list of APIs that have breaking changes:
```

MIGRATION.md
```diff
@@ -0,0 +1,126 @@
...
+-export { App, findResultsState };
++export { App };
+```
+
+You can notice that we have some duplication inside the module `server` and `browser`. We have duplicated the declaration of `indexName` and `searchClient`. We can avoid this easily with a `createApp` function that returns the `App` and the `props` that we have to provide. You can find this pattern applied inside the complete example on [GitHub](https://github.com/algolia/react-instantsearch/tree/fbc89aa7a7c02e572081444ab4b5039e7b1df1a9/examples/server-side-rendering).
+
+### Remove the support for `appId` & `apiKey`
+
+We've removed the support for `appId` & `apiKey` on `InstantSearch`. Those props were not documented anymore but they were still supported. The preferred solution is to use the prop [`searchClient`](https://www.algolia.com/doc/api-reference/widgets/instantsearch/react/#widget-param-searchclient) with the package the [`algoliasearch`](https://www.algolia.com/doc/api-client/getting-started/install/javascript/). The benefit of this change is that we don't ship `algoliasearch` anymore with React InstantSearch. When the `searchClient` used is not the Algolia one you don't pay the cost of it anymore.
+
```
Haroenv: Let's put a code sample here: 

```jsx
import algoliasearch from 'algoliasearch/lite';
import { InstantSearch } from 'react-instantsearch-dom';

const searchClient = algoliasearch(
  "latency",
  "6be0576ff61c053d5f9a3225e2a90f76"
);

const App = ({ children }) => (
  <InstantSearch searchClient={searchClient>
    {children}
  </InstantSearch>
);
```

Haroenv:
Should we mention lodash too? or in the next beta?

MIGRATION.md
```diff
@@ -0,0 +1,126 @@
+## Upgrade to 6.x.x
+
+This new major version of React InstantSearch leverages the [`Context`](https://reactjs.org/docs/context.html) API that has been released part of React 16.3.0. It means that now the **minimal required version for React InstantSearch is React 16.3.0**. We took this opportunity to make React InstantSearch compliant with the [`Strict`](https://reactjs.org/docs/strict-mode.html) mode too. Prior to this version, the lib was not working with those constraints. Here is the list of APIs that have breaking changes:
+
+### SSR
+
+We have made changes to the SSR API to only use the `server` module, well on the server. Previously it was used on both sides: client and server. Those changes in the API also simplifies the flow of information and makes the SSR simpler to reason about. The neat addition that we've made to the API is the `searchClient` hydration. The request was done on the server, but also on the client when the App is mounted. Since the state is passed from the server to the client the initial request on the client was useless. We're now able to avoid this initial request by re-using the results retrieved on the server.
```
francoischalifour:
```suggestion
We made changes to the SSR API to use the `server` module only on the server. It was previously used on both client and server. Those changes in the API also simplify the flow of information and makes the SSR simpler to reason about. The neat addition that we've made to the API is the `searchClient` hydration. The request was done on the server, but also on the client when the app was mounted. Since the state is passed from the server to the client, the initial request on the client was useless. We're now able to avoid this initial request by re-using the results retrieved on the server.
```

MIGRATION.md
```diff
@@ -0,0 +1,126 @@
...
+  </InstantSearch>
+);
+
+-export { App, findResultsState };
++export { App };
+```
+
+You can notice that we have some duplication inside the module `server` and `browser`. We have duplicated the declaration of `indexName` and `searchClient`. We can avoid this easily with a `createApp` function that returns the `App` and the `props` that we have to provide. You can find this pattern applied inside the complete example on [GitHub](https://github.com/algolia/react-instantsearch/tree/fbc89aa7a7c02e572081444ab4b5039e7b1df1a9/examples/server-side-rendering).
+
+### Remove the support for `appId` & `apiKey`
```
francoischalifour:
```suggestion
### Remove support for `appId` and `apiKey`
```

MIGRATION.md
```diff
@@ -0,0 +1,126 @@
...
+
+We've removed the support for `appId` & `apiKey` on `InstantSearch`. Those props were not documented anymore but they were still supported. The preferred solution is to use the prop [`searchClient`](https://www.algolia.com/doc/api-reference/widgets/instantsearch/react/#widget-param-searchclient) with the package the [`algoliasearch`](https://www.algolia.com/doc/api-client/getting-started/install/javascript/). The benefit of this change is that we don't ship `algoliasearch` anymore with React InstantSearch. When the `searchClient` used is not the Algolia one you don't pay the cost of it anymore.
+
+### Remove the support for `root` in `InstantSearch` & `Index`
```
francoischalifour:
```suggestion
### Remove support for `root` in `InstantSearch` and `Index`
```

Haroenv:
New section:

```markdown
### algoliasearch-helper v3

In this major release we also migrate to the next major release of the Algolia helper. This means if at any point you interact with a `SearchResults` or `SearchParameters` object, the breaking changes there will also take effect here. For use in React InstantSearch this should be only when creating a custom connector using `createConnector`, 

There will no longer be the default value of page (0) or query ('') in `SearchParameters`. Both of those will now be undefined if no value is given.

The function `getQueryParameter` has been removed in favor of simply calling the getter for that parameter.
```

see #2621

MIGRATION.md
samouss: https://github.com/algolia/react-instantsearch/pull/2627/commits/3a39052da5e98ccd5494dbbce1e7fe62277db5f1

PR #2644, opened 2019-07-09
fix(geo): remove lifecycle compat

Haroenv:
This is no longer necessary, since we now require React 16.3 as a peer dependency (#2626)

Result: smaller bundle size :)


Haroenv:
it's still in the lock file for the examples which use @reach/router etc.

Haroenv:
```

PR #2687, opened 2019-07-19
feat(autocomplete): add queryID & position to provided hits

Haroenv:
This was accidentally forgotten when we added the queryID to hits & infinite hits.

The PR is made to `next`, because I was already on that branch, can also make a PR to develop if that's what we want.

IFW-902

**Summary**

you need to be able to retrieve the query ID & absolute position if you are implementing insights. However, you need to nest a StateResults inside connectAutoComplete if you'd want to do this now because the results object isn't exposed.

Absolute position is less important in autocomplete, because you're usually on page 0, but you never know!

**Result**

queryID is injected like in Hits & InfiniteHits

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectAutoComplete.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectAutoComplete.js
@@ -13,11 +13,11 @@ describe('connectAutoComplete', () => {
         { contextValue },
         {},
         {
-          results: { hits },
+          results: { hits, page: 0, hitsPerPage: 20 },
         }
       );
       expect(props).toEqual({
-        hits,
+        hits: [{ __position: 1 }],
         currentRefinement: '',
       });
 
@@ -25,11 +25,11 @@ describe('connectAutoComplete', () => {
         { contextValue },
         { query: 'query' },
         {
-          results: { hits },
+          results: { hits, page: 0, hitsPerPage: 20 },
         }
       );
       expect(props).toEqual({
-        hits,
+        hits: [{ __position: 1 }],
         currentRefinement: 'query',
       });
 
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-core/src/connectors/connectAutoComplete.js
```diff
@@ -29,14 +30,25 @@ function getHits(searchResults) {
...
+        searchResults.results.page
+      );
     } else {
       return Object.keys(searchResults.results).reduce(
         (hits, index) => [
           ...hits,
           {
             index,
-            hits: searchResults.results[index].hits,
+            hits: addAbsolutePositions(
```
tkrugg: I think they make be more readable of it was split in 
```js
let hits = searchResults.results[index].hits
hits = addQueryID(hits)
hits = addAbsolutePositions(hits)
```

Haroenv: could be, but then I would need to change the function to a "block arrow", and didn't feel like that was very necessary here

packages/react-instantsearch-core/src/connectors/__tests__/connectAutoComplete.js
```diff
@@ -13,38 +13,79 @@ describe('connectAutoComplete', () => {
...
         }
       );
       expect(props).toEqual({
-        hits,
+        hits: [{ __position: 1 }],
         currentRefinement: 'query',
       });
     });
+
+    it('provides current hits to the component with extra info', () => {
```
samouss: We could be more descriptive for the test name.

Haroenv:
```suggestion
    it('provides current hits to the component with queryID & position', () => {
```

PR #2701, opened 2019-07-22
fix(connectToggleRefinement): cast currentRefinement to boolean

### LINKED ISSUE #2689: ToggleRefinement not working with Routing URLs example
**Describe the bug 🐛**

When using the [ToggleRefinement](https://www.algolia.com/doc/api-reference/widgets/toggle-refinement/react/) widget with the routing enabled (following the [Routing URLs
 guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/react/) in the documentation) then the toggle parameter is passed as a `string` to the widget instead of a `boolean`.

We got this warning from React:

```js
Warning: Failed prop type: Invalid prop `currentRefinement` of type `string` supplied to `ToggleRefinement`, expected `boolean`.
```

And when refreshing the page the checkbox is always checked (since we receive the `"false"` string which is truthy).

This is because the [qs module](https://github.com/ljharb/qs) parses the url parameters as string by default (https://github.com/ljharb/qs/issues/91). You can fix this by using `qs.decoder` (https://github.com/ljharb/qs/issues/91#issuecomment-437926409) but this is not very convenient (and not precised in the documentation).

**To Reproduce 🔍**

1. Go to https://codesandbox.io/s/example-react-router-b3u4m
2. Click on 'Free shipping'
3. See warning in console.

**Expected behavior 💭**

The `toggle` parameter should be correctly passed to the widget as a `boolean`.

**Environment:**

 - OS: Linux
 - Browser: Chrome
 - Version: 75

yannickcr:
**Summary**

Fix #2689 by always casting `currentRefinement` as a `boolean` before passing it to the component.

Some strings are considered as "falsy" (`0`, `false`, `null`, `undefined`) to ensure compatibility with `qs` in the existing [Routing URL guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/react/).


packages/react-instantsearch-core/src/connectors/connectToggleRefinement.js
```diff
@@ -24,10 +26,11 @@ function getCurrentRefinement(props, searchState, context) {
     false
   );
 
-  if (currentRefinement) {
-    return currentRefinement;
+  if (falsyStrings.indexOf(currentRefinement) !== -1) {
```
samouss: We could have wrote the condition on the `'true'` that would include all falsy values.

```js
if (typeof currentRefinement === 'string') {
  return currentRefinement === 'true';
}
```

PR #2722, opened 2019-07-30
fix(core): searchState can be non-Object object

### LINKED ISSUE #2568: searchState.hasOwnProperty is not a function due to Object.create(null)
**Describe the bug 🐛**

I'm using the npm module [`query-string`](https://www.npmjs.com/package/query-string) to parse search terms previously encoded into URLs. The `parse` method this module exposes creates objects using `Object.create(null)`, which means they don't have a prototype.

If I use the `query-string` module to parse an Algolia searchState from the URL, and then pass that searchState to the `InstantSearch` component, I get an error:

`Uncaught TypeError: searchState.hasOwnProperty is not a function`

This is coming from react-instantsearch-core, `src/core/indexUtils.js`, `hasRefinements` function. The error occurs because the `searchState` I'm passing has no prototype, and so it has no `hasOwnProperty` function.

**To Reproduce 🔍**

Pass any `searchState` object created with `Object.create(null)` to `InstantSearch`.

**Expected behavior 💭**

No error.

**Environment:**

 - macOS 10.14.5
 - Chrome 74

**Additional context**

Consider replacing the current calls to `searchState.hasOwnProperty(x)` with `{}.hasOwnProperty.call(searchState, x)`

Haroenv:
If someone calls `Object.create(null)` and then adds the expected properties, `hasOwnProperty` will not be available on the object itself. 

fixes #2568 

TODO:

- [x] add a test for this behaviour with `Object.create(null)`

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/indexUtils.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/indexUtils.js
@@ -96,68 +96,115 @@ describe('utility method for manipulating the search state', () => {
       });
     });
 
-    it('retrieve the current refinement value', () => {
-      const searchState = {
-        page: 1,
-        last: 'last',
-        refinement: 'refinement',
-        another: 'another',
-        namespace: {
+    describe('getCurrentRefinementValue', () => {
+      it('retrieves the current refinement value', () => {
+        const searchState = {
+          page: 1,
+          last: 'last',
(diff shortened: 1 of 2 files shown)


Haroenv:
One thing to note is that I made the PR to v6, but maybe it could be made on v5 as well. Same for me tbh

PR #2726, opened 2019-08-01
fix(ssr): avoid duplicate serializing

Haroenv:
Response of `findResultsState` (and `resultsState` prop) before:

```js
const resultsState = {
  content: SearchResults(state: searchParameters, results: rawResults),
  state: searchParameters,
  _originalResponse: {
    results: rawResults
  },
};
```

Note that `SearchResults` has the properties `_rawResults` & `_state`. This means that when we serialize, we are retrieving the original response and state twice.

Then on further searching, I also find that we don't need to serialize the `SearchResults` object at all, since only the raw response and state is used.

I modified this to be more consistent and no longer have duplication:

```js
const resultsState = {
  state: searchParameters,
  rawResults,
};
```

The `SearchParameters` object gets serialized without issue, as before.

This change in shape has of course also an impact on the shape of `resultsState`, the prop, but that is fine, since it was previously typed as `any`.

To be able to type it correctly, I had to add `@types/algoliasearch` & update `algoliasearch-helper` for typescript though.

This was initially noticed by @cloakedninjas, thanks!

(this can also be done in user land by manually dropping the references, and recreating the expected shape, but this is better, since it needs no user work)

### DIFF
--- a/packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
+++ b/packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
@@ -3,14 +3,13 @@ import PropTypes from 'prop-types';
 import createInstantSearchManager from '../core/createInstantSearchManager';
 import { InstantSearchProvider, InstantSearchContext } from '../core/context';
 import { Store } from '../core/createStore';
+import { PlainSearchParameters, SearchParameters } from 'algoliasearch-helper';
+import { MultiResponse } from 'algoliasearch';
 
-function isControlled(props: Props) {
-  return Boolean(props.searchState);
-}
-
-// @TODO: move this to the helper?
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -53,13 +52,17 @@ type Props = {
...
   stalledSearchDelay?: number;
-  resultsState: SearchResults | { [indexId: string]: SearchResults };
+  resultsState: ResultsState | { [indexId: string]: ResultsState };
 };
 
 type State = {
   contextValue: InstantSearchContext;
 };
 
+function isControlled(props: Props) {
```
Haroenv: TS complained about "use before declare", but not sure why it only now complained. It's not that big of a deal though.

PR #2811, opened 2019-09-01
Adds a missing peer dependency

arcanis:
**Summary**

The `algoliasearch-helpers` package has a peer dependency on `algoliasearch`, but it's not provided by `react-instantsearch-core` and thus generates a warning.

Ref https://github.com/yarnpkg/berry/issues/418

**Result**

One less warning 🙂

### DIFF
--- a/packages/react-instantsearch-core/package.json
+++ b/packages/react-instantsearch-core/package.json
@@ -44,6 +44,7 @@
     "prop-types": "^15.5.10"
   },
   "peerDependencies": {
+    "algoliasearch": ">= 3.1 < 4",
     "react": ">= 15.3.0 < 17"
   }
 }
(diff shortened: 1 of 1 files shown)


Haroenv:
Hmm, this is true, but it’s actually optional to use this package (you can use any API compatible object). Maybe let’s make it an optional peer dependency, or remove the peer dependency in helper :)

arcanis:
Now optional ✨

arcanis:
Oh you're right, I don't see any file that actually uses `algoliasearch` in the helper apart from the test ones 🤔

arcanis:
Closing in favour of https://github.com/algolia/algoliasearch-helper-js/pull/746

PR #2823, opened 2019-09-10
chore(examples): update examples to match the ones on instantsearch.js

yannickcr:
**Summary**

This PR updates the examples to match the ones from [`instantsearch.js`](https://github.com/algolia/instantsearch.js) to allow us to run the same [End-2-End test suite](https://github.com/algolia/instantsearch-e2e-tests) on them.

The changes are:

* Moving examples from `website/examples` to `examples`.
* Examples and Storybook that were compiled in`website/dist` are now compiled in `website`.
* Fix e-commerce demo for IE11 (polyfills).
* Update e-commerce routing configuration to match the one from `instantsearch.js` (most of it is just copy/pasted from https://github.com/algolia/instantsearch.js/blob/17c70e9520349e9a2f5985dbe1320b14dd6109ef/examples/e-commerce/src/routing.ts).
* Update all examples to use `react-instantsearch` build from the current branch.


Haroenv:
Will they still be available from the same url?

yannickcr:
Yes (https://github.com/algolia/react-instantsearch/pull/2823/commits/f50eb14d6d74a4477ac121055cc2b0753d0120ce :innocent: )

Haroenv:
can you make the build --quiet or something else which gives less output on netlify? 

A certain command failed with `2`, but not sure from the [output](https://app.netlify.com/sites/react-instantsearch/deploys/5d77cd1c7728b8000845d282) which one.

examples/autocomplete/package.json
```diff
@@ -19,7 +19,7 @@
     "react": "16.8.6",
     "react-autosuggest": "9.4.3",
     "react-dom": "16.8.6",
-    "react-instantsearch-dom": "5.7.0"
```
Haroenv: this means the example no longer works on codesandbox and I use this quite frequently; can we make this change only for the ecommerce demo?

examples/e-commerce/package.json
```diff
@@ -8,7 +8,6 @@
     "qs": "6.8.0",
     "react": "16.8.6",
     "react-compound-slider": "2.2.0",
-    "react-dom": "16.8.6",
-    "react-instantsearch-dom": "5.7.0"
```
Haroenv: how is this imported now then?

examples/autocomplete/package.json
yannickcr: We updated all examples on `instantsearch.js` so I made the same here.

But I see no issue to only do it on the e-commerce demo if it's better for you.

examples/e-commerce/package.json
yannickcr: It is using an alias in the webpack configuration https://github.com/algolia/react-instantsearch/pull/2823/files#diff-6420f9faefa3a9a37b0b32aa2332d22fR28-R39

examples/autocomplete/package.json
Haroenv: yes, especially the server side rendering ones for example these are used for, thanks!

yannickcr: Done, but I did not touch the other "real" demos (default-theme, media, tourism) since they are all using the same webpack configuration as e-commerce. Is it ok for you?

examples/tourism/package.json
```diff
@@ -1,12 +1,12 @@
...
-  "name": "media",
+  "name": "tourism",
   "version": "1.0.0",
   "private": true,
   "dependencies": {
     "algoliasearch": "3.33.0",
     "qs": "6.8.0",
     "react": "16.8.6",
     "react-dom": "16.8.6",
-    "react-instantsearch-dom": "5.7.0"
```
Haroenv: diff is going a big weird here, but seems correct

yannickcr: yeah, git mixed tourism and media `package.json` files :neutral_face:

PR #2824, opened 2019-09-11
test: add End-2-End test suite

yannickcr:
**Summary**

Same as https://github.com/algolia/instantsearch.js/pull/4035 but for `react-instantsearch`.

This PR makes the following changes:
* Add `algolia/instantsearch-e2-tests` as devDependency.
* Add `algolia/instantsearch-e2-tests` peerDependencies.
* Add 2 WebdriverIO configurations: one to run tests locally and one to run them on Sauce Labs.
* Add `.env` file to `.gitignore` so you can use it to store Sauce Labs credentials (see https://github.com/algolia/instantsearch-e2e-tests#credentials ).
* Add npm scripts to run End-2-End tests:
  * `test:e2e`: Alias for `test:e2e:local`.
  * `test:e2e:local`: Run the test suite on Chrome browser on your local machine.
  * `test:e2e:saucelabs`: Run the test suite on multiple browsers on the Sauce Labs service.
* Rename existing `test:e2e` to `test:integration`.
* Update CircleCI configuration
  * Rename `test_e2e` task to `test_integration`.
  * Add the new `test_e2e` task.


.circleci/config.yml
```diff
@@ -68,8 +68,22 @@ jobs:
...
+      - checkout
+      - run: *install_yarn_version
+      - restore_cache: *restore_yarn_cache
+      - run: *run_yarn_install
+      - save_cache: *save_yarn_cache
+      - run:
+          name: End-2-End tests
+          command: |
+            yarn run website:build
+            yarn run test:e2e:saucelabs
```
Haroenv: Looks like this builds the lib again, could use the ci workflow more efficiently

yannickcr: I had the same issue on `instantsearch.js`, but it seems that in the current configuration each CircleCI job is isolated in it's own environment, which forces us to re-do a lot of stuff in each job (checkout, yarn install, yarn build).

IMHO it is out for scope for this PR (unless I made an error or I missed an obvious optimization :wink: ) but there is room for improvement for sure. I think we could tackle this in a later task.

tkrugg: IMHO it's simpler if it's redoing it, especially if you want to rerun a particular job like it's often the case with E2E. So this is ok to me.

package.json
```diff
@@ -60,6 +63,13 @@
     "@types/storybook__react": "^4.0.0",
     "@typescript-eslint/eslint-plugin": "2.2.0",
     "@typescript-eslint/parser": "2.2.0",
+    "@wdio/cli": "5.11.14",
+    "@wdio/jasmine-framework": "5.12.1",
+    "@wdio/local-runner": "5.12.1",
+    "@wdio/sauce-service": "5.12.1",
+    "@wdio/selenium-standalone-service": "5.12.1",
+    "@wdio/spec-reporter": "5.12.1",
+    "@wdio/static-server-service": "5.12.1",
```
tkrugg: I think I asked this question already but I'm no longer sure what's the answer. 
Is it not possible to hide this away in `instantsearch-e2e-tests` ?

Haroenv: I also think we should hide this

package.json
```diff
@@ -85,6 +95,7 @@
     "file-loader": "4.2.0",
     "glob": "7.1.4",
     "html-webpack-plugin": "3.2.0",
+    "instantsearch-e2e-tests": "algolia/instantsearch-e2e-tests#v1.2.1",
```
francoischalifour: Last release is v1.2.2.

```suggestion
    "instantsearch-e2e-tests": "algolia/instantsearch-e2e-tests#v1.2.2",
```

package.json
francoischalifour: It would also make sense to me.

samouss: Agree.

yannickcr: Well... I'm not for this.

First of all there is very few advantages (if any?) to "hide" dependencies but the real problem is mainly that it will makes us call a subdependency from the parent project (`react-instantsearch` calls `wdio` directly), which is not recommenced (the package version we're targeting may not be at the `node_modules` top level, depending how npm/yarn deduplicated the dependencies).

Haroenv: can we also hide `wdio` call then?

yannickcr: What would be the objective of hiding these dependencies?

francoischalifour: It would remove the overall perceived complexity, as well as reduce the noise and cognitive overhead of managing these dependencies.

I find it surprising that we went all the way to packaging the E2E tests (e.g. see also `wdio.saucelabs.conf.js` which only exports the E2E package import), but are reluctant to hide the dependencies themself.

The E2E test suite would be more predictable if the `wdio` dependencies are packaged within it. Here, we are responsible for the `wdio` deps to work with a black-boxed test suite, which is quite contradictory.

PR #2833, opened 2019-09-17
Upgrade select dev dependencies for Node 10+ compatibility

coreyward:
**Summary**

As a result of incompatibilities of some dependencies with newer versions of Node, the first-time contribution experience for users on Node 10+ is rather poor with multiple errors upon running `yarn` on the cloned codebase.

- upath@1.0.4 is incompatible with Node >v9
- fsevents@1.2.7 is incompatible with Node >v11

This minimum viable dependency upgrade resolves both instances of chokidar (`^2.0.2` and `^2.0.4`) to the same version (`2.1.8`) that has an updated dependency on `upath`. It also upgrades resolution of `fsevents` from `1.2.7` to `1.2.9`. 

The remaining changes to yarn.lock are a result of the downstream dependency resolutions performed by Yarn during `yarn install`.

Test suite passes fully green after the changes.


coreyward:
@samouss Done.

PR #2834, opened 2019-09-17
chore(index): don't require `indexId` on `IndexWrapper`

### LINKED ISSUE #2832: v6.0.0-beta.0 requires `indexId` on `<Index>`
There appears to be a transition going on from `indexName` to `indexId`, but this is incomplete and not documented. I think the `IndexWrapper` component was added to make this transition easier, but the PropTypes set for it require _both_ `indexName` and `indexId` despite synthesizing the latter from the former:

https://github.com/algolia/react-instantsearch/blob/35914beae2e6bc4cc6229f3627bc3cbba08c5760/packages/react-instantsearch-core/src/widgets/Index.tsx#L126-L144

The result of this is PropType errors being displayed on the client inside of downstream applications when running React in `development` mode.

coreyward:
This loosens the PropType requirement on `IndexWrapper` to allow it to work as it seems to be intended (inferring the `indexId` from the `indexName` during the transition phase).

This fixes #2832.

### DIFF
--- a/packages/react-instantsearch-core/src/widgets/Index.tsx
+++ b/packages/react-instantsearch-core/src/widgets/Index.tsx
@@ -123,7 +123,12 @@ class Index extends Component<InnerProps, State> {
   }
 }
 
-const IndexWrapper: React.FC<Props> = props => {
+type IndexWrapperProps = {
+  indexName: string;
+  indexId?: string;
+};
+
+const IndexWrapper: React.FC<IndexWrapperProps> = props => {
   const inferredIndexId = props.indexName;
   return (
     <InstantSearchConsumer>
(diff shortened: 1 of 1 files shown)


Haroenv:
I can't seem to suggest this, but this will be correct:

```ts
type IndexWrapperProps = {
  indexName: string;
  indexId?: string;
};

const IndexWrapper: React.FC<Props> = props => {
  const inferredIndexId = props.indexName;
  return (
    <InstantSearchConsumer>
      {contextValue => (
        <Index
          contextValue={contextValue}
          indexId={inferredIndexId}
          {...props}
        />
      )}
    </InstantSearchConsumer>
  );
};

IndexWrapper.propTypes = {
  indexName: PropTypes.string.isRequired,
  indexId: PropTypes.string,
};
```

thanks again!

coreyward:
Whoops, not in the habit of writing TypeScript. Added the type declaration for the IndexWrapperProps as suggested.

Haroenv:
thanks again!

PR #2852, opened 2019-10-03
fix(InstantSearch): update the searchState on props update

samouss:
**Summary**

This PR fixes an issue where an infinite loop occurs with the controlled mode. The update of the `searchState` was done too late. With V5 the update was done inside `cWRP` but with V6 we move the update inside `cDU`. This change has an impact on when the `searchState` is updated. With `cWRP` it was executed before the `render` of the children, it's not the case anymore with `cDU`. It's done after.

Those children can read the `searchState` provided by the manager. When it's not up to date they alter the previous version of the state. With an `onSearchStateChange` function that relies on the value of `searchState` to add remove parameters, it can lead to infinite loop because the props are changes/reverted at each render.

The fix uses `getDerivedStateFromProps` to keep the `searchState` up to date. The issue with this solution (on V5 too) is that we might trigger one useless request when the props change. It's because the update and the search are tied together inside `onExternalStateUpdate`. One solution to this would be to decouple the two. Apply the `searchState` update as soon as possible but trigger the search on `cDU` once we've computed the state from the children.

**Before**

before

**After**

after

### DIFF
--- a/packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
+++ b/packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
@@ -1,4 +1,5 @@
 import React, { Component, Children } from 'react';
+import isEqual from 'fast-deep-equal';
 import PropTypes from 'prop-types';
 import createInstantSearchManager from '../core/createInstantSearchManager';
 import { InstantSearchProvider, InstantSearchContext } from '../core/context';
@@ -56,6 +57,8 @@ type Props = {
 };
 
 type State = {
+  isControlled: boolean;
+  instantSearchManager: InstantSearchManager;
   contextValue: InstantSearchContext;
 };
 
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -136,7 +139,19 @@ class InstantSearch extends Component<Props, State> {
     nextProps: Props,
     prevState: State
   ): Partial<State> {
+    const nextIsControlled = isControlled(nextProps);
+    const previousSearchState = prevState.instantSearchManager.store.getState()
+      .widgets;
+    const nextSearchState = nextProps.searchState;
+
+    if (nextIsControlled && !isEqual(previousSearchState, nextSearchState)) {
```
Haroenv: it's indeed more likely to cause issues with infinite rendering, than this being bad for perf

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -145,83 +160,83 @@ class InstantSearch extends Component<Props, State> {
...
+    this.state = {
+      isControlled: isControlled(this.props),
+      instantSearchManager,
+      contextValue,
+    };
+  }
 
   componentDidUpdate(prevProps: Props) {
-    const nextIsControlled = isControlled(this.props);
+    const prevIsControlled = isControlled(prevProps);
```
Haroenv: good catch!

PR #2856, opened 2019-10-11
refactor(nextjs): move all fetch logic to `getInitialProps`

wedneyyuri:
**Summary**

After refining a search, it is usually necessary to fetch related content such as suggestions, banners, ads, ...
For this reason, I'm suggesting to to keep all the fetch logic inside a single method: `getInitialProps`. Another benefit of this change is that we can remove the `Page` state, making the component easier to understand.

From the [Nextjs docs](https://nextjs.org/learn/basics/fetching-data-for-pages):
> In practice, we usually need to fetch data from a remote data source. Next.js comes with a standard API to fetch data for pages. We do it using an async function called `getInitialProps`.

I believe that with this change users who are using NextJS will have an easier time understanding and adapting the application to be able to use Algolia.

**Result**

I'm still trying to solve this issue, but after this change the component is **searching twice**. Any help will be appreciated.

algolia-pull-request

### DIFF
--- a/examples/next/components/app.js
+++ b/examples/next/components/app.js
@@ -2,13 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {
   RefinementList,
-  SearchBox,
   Hits,
   Configure,
   Highlight,
   Pagination,
 } from 'react-instantsearch-dom';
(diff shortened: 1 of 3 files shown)


wedneyyuri:
Closed by #2857

PR #2857, opened 2019-10-11
refactor(nextjs): move all fetch logic to `getInitialProps`

wedneyyuri:
**Summary**

After refining a search, it is usually necessary to fetch related content such as suggestions, banners, ads, ...
For this reason, I'm suggesting to to keep all the fetch logic inside a single method: `getInitialProps`. Another benefit of this change is that we can remove the `Page` state, making the component easier to understand.

From the [Nextjs docs](https://nextjs.org/learn/basics/fetching-data-for-pages):
> In practice, we usually need to fetch data from a remote data source. Next.js comes with a standard API to fetch data for pages. We do it using an async function called `getInitialProps`.

I believe that with this change users who are using NextJS will have an easier time understanding and adapting the application to be able to use Algolia.

**Result**

algoliapullrequest-2

### DIFF
--- a/examples/next/components/app.js
+++ b/examples/next/components/app.js
@@ -2,13 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {
   RefinementList,
-  SearchBox,
   Hits,
   Configure,
   Highlight,
   Pagination,
   InstantSearch,
 } from 'react-instantsearch-dom';
+import SearchBox from './searchbox';
 
 const HitComponent = ({ hit }) => (
   <div className="hit">
--- a/examples/next/components/searchbox.js
+++ b/examples/next/components/searchbox.js
@@ -1,0 +1,38 @@
+import React, { Component } from 'react';
+import { connectSearchBox, SearchBox } from 'react-instantsearch-dom';
+
+const updateAfter = 700;
+
+class SearchBoxWithDebounce extends Component {
+  timerId = null;
+
+  state = {
+    value: this.props.currentRefinement,
+  };
+
+  onChange = event => {
+    const value = event.currentTarget.value;
+    const { refine } = this.props;
+
+    clearTimeout(this.timerId);
+    this.timerId = setTimeout(() => refine(value), updateAfter);
+
+    this.setState(() => ({
+      value,
+    }));
+  };
+
+  render() {
+    const { value } = this.state;
+    return (
+      <SearchBox
+        {...this.props}
+        searchAsYouType={false}
+        currentRefinement={value}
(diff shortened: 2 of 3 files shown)


Haroenv:
Thanks for your contribution, but I don't see how this is related to Next. Instead you have made the search box debounced, which means that it's no longer "instant search". We have a guide on how to make a debounced searchbox already here: https://www.algolia.com/doc/guides/building-search-ui/going-further/improve-performance/react/#debouncing

Can you clarify what you meant about `getInitialProps`, I don't see it in the difference you suggest.

Thanks again for contributing!

wedneyyuri:
Thank you for your comments @Haroenv. I will try to be a clearer.

> Instead you have made the search box debounced, which means that it's no longer "instant search".

The original searchbox was already debounced, you can see it here (https://github.com/algolia/react-instantsearch/blob/29a7351304c493b1d2cd8dcd4f5305db63ce88e1/examples/next/pages/index.js#L69). This is necessary to avoid page slowness while the user is typing the search term. I believe that extracting the debounce logic into another component helps simplify the component `pages/index.js`.

>
> Can you clarify what you meant about getInitialProps, I don't see it in the difference you suggest.

My goal is to keep the example as simple as possible, showing how simple it is to combine NextJS and Algolia to load data on the server and rehydrate on the client. 

A few points that I believe we can improve on compared to the current implementation:
- It is necessary to maintain a state. It can be confusing to understand that during server side rendering `searchState` is part of the props, but during rehydratation `searchState` in inside state.

- In the current implementation it is difficult to load additional resources. For example, how to load other APIs in paralell (ex: price, stock, media, ads, promotion) every time the search change?

- The commonly used method for side effects is being (`getInitialProps`) underused. It is much more natural for the developer to fetch data inside the method `getinitialprops` than inside any Algolia component life cycle.

Haroenv:
An important distinction between before & after is that before the URL is debounced, now the search is debounced. This behaviour is not the same :)

Haroenv:
> load other APIs in paralell 

This is a valid use case which isn't fully solved now indeed. I'd suggest a strategy described like here: https://github.com/algolia/instantsearch.js/issues/4166#issuecomment-542029798, which is indeed side-stepping getInitialProps. 

Do you have an example of another library which handles their own fetches in the context of Next, like e.g. with Apollo? 

While I understand your point about getInitialProps, React InstantSearch works without Next as well, where we can't design for getInitialProps

Haroenv:
As discussed, while I understand your idea, the behaviour is different than the current one, and thus we can't give this as an example. 

Thanks for your contribution!

Issue #2828, opened 2019-09-15
JSON parse error when reading the cache


andreyvital:
**Describe the bug 🐛**

```
SyntaxError: Unexpected token o in JSON at position
```

Sorry I am not able to create an example right now, but something fishy is going on: I am using Next.js to SSR Algolia with results, and:

1. I enable one filter in my app then it updates the URL;
2. Right after that the result is served and a cache is stored for that state;
3. Then when I "remove" that filter (resetting to the initial value) it crashes with the error above.

I was stepping through it in the debugger and apparently the 1st cache entry is correct: JSON serialized response. When I change the filter to the initial value, the object that gets stored there is an object (not a JSON string) then it crashes when `JSON.parse` is executed.

I'll try to provide an example later on but do you have any thoughts? I am currently stuck with this issue and it is limiting me to move forward. 😢

Haroenv:
Hi @andreyvital, I'm not sure which cache you are talking about, can you clarify with an example? We have a demo setup with next here: https://codesandbox.io/s/github/algolia/react-instantsearch/tree/master/examples/next

andreyvital:
Yeah sure, try this one: https://codesandbox.io/s/example-next-cifw3

tkrugg:
Hi @andreyvital 

I just released React InstantSearch 6.0.0-beta.2 and updated your example to prevent the second query.

https://codesandbox.io/s/example-next-93sq3

There 2 things I did:
1. update react-instantsearch-dom to 6.0.0-beta.2
2. add onSearchParameters prop (see my screenshot of your modified `components/app.js` file). We'll make sure to make all this clear in the docs.

![image](https://user-images.githubusercontent.com/2982512/67556239-16c59e80-f713-11e9-9a7f-10d6385b017c.png)

I hope this helps!

Haroenv:
The docs says to spread the props (which can be done here too)

PR #2862, opened 2019-10-23
fix: serialize cache value on hydrate

tkrugg:
**Summary**

It appears the search client cache uses string for values `{ [key: string]: string }`.
When hydrating it for preventing a new request on the client, we need to
serialize the value into a JSON string.

fixes: #2828

**Result**

N/A


Haroenv:
this fixes the error as far as I can tell! It used to be wrong, but not noticed, since the missed cache throws a JSON error, which is caught by the helper, and not displayed to users.

Issue #2835, opened 2019-09-17
v6.0.0-beta sends two queries (with and without `page`)


coreyward:
I did a straight upgrade from 5.7 to 6.0.0-beta and noticed in the Network panel of the devtools that the updated version of react-instantsearch-dom is sending two POST requests to the Algolia service for each query. The requests appear to be identical except for an additional `page` param (always set to `0` in my testing) that is in the POSTed parameters of one of the two requests. 

This only occurs when using two different `<Index>` components with unique `indexName` props. Sometimes the `page` version of the query is shown first, and other times its shown second. This is not an issue with v5.7.0.

**To Reproduce 🔍**

I wasn't able reproduce the issue in the CodeSandbox example as it does not document a second index.

For reference, here’s a diff of the two POST bodies for the requests made:

```diff
 {
   "requests": [
     {
       "indexName": "collections",
       "params": "highlightPreTag=%3Cais-highlight-0000000000%3E&highlightPostTag=%3C%2Fais-highlight-0000000000%3E&hitsPerPage=6&query=good%20doggo&facets=%5B%5D&tagFilters="
     },
     {
       "indexName": "stories",
-      "params": "highlightPreTag=%3Cais-highlight-0000000000%3E&highlightPostTag=%3C%2Fais-highlight-0000000000%3E&hitsPerPage=8&query=good%20doggo&page=0&facets=%5B%5D&tagFilters="
+      "params": "highlightPreTag=%3Cais-highlight-0000000000%3E&highlightPostTag=%3C%2Fais-highlight-0000000000%3E&hitsPerPage=8&query=good%20doggo&facets=%5B%5D&tagFilters="
     }
   ]
 }

```

Haroenv:
It would be ideal if you could reproduce this in a sandbox, here's an example on v6 with two indices that I don't see having this issue. https://codesandbox.io/s/multi-index-hits-ds70x

Thanks already for all the feedback you've given, this is very appreciated

andreyvital:
@Haroenv I am having the same problem here with v6.0.0-beta.1. It's sending two queries...

tkrugg:
@andreyvital 's issue has been addressed: see https://github.com/algolia/react-instantsearch/issues/2829

@coreyward I'll close this for now because we're missing a reproducible example but feel free to reopen if you're still experiencing the issue and have managed to reproduce it in our boilerplate.

PR #2868, opened 2019-11-06
fix(ssr): make `page` param in `searchState` to be included in query

eunjae-lee:
**Summary**

In SSR, adding `page` into `searchState` doesn't work. `page` isn't included in the query.

**Result**

This test should pass. (which currently doesn't)


tkrugg:
this looks related: https://github.com/algolia/react-instantsearch/issues/2835

PR #2869, opened 2019-11-07
Update readme to include ReactDOM.render

mateuspv:
I would like to include `ReactDOM.render` on the doc as well,
I've spent some time debugging why it was not rendering `App` instead of `<App/>`

**Summary**

**Result**

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,8 @@ const App = () => (
     <Hits />
   </InstantSearch>
 );
+
+ReactDOM.render(<App />, document.querySelector('#container'))
 ```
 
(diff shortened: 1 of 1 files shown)


Haroenv:
Hi @mateuspv, thanks for your contribution. I'm interested to see why you think that adding this line is essential? It might be more logical to remove the react-dom import as well. This kind of info seems more relevant in a react getting started guide that here no? 

Thanks!

Haroenv:
Thanks for your contribution, but we did not feel this change was necessary. Hope to see more of you though!

PR #2871, opened 2019-11-18
fix(helper): rely on stable version of algoliasearch-helper

Haroenv:
**Summary**

We released while still relying on a canary version of the helper, this means that people would not get patch updates to the helper if we ever did so

Note that also we were relying on two different canary releases, so people likely had two separate versions of the helper in a react-instantsearch-dom bundle.

**Result**

By relying on the real released v3 now, we are sure people will get the latest bug fixes and no longer having duplicate versions of the helper

### DIFF
--- a/packages/react-instantsearch-dom/package.json
+++ b/packages/react-instantsearch-dom/package.json
@@ -41,7 +41,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch-helper": "0.0.0-27095c0",
+    "algoliasearch-helper": "^3.0.0",
     "classnames": "^2.2.5",
     "prop-types": "^15.5.10",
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-dom/package.json
```diff
@@ -41,7 +41,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch-helper": "0.0.0-27095c0",
```
eunjae-lee: It's weird that we had

`"algoliasearch-helper": "0.0.0-6ac260d"` in `react-instantsearch-core` and
`"algoliasearch-helper": "0.0.0-27095c0"` in `react-instantsearch-dom`.

Anyway the changes look good to me.

PR #2877, opened 2019-12-03
feat(insights): show an error when 'clickAnalytics: true' is missing.

tkrugg:
**Summary**
Similar to its InstantSearch.js counterpart https://github.com/algolia/instantsearch.js/pull/4231 
When clickAnalytics: true is missing, we fire an error inviting the user to add `clickAnalytics: true`

**Result**

N/A

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectHitInsights.ts
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectHitInsights.ts
@@ -102,4 +102,39 @@ describe('connectHitInsights', () => {
       );
     });
   });
+
+  describe('when queryID is undefined', () => {
+    it('should throw an error message inviting to add clickAnalytics: true', () => {
+      const insightsClient = jest.fn();
+
+      const contextValue = {
+        mainTargetedIndex: 'firstIndex',
+      };
+      const indexContextValue = {
+        targetedIndex: 'theIndex',
+      };
+
+      const hit = {
+        objectID: 'objectID_42',
+        __position: 42,
+        // no queryID
+      };
+
+      const searchResults = { results: { theIndex: { index: 'theIndex' } } };
+      const props = connect(insightsClient).getProvidedProps(
+        { hit, contextValue, indexContextValue },
+        null,
+        searchResults
+      );
+
+      expect(() => {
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-core/src/connectors/connectHitInsights.ts
```diff
@@ -33,6 +33,12 @@ function inferPayload({
   const { index } = results;
   const queryID = currentHit.__queryID;
   const objectIDs = [currentHit.objectID];
+
+  if (!queryID) {
+    throw new Error(`Could not infer \`queryID\`. Ensure \`clickAnalytics: true\` was added with the Configure widget.
+See: https://alg.li/lNiZZ7`);
```
francoischalifour: Here we could be pointing to the [React doc](https://www.algolia.com/doc/guides/getting-insights-and-analytics/search-analytics/click-through-and-conversions/how-to/send-click-and-conversion-events-with-instantsearch/react/#prepare-react-instantsearch-to-connect-with-insights-client-for-javascript) directly.

packages/react-instantsearch-core/src/connectors/__tests__/connectHitInsights.ts
```diff
@@ -102,4 +102,39 @@ describe('connectHitInsights', () => {
...
+        searchResults
+      );
+
+      expect(() => {
+        props.insights('clickedObjectIDsAfterSearch', {
+          eventName: 'Add to wishlist',
+        });
+      }).toThrowErrorMatchingInlineSnapshot(`
+"Could not infer \`queryID\`. Ensure \`clickAnalytics: true\` was added with the Configure widget.
+See: https://alg.li/lNiZZ7"
```
tkrugg:
```suggestion
See: https://alg.li/VpPpLt"
```

packages/react-instantsearch-core/src/connectors/connectHitInsights.ts
tkrugg:
```suggestion
See: https://alg.li/VpPpLt`);
```

tkrugg: Good point! changed it to https://alg.li/VpPpLt

PR #2880, opened 2019-12-11
feat(widgets): introduce `ConfigureRelatedItems` as experimental

francoischalifour:
This PR introduces a new set of experimental widgets:

- `EXPERIMENTAL_ConfigureRelatedItems`
- `EXPERIMENTAL_connectConfigureRelatedItems`

Combined with the `Hits` component, these widgets allow to create related items experiences as seen in e-commerce and media use cases:

Example

<details>

<summary>See more examples</summary>

Example

Example

</details>

## Result

- [Usage](https://github.com/algolia/react-instantsearch/blob/4ab5da03bc35e2739f9dd0ce6974d55ebb8e1840/stories/ConfigureRelatedItems.stories.tsx)
- [Story](https://deploy-preview-2880--react-instantsearch.netlify.com/storybook/?path=/story/configurerelateditems--default)

## Related

- Widgets in InstantSearch.js: algolia/instantsearch.js#4233
- In the story I used `React.memo` for the component not to rerender when the `objectID` doesn't change between renders. Our PropTypes don't support memoized components because we expect [`hitComponent` to be a `func`](https://github.com/algolia/react-instantsearch/blob/621656a8f25377094c6a3b06debc8d43f4e3d6c6/packages/react-instantsearch-dom/src/components/Hits.tsx#L60) while it should be an `elementType`. We can tackle that in another PR.

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
+++ b/packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
@@ -122,7 +122,7 @@ See https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/`
       // `sumOrFiltersScores`.
       // See https://github.com/algolia/algoliasearch-helper-js/pull/753
       sumOrFiltersScores: true,
-      filters: `NOT objectID:${props.hit.objectID}`,
+      facetFilters: [`objectID:-${props.hit.objectID}`],
       optionalFilters,
     })
   );
--- a/packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
+++ b/packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
@@ -1,0 +1,220 @@
+import algoliasearchHelper, {
+  PlainSearchParameters,
+  SearchParameters,
+} from 'algoliasearch-helper';
+import createConnector, { ConnectedProps } from '../core/createConnector';
+import {
+  omit,
+  getObjectType,
+  removeEmptyKey,
+  removeEmptyArraysFromObject,
+} from '../core/utils';
+import {
+  refineValue,
+  getIndexId,
+  hasMultipleIndices,
+} from '../core/indexUtils';
+
+type Hit = any;
+
+export type MatchingPatterns = {
+  [attribute: string]: {
+    /**
+     * The score of the optional filter.
+     *
+     * @see https://www.algolia.com/doc/guides/managing-results/rules/merchandising-and-promoting/in-depth/optional-filters/
+     */
+    score: number;
+  };
+};
+
+interface ConfigureRelatedItemsProps {
+  /**
+   * The reference hit to extract the filters from.
+   */
+  hit: Hit;
+  /**
+   * The schema to create the optional filters.
+   * Each key represents an attribute from the hit.
+   */
+  matchingPatterns: MatchingPatterns;
+  /**
+   * Function called to transform the search parameters generated.
+   */
+  transformSearchParameters?(
+    searchParameters: SearchParameters
+  ): PlainSearchParameters;
+}
+
+function createOptionalFilter({
+  attributeName,
+  attributeValue,
+  attributeScore,
+}) {
+  return `${attributeName}:${attributeValue}<score=${attributeScore || 1}>`;
+}
+
+const defaultProps: Partial<ConfigureRelatedItemsProps> = {
+  transformSearchParameters: x => ({ ...x }),
+};
+
+function getId(): string {
+  // We store the search state of this widget in `configure`.
+  return 'configure';
+}
+
+type InternalConfigureRelatedItemsProps = ConfigureRelatedItemsProps &
+  Required<typeof defaultProps>;
+
+function getSearchParametersFromProps(
+  props: ConnectedProps<InternalConfigureRelatedItemsProps>
+): PlainSearchParameters {
+  const optionalFilters = Object.keys(props.matchingPatterns).reduce<
+    Array<string | string[]>
+  >((acc, attributeName) => {
+    const attribute = props.matchingPatterns[attributeName];
+    const attributeValue = props.hit[attributeName];
+    const attributeScore = attribute.score;
+
+    if (Array.isArray(attributeValue)) {
+      return [
+        ...acc,
+        attributeValue.map(attributeSubValue => {
+          return createOptionalFilter({
+            attributeName,
+            attributeValue: attributeSubValue,
+            attributeScore,
+          });
+        }),
+      ];
+    }
+
+    if (typeof attributeValue === 'string') {
+      return [
+        ...acc,
+        createOptionalFilter({
+          attributeName,
+          attributeValue,
+          attributeScore,
+        }),
+      ];
+    }
+
+    if (process.env.NODE_ENV === 'development') {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `The \`matchingPatterns\` option returned a value of type ${getObjectType(
+          attributeValue
+        )} for the "${attributeName}" key. This value was not sent to Algolia because \`optionalFilters\` only supports strings and array of strings.
+
+You can remove the "${attributeName}" key from the \`matchingPatterns\` option.
+
+See https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/`
+      );
+    }
+
+    return acc;
+  }, []);
+
+  return props.transformSearchParameters(
+    new algoliasearchHelper.SearchParameters({
+      // @ts-ignore @TODO algoliasearch-helper@3.0.1 will contain the type
+      // `sumOrFiltersScores`.
+      // See https://github.com/algolia/algoliasearch-helper-js/pull/753
+      sumOrFiltersScores: true,
+      filters: `NOT objectID:${props.hit.objectID}`,
+      optionalFilters,
+    })
+  );
+}
+
+interface ConnectorState {
+  _searchParameters: PlainSearchParameters;
+}
+
+export default createConnector({
+  displayName: 'AlgoliaConfigureRelatedItems',
+
+  defaultProps,
+
+  getProvidedProps() {
+    return {};
+  },
+
+  getSearchParameters(
+    searchParameters: SearchParameters,
+    props: ConnectedProps<InternalConfigureRelatedItemsProps>
+  ) {
+    return searchParameters.setQueryParameters(
+      getSearchParametersFromProps(props)
+    );
+  },
+
+  transitionState(
+    this: ConnectorState,
+    props,
+    _prevSearchState,
+    nextSearchState
+  ) {
+    const id = getId();
+    // We need to transform the exhaustive search parameters back to clean
+    // search parameters without the empty default keys so we don't pollute the
+    // `configure` search state.
+    const searchParameters = removeEmptyArraysFromObject(
+      removeEmptyKey(getSearchParametersFromProps(props))
+    );
+
+    const searchParametersKeys = Object.keys(searchParameters);
+    const nonPresentKeys = this._searchParameters
+      ? Object.keys(this._searchParameters).filter(
+          prop => searchParametersKeys.indexOf(prop) === -1
+        )
+      : [];
+    this._searchParameters = searchParameters;
+    const nextValue = {
+      [id]: {
+        ...omit(nextSearchState[id], nonPresentKeys),
+        ...searchParameters,
+      },
+    };
+
+    return refineValue(nextSearchState, nextValue, {
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
+    });
+  },
+
+  cleanUp(this: ConnectorState, props, searchState) {
+    const id = getId();
+    const indexId = getIndexId({
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
+    });
+
+    const subState =
+      hasMultipleIndices({
+        ais: props.contextValue,
+        multiIndexContext: props.indexContextValue,
+      }) && searchState.indices
+        ? searchState.indices[indexId]
+        : searchState;
+
+    const configureKeys =
+      subState && subState[id] ? Object.keys(subState[id]) : [];
+
+    const configureState = configureKeys.reduce((acc, item) => {
+      if (!this._searchParameters[item]) {
+        acc[item] = subState[id][item];
+      }
+
+      return acc;
+    }, {});
+
+    const nextValue = { [id]: configureState };
+
+    return refineValue(searchState, nextValue, {
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
+    });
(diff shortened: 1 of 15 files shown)


packages/react-instantsearch-core/src/core/utils.ts
```diff
@@ -128,3 +140,7 @@ export const getPropertyByPath = (object: object, path: string[] | string) =>
     ? path
     : path.replace(/\[(\d+)]/g, '.$1').split('.')
   ).reduce((current, key) => (current ? current[key] : undefined), object);
+
+export function getObjectType(object: unknown): string {
```
francoischalifour: See algolia/instantsearch.js#4249.

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+      return acc;
+    }, {});
+
+    const nextValue = { [id]: configureState };
+
+    return refineValue(searchState, nextValue, {
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
+    });
+  },
```
francoischalifour: The implementation is mostly taken from:

https://github.com/algolia/react-instantsearch/blob/621656a8f25377094c6a3b06debc8d43f4e3d6c6/packages/react-instantsearch-core/src/connectors/connectConfigure.js#L38-L68

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+        ...omit(nextSearchState[id], nonPresentKeys),
+        ...searchParameters,
+      },
+    };
+
+    return refineValue(nextSearchState, nextValue, {
+      ais: props.contextValue,
+      multiIndexContext: props.indexContextValue,
+    });
+  },
```
francoischalifour: The implementation is mostly taken from:

https://github.com/algolia/react-instantsearch/blob/621656a8f25377094c6a3b06debc8d43f4e3d6c6/packages/react-instantsearch-core/src/connectors/connectConfigure.js#L22-L37

packages/react-instantsearch-core/src/index.ts
```diff
@@ -8,6 +8,10 @@ export { default as translatable } from './core/translatable';
 
 // Widgets
 export { default as Configure } from './widgets/Configure';
+export {
+  // eslint-disable-next-line @typescript-eslint/camelcase
```
Haroenv: can we change the rule, like in InstantSearch.js?

stories/ConfigureRelatedItems.stories.tsx
```diff
@@ -0,0 +1,115 @@
+/* eslint react/prop-types: 0 */
```
Haroenv: we can disable this on .tsx in the stories globally maybe?

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+  removeEmptyKey,
+  removeEmptyArraysFromObject,
+} from '../core/utils';
+import {
+  refineValue,
+  getIndexId,
+  hasMultipleIndices,
+} from '../core/indexUtils';
+
+type Hit = any;
```
Haroenv: good enough for now, but we should consolidate the hit types at some point next quarter

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+   * The reference hit to extract the filters from.
+   */
+  hit: Hit;
+  /**
+   * The schema to create the optional filters.
+   * Each key represents an attribute from the hit.
+   */
+  matchingPatterns: MatchingPatterns;
+  /**
+   * Function called to transform the search parameters generated.
```
Haroenv:
```suggestion
   * Function to transform the generated search parameters.
```

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+    searchParameters: SearchParameters
+  ): PlainSearchParameters;
+}
+
+function createOptionalFilter({
+  attributeName,
+  attributeValue,
+  attributeScore,
+}) {
+  return `${attributeName}:${attributeValue}<score=${attributeScore || 1}>`;
```
Haroenv: probably the same for InstantSearch.js, but don't we need to filter out & escape some values?

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+}
+
+const defaultProps: Partial<ConfigureRelatedItemsProps> = {
+  transformSearchParameters: x => ({ ...x }),
+};
+
+function getId(): string {
+  // We store the search state of this widget in `configure`.
+  return 'configure';
```
Haroenv: will this merge or override if you have a separate configure in the UI too?

Haroenv: We probably need a namespace (hit's objectID)? Although OTOH it doesn't make sense to make related items multiple times?

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+     */
+    score: number;
+  };
+};
+
+interface ConfigureRelatedItemsProps {
+  /**
+   * The reference hit to extract the filters from.
+   */
+  hit: Hit;
```
Haroenv: It's somewhat late now to make this change, but what if we allow `hits: Hit[]`, so people can create related hits when a product page or home page should be related to multiple items?

We can still consider this later before we move away from `EXPERIMENTAL_`

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+type InternalConfigureRelatedItemsProps = ConfigureRelatedItemsProps &
+  Required<typeof defaultProps>;
+
+function getSearchParametersFromProps(
+  props: ConnectedProps<InternalConfigureRelatedItemsProps>
+): PlainSearchParameters {
+  const optionalFilters = Object.keys(props.matchingPatterns).reduce<
+    Array<string | string[]>
+  >((acc, attributeName) => {
+    const attribute = props.matchingPatterns[attributeName];
```
Haroenv:
```suggestion
    const attributePattern = props.matchingPatterns[attributeName];
```

not sure if this makes more sense, but I was confused what this could be initially

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+    const attributeValue = props.hit[attributeName];
+    const attributeScore = attribute.score;
```
Haroenv:
```suggestion
    const attributeScore = attributePattern.score;
```

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
Haroenv: maybe also switch line 75 & 76

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
eunjae-lee: Do you mean to create `optionalFilters` based on multiple hits?
For example, if there're two hits(Apple, Amazon), then `optionalFilters` becomes like this?
```
[
    ["brand:Apple<score=3> OR brand:Amazon<score=3>"]
]
```

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
francoischalifour: Yep! Depending our on plan on the IS.js migration, we might leverage the existing one from IS.js.

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
francoischalifour: Interesting! Definitely worth exploring before getting out of the experimental phase.

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
francoischalifour: Filtering out based on what? Only `string | string[]` are taken into account.

Regarding escaping, yes, we probably should. Any suggestions as to how escaping the values for Algolia to understand them all the time?

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
francoischalifour: Clearer indeded.

packages/react-instantsearch-core/src/index.ts
francoischalifour: Will do.

stories/ConfigureRelatedItems.stories.tsx
francoischalifour: Will do.

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
francoischalifour: It merges them, as [tested here](https://github.com/algolia/react-instantsearch/blob/4ab5da03bc35e2739f9dd0ce6974d55ebb8e1840/packages/react-instantsearch-core/src/connectors/__tests__/connectConfigureRelatedItems.ts#L179-L231).

francoischalifour: I'm not sure it makes much sense to use multiple `ConfigureRelatedItems` widget within an `Index`, but we'll see which patterns we'll come up with when using the feature.

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
Haroenv: a value can not start with `- `, since that's a negation, quotes need to be escaped, since they have special meaning to group things, and `<`, `>` needs to be escaped because it's xml and used in score (escaped means that in the final json there's `\\<`, because the backslash needs to be escaped itself.

I can't find a reference document for that

Haroenv: https://www.algolia.com/doc/api-reference/api-parameters/facetFilters/#usage-notes, I was correct according to @xroche, but `<` & `>` need to be removed, not escaped

francoischalifour: Thanks for looking into it! I'll take care of that in a stacked PR.

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
eunjae-lee: but doesn't it also override the things set by `connectConfigure`, not only the things by `connectConfigureRelatedItems`? I think that's what Haroen is talking about.

francoischalifour: Given

```
- Index
  - Configure
  - ConfigureRelatedItems
```

If you've got `filters` in `Configure`, they'll get overridden by the ones computed in `ConfigureRelatedItems`. I don't think we should start merging them because we'll face a bunch of edge cases.

We could however display a development warning when we detect collisions and that some were overridden.

Wdyt?

Haroenv: Hmm, good point, but we can avoid it with a separate key no?

francoischalifour: I would complexify everything just for an edge case: we'll introduce another level of nesting in the search state. I don't think we should update the structure, especially given that it's experimental. Note that we don't do this either in InstantSearch.js.

francoischalifour:
I will address https://github.com/algolia/react-instantsearch/pull/2880#discussion_r356567196 and https://github.com/algolia/react-instantsearch/pull/2880#discussion_r357068143 in other PRs to minimize the changes here. You can consider this PR as ready to review @eunjae-lee @yannickcr.

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
eunjae-lee: It makes sense. It sounds good enough with dev warning 👍

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,220 @@
...
+    Array<string | string[]>
+  >((acc, attributeName) => {
+    const attributePattern = props.matchingPatterns[attributeName];
+    const attributeValue = props.hit[attributeName];
+    const attributeScore = attributePattern.score;
+
+    if (Array.isArray(attributeValue)) {
+      return [
+        ...acc,
+        attributeValue.map(attributeSubValue => {
```
eunjae-lee: I've got a question.
When optionalFilters has a nested array, does it mean the values are used with 'OR' operator?

image

For example, that's equal to the following?

> attribute=value AND (attribute1=value OR attribute2=value)

I couldn't find the explanation from [here](https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/?language=javascript).

francoischalifour: When we use the array syntax, the same rules as `facetFilters` apply. You can read the [usage notes](https://www.algolia.com/doc/api-reference/api-parameters/facetFilters/#usage-notes).

In your example, it would get translated to:

```
attribute:value AND (attribute1:value OR attribute2:value)
```

We're manipulating optional filters here, so whenever something matches, it gets ranked higher.

eunjae-lee: Thanks for the link! I thought I saw it somewhere but couldn't find it 👍

packages/react-instantsearch-core/src/connectors/connectConfigureRelatedItems.ts
francoischalifour: Done in #2884.

Issue #867, opened 2018-01-17
NumericMenu items ignore the value "0" in their bounds


bgelineau:
**Do you want to request a *feature* or report a *bug*?**

a bug

**Bug: What is the current behavior?**

MultiRange items ignore the value 0 in their bounds. In the following fiddle, we would expect the "Free" option to return nothing and not every item: https://jsfiddle.net/ar80rgpc/33/ This also impacts ranges with negative values as start:0 will return items with negative values.

**Bug: What is the expected behavior?**

To handle the 0 value differently than a missing value.

**Bug: What browsers are impacted? Which versions?**

all

**What is the version you are using? Always use the latest one before opening a bug issue.**

react-instantsearch@4.4.1

levic92:
I am experiencing a similar issue in my connected range slider component. When I have a min value of 0 and try to refine the max value it won't actually filter correctly, but if min is set to anything but 0 it will work. This even occurs when setting a default refinement, if the min is 0 it won't actually work (but if min is set to 1 it filters correctly).

Haroenv:
This is still reproducible with NumericMenu; the replacement of MultiRanges: https://codesandbox.io/s/9oxqnnrkno

jhalborg:
Ran into this bug recently, and would very much appreciate a fix!

jhalborg:
Ran into this bug recently, and would very much appreciate a fix! 

(PS: Was encouraged to comment here [overhere](https://discourse.algolia.com/t/numericmenu-start-0-end-0-doesnt-work-for-me/9095/12?u=jm1) )

PR #2882, opened 2019-12-11
fix(connectNumericMenu): support numeric refinement 0

tkrugg:
**Summary**

This PR adds in connectNumericMenu support for numericRefinement `0`. 
There has been complaints about the following not producing a refinement

```jsx
import { NumericMenu } from 'react-instantsearch-dom';

<NumericMenu
  attribute="price"
  items={[
    { label: '<= 0', start: 0 },
    /* ... */
  ]}
/>

```

**Result**

fixes https://github.com/algolia/react-instantsearch/issues/867

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectNumericMenu.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectNumericMenu.js
@@ -120,6 +120,37 @@ describe('connectNumericMenu', () => {
         currentRefinement: '',
         canRefine: true,
       });
+
+      props = connect.getProvidedProps(
+        {
+          items: [
+            { label: 'is 0', start: 0, end: 0 },
+            { label: 'in 0..10', start: 0, end: 10 },
+          ],
+          contextValue,
+        },
+        {},
+        { results }
+      );
+      expect(props).toEqual({
+        items: [
+          {
+            label: 'is 0',
+            value: '0:0',
+            isRefined: false,
+            noRefinement: false,
+          },
+          {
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-core/src/connectors/__tests__/connectNumericMenu.js
```diff
@@ -279,6 +279,34 @@ describe('connectNumericMenu', () => {
         '>=': [100],
         '<=': [200],
       });
+
+      params = connect.getSearchParameters(
```
Haroenv: I feel like this should be its own test, what do you  think?

tkrugg: I felt the same way, I wasn't sure how to name it without breaking up the whole test thing into tests like:
```js
 it('refines the numeric facet for 100:200', () = > {
      params = connect.getSearchParameters(
        initSP,
        { attribute: 'facet', contextValue },
        { multiRange: { facet: '100:200' } }
      );
      expect(params.getNumericRefinements('facet')).toEqual({
        '>=': [100],
        '<=': [200],
      });
  }
```

francoischalifour:
Could you add a story with an appropriate Algolia index?

packages/react-instantsearch-core/src/connectors/__tests__/connectNumericMenu.js
Haroenv: `refines the corresponding numeric facet` and `refines the corresponding numeric facet with value 0`

PR #2890, opened 2020-01-15
feat(algoliasearch): add support for algoliasearch v4

yannickcr:
**Summary**

This change add support for algoliasearch v4 to react-instantsearch.

Some changes were needed on cache hydration for the server-side rendering. The test suite must still be updated to support both v3 and v4 versions of algoliasearch client.

We also added a new job to CircleCI to test it against the version 3 of the client.


packages/react-instantsearch-native/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch": "^3.27.1",
+    "algoliasearch": "^4.0.0-beta.14",
```
nunomaduro: This should be something like `>= 3.27.1 < 5`.

Haroenv: it can stay on v3, but install v4 on CI. That's still possible :)

examples/react-native-query-suggestions/package.json
```diff
@@ -20,7 +20,7 @@
     "react-test-renderer": "16.8.6"
   },
   "dependencies": {
-    "algoliasearch": "3.35.1",
+    "algoliasearch": "4.0.0-beta.14",
```
Haroenv: do we really want to ship examples with the beta version? I'm not sure it's fully stable, wdyt?

packages/react-instantsearch-native/package.json
yannickcr: Right, also it is weird to have it as a direct dependency since it is not used in the package :thinking:

examples/react-native-query-suggestions/package.json
nunomaduro: To be honest `4.0.0-beta.14` is more stable than 3.35.1. 😄

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -296,14 +321,43 @@ export default function createInstantSearchManager({
...
+    // search on it first.
+    if (client.transporter) {
+      const baseMethod = client.search;
+      client.search = (requests, ...methodArgs) => {
+        const requestsWithSeriazizedParams = requests.map(request => ({
+          ...request,
+          params: serializeQueryParameters(request.params),
+        }));
+
+        return client.transporter.responsesCache.get(
```
Haroenv: responsesCache can be undefined

examples/react-native-query-suggestions/package.json
Haroenv: With stable I mean: few changes

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
nunomaduro: @Haroenv We have a `if (client.transporter)` to ensure we on the `v4` where the responsesCache can't be undefined.

Haroenv: unless someone has a transporter which is an empty object as their custom client. It's not likely, but technically possible. Up to us to then decide if we can assume people won't do that

nunomaduro: I would say that scenario will never happen. What are your thoughts @yannickcr?

yannickcr: I agree that it is unlikely to happen. But I can rework the conditions here to make it clearer.

What is missing here is a proper way to know if we are using the `official` client and to get it's version. For now we detect if `client.addAlgoliaAgent` exists or not (does not exists -> `custom`) and we planned to add a `version` property to the client `>= v4` (does not exists -> `<= v4`) to replace the check on `client.transporter`. We could put a more obscure property to avoid any unintentional clash like `__ALGOLIA_CLIENT_VERSION` (a Symbol() would work perfectly for this but IE11...)

nunomaduro: We could have a `name` property on the client with the value `algoliasearch`.

`if (client.name === 'algoliasearch' && client.version.charAt(0) === 4)`

packages/react-instantsearch-native/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch": "^3.27.1",
+    "algoliasearch": ">= 3.27.1 < 5",
```
francoischalifour: Was this a mistake already? We don't use this dependency in this package.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
yannickcr: That's also a possibility

nunomaduro: Let's improve this later, I think for now it's fine to keep this.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -38,6 +38,31 @@ const sortIndexWidgetsFirst = (firstWidget, secondWidget) => {
   return 0;
 };
 
+// This function  is copy/pasted from the algoliasearch v4 API Client. Once modified,
```
```suggestion
// This function is copied from the algoliasearch v4 API Client. If modified,
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -38,6 +38,31 @@ const sortIndexWidgetsFirst = (firstWidget, secondWidget) => {
...
+// consider update it also here: `serializeQueryParameters` from `@algolia/transporter`.
```
```suggestion
// consider updating it also in `serializeQueryParameters` from `@algolia/transporter`.
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -313,9 +367,39 @@ export default function createInstantSearchManager({
...
+            (acc, result) => acc.concat(result.rawResults),
+            []
+          ),
+        }
+      );
+      return;
+    }
+
+    // Algoliasearch API Client < v4
+    // Prior to client v4 we didn't had a proper API to hydrate the client
```
```suggestion
    // Prior to client v4 we didn't have a proper API to hydrate the client
```

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -343,9 +427,29 @@ export default function createInstantSearchManager({
...
+          ],
+        },
+        {
+          results: results.rawResults,
+        }
+      );
+      return;
+    }
+    // Algoliasearch API Client < v4
+    // Prior to client v4 we didn't had a proper API to hydrate the client
```
```suggestion
    // Prior to client v4 we didn't have a proper API to hydrate the client
```

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
```diff
@@ -68,6 +68,12 @@ describe('createInstantSearchManager', () => {
         _cache: true, // cache is not enabled by default inside Node
       });
 
+      // Skip this test with Algoliasearch API Client >= v4
```
francoischalifour: I feel like we should adapt the tests for client v4 usage, here we don't test anything related to SSR working with RIS + Client v4.

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -5,7 +5,11 @@ import createInstantSearchManager from '../core/createInstantSearchManager';
 import { InstantSearchProvider, InstantSearchContext } from '../core/context';
 import { Store } from '../core/createStore';
 import { PlainSearchParameters, SearchParameters } from 'algoliasearch-helper';
-import { MultiResponse } from 'algoliasearch';
+import { SearchResponse } from '@algolia/client-search';
```
nunomaduro: Ups - is this code work on the v3 too?

packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
yannickcr: Yeah I agree. We should write an integration test here I believe.

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
yannickcr: I changed it to be closer to what we do in InstantSearch.js, same for the test with the client v3 on the CI.

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -5,7 +5,26 @@ import createInstantSearchManager from '../core/createInstantSearchManager';
 import { InstantSearchProvider, InstantSearchContext } from '../core/context';
 import { Store } from '../core/createStore';
 import { PlainSearchParameters, SearchParameters } from 'algoliasearch-helper';
-import { SearchResponse } from '@algolia/client-search';
+
+import algoliasearch, {
```
Haroenv: I feel like this block fits better in its own file in /types (if we have one), like in InstantSearch.js

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           name: Install algoliasearch v3
           command: |
             yarn remove -W @algolia/client-search
-            yarn add -W @types/algoliasearch@3.34.8 algoliasearch@3.35.1
```
Haroenv: don't we need to test typescript on v3 with the right type?

nunomaduro: Good catch.

yannickcr: No sure to understand here, what to you mean?

Haroenv: If we're testing v3 types, then we need to install @types/algoliasearch, to have a correct setup I think

yannickcr: Ho ok. I added the types to the setup to test with algoliasearch v3.

I also had to remove the examples before that, since they are relying on algoliasearch v4 we had both algoliasearch v3 and v4 in the lockfile, messing with the test.

PR #2894, opened 2020-01-29
chore(examples): make html & css of next example valid

Haroenv:
**Summary**

Invalid html elements were used here, which worked as divs, but is not valid.

I've also moved the static folder inside /public, since there was a warning: https://err.sh/zeit/next.js/static-dir-deprecated

**Result**

- css etc. is in the public folder
- removed invalid favicons
- update html to be valid


examples/next/components/head.js
```diff
@@ -15,10 +15,7 @@ export const Head = props => (
       content={props.description || defaultDescription}
     />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <link rel="icon" sizes="192x192" href="/static/touch-icon.png" />
-    <link rel="apple-touch-icon" href="/static/touch-icon.png" />
-    <link rel="mask-icon" href="/static/favicon-mask.svg" color="#49B882" />
-    <link rel="icon" href="/static/favicon.ico" />
+    <link rel="icon" href="/favicon.ico" />
```
Haroenv:
```suggestion
```

this is optional, so let's remove it too.

PR #2899, opened 2020-03-05
fix(deps): fix "too much recursion" error with circular deps

francoischalifour:
Since React InstantSearch 6, there's an issue with props equality with circular dependencies because of the dep `fast-deep-equal` that was introduced.

Errors like "too much recursion" appeared in browsers (mostly Firefox). More and more users [face this issue](https://discourse.algolia.com/t/instantsearch-conditional-display-too-much-recursion/9555), and I also faced this issue in some apps I built.

I tried upgrading the `fast-deep-equal` dependency to [`^3.1.0`](https://github.com/epoberezkin/fast-deep-equal/releases) which supposedly fixes this, but the build seems broken since I get an error about no default exports exist in this package.

I tried [`react-fast-compare`](https://github.com/FormidableLabs/react-fast-compare) and although it's a bit bigger, it fixes the issue. I tried in the Next app that the user sent.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -15864,6 +15864,11 @@ react-fast-compare@^2.0.2:
   resolved "https://registry.yarnpkg.com/react-fast-compare/-/react-fast-compare-2.0.4.tgz#e84b4d455b0fec113e0402c329352715196f81f9"
   integrity sha512-suNP+J1VU1MWFKcyt7RtjiSWUjvidmQSlqu+eHslq+342xCbGTYmC0mEhPCOHxlW0CywylOC1u2DFAT+bv4dBw==
 
+react-fast-compare@^3.0.0:
+  version "3.0.1"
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -15864,6 +15864,11 @@ react-fast-compare@^2.0.2:
   resolved "https://registry.yarnpkg.com/react-fast-compare/-/react-fast-compare-2.0.4.tgz#e84b4d455b0fec113e0402c329352715196f81f9"
   integrity sha512-suNP+J1VU1MWFKcyt7RtjiSWUjvidmQSlqu+eHslq+342xCbGTYmC0mEhPCOHxlW0CywylOC1u2DFAT+bv4dBw==
 
+react-fast-compare@^3.0.0:
```
Haroenv: how come nothing is removed from the lockfile? Does another package already use it?

francoischalifour: Yes, Storybook.

such:
Could this be released?

Haroenv:
Thanks for the ping, 6.4.0 is released :)

PR #2913, opened 2020-04-08
feat(SearchBox): provide input element ref

### LINKED ISSUE #2910: Get React ref to input element in SearchBox
**Is your feature request related to a problem? Please describe 🙏**

I would like to be able to pass a React ref to a `SearchBox` in order to manage focus on its `form` element.

**Describe the solution you'd like 🤔**

`SearchBox` currently has 
https://github.com/algolia/react-instantsearch/blob/478346916ec176ba978576759eb5d5c37b0dc4ee/packages/react-instantsearch-dom/src/components/SearchBox.js#L85-L86

 which can be used to get at the ref attached to the input element:
https://github.com/algolia/react-instantsearch/blob/478346916ec176ba978576759eb5d5c37b0dc4ee/packages/react-instantsearch-dom/src/components/SearchBox.js#L253-L254

However, given that it doesn't appear in the documentation, I don't want to rely on it.

It would be great if this was made part of the api.

**Describe alternatives you've considered ✨**

I tried using `focusShortcuts`, but I want to use `/` to focus search and `?` as a separate shortcut.  Unfortunately the search box gets focused in both cases as `?` is `shift+/`.

I can create a custom component or use `connectSearchBox`, but the widget works perfectly out of the box - except for this one issue.

ojeytonwilliams:
**Summary**

This makes it clear that `inputRef` is part of the api, so it can be used to get a React ref to the search input.  Closes #2910

**Result**

```jsx
<SearchBox
     inputRef={input => {this.input = input;}}
/>

focusSearch() {
  if(this.input) this.input.focus();
}
```

### DIFF
--- a/packages/react-instantsearch-dom/src/components/SearchBox.js
+++ b/packages/react-instantsearch-dom/src/components/SearchBox.js
@@ -84,9 +84,7 @@ class SearchBox extends Component {
 
     inputRef: PropTypes.oneOfType([
       PropTypes.func,
-      PropTypes.shape({
-        current: PropTypes.instanceOf(HTMLInputElement),
-      }),
+      PropTypes.exact({ current: PropTypes.object }),
     ]),
   };
 
--- a/packages/react-instantsearch-dom/src/components/SearchBox.js
+++ b/packages/react-instantsearch-dom/src/components/SearchBox.js
@@ -82,8 +82,7 @@ class SearchBox extends Component {
     isSearchStalled: PropTypes.bool,
     showLoadingIndicator: PropTypes.bool,
 
-    // For testing purposes
-    __inputRef: PropTypes.func,
+    inputRef: PropTypes.func,
   };
 
   static defaultProps = {
@@ -133,8 +132,8 @@ class SearchBox extends Component {
 
   onInputMount = input => {
     this.input = input;
-    if (this.props.__inputRef) {
-      this.props.__inputRef(input);
+    if (this.props.inputRef) {
+      this.props.inputRef(input);
     }
   };
 
--- a/packages/react-instantsearch-dom/src/components/SearchBox.js
+++ b/packages/react-instantsearch-dom/src/components/SearchBox.js
@@ -82,7 +82,12 @@ class SearchBox extends Component {
     isSearchStalled: PropTypes.bool,
     showLoadingIndicator: PropTypes.bool,
 
-    inputRef: PropTypes.func,
+    inputRef: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.shape({
+        current: PropTypes.instanceOf(HTMLInputElement),
+      }),
+    ]),
   };
 
   static defaultProps = {
@@ -132,8 +137,11 @@ class SearchBox extends Component {
 
   onInputMount = input => {
     this.input = input;
-    if (this.props.inputRef) {
+    if (!this.props.inputRef) return;
+    if (typeof this.props.inputRef === 'function') {
       this.props.inputRef(input);
+    } else {
+      this.props.inputRef.current = input;
     }
   };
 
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-dom/src/components/SearchBox.js
```diff
@@ -82,8 +82,7 @@ class SearchBox extends Component {
     isSearchStalled: PropTypes.bool,
     showLoadingIndicator: PropTypes.bool,
 
-    // For testing purposes
-    __inputRef: PropTypes.func,
+    inputRef: PropTypes.func,
```
francoischalifour, line 85: Since we expose this prop now, users can use new React APIs like [`createRef`](https://reactjs.org/docs/react-api.html#reactcreateref) or [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref) that do not return functions—they will therefore get a PropType error. This should be more correct now:

```suggestion
    inputRef: PropTypes.oneOfType([
      PropTypes.func,
      PropTypes.shape({ current: PropTypes.instanceOf(HTMLInputElement) }),
    ]),
```

ojeytonwilliams:
Thanks @Haroenv!  Let me know if there's anything else you need from me.

Haroenv:
Why does `focusShortCuts` need to be improved? It's a useful feature no?

packages/react-instantsearch-dom/src/components/SearchBox.js
ojeytonwilliams: The ref's already created inside this `SearchBox` here:

https://github.com/algolia/react-instantsearch/blob/26b8c682dde4587bdbf70caae6121f22e95b49cb/packages/react-instantsearch-dom/src/components/SearchBox.js#L252-L253

via 

https://github.com/algolia/react-instantsearch/blob/26b8c682dde4587bdbf70caae6121f22e95b49cb/packages/react-instantsearch-dom/src/components/SearchBox.js#L133-L138

So users should just provide a callback that will get passed the ref.  

If it would be better to ask users to create and pass in refs, I can do that instead.

francoischalifour: I see, thanks for the context. Yeah, we should support native React refs so that the API is as intuitive as using a basic React component.

francoischalifour:
@Haroenv `keyboardShortcuts`' limitations are similar to the ones in the current Autocomplete.js version: you cannot use key combinaisons (see initial issue #2910).

Now that we introduce this new `inputRef` option, it brings a lot of new possibilities, including better user-land support of keyboard shortcuts.

We now have multiple ways to achieve the same result, with `focusShortcuts` being more limiting. This would go in our "reduce API surface area" effort.

packages/react-instantsearch-dom/src/components/SearchBox.js
ojeytonwilliams: @francoischalifour The latest commit should cover both cases.

One thing I was not sure about: `HTMLInputElement`, is that going to be okay for server side rendering, even though it's a browser global?  I tried simply copying this PR's version `SearchBox` into a Gatsby project I'm working on and ran into `WebpackError: ReferenceError: HTMLInputElement is not defined`.  Is that dealt with somehow?

ojeytonwilliams: The CI was complaining about `HTMLInputElement`, so l investigated.  Turns out `React.createRef` just makes: `{ current: null }`, so I changed the PropType to look for, basically, that.

francoischalifour: That looks good—thanks for the investigation!

ojeytonwilliams:
No problem, thanks for the reviews and merge!

PR #2920, opened 2020-04-27
fix(createBoundingBoxFromHits): include check in the hit property "_geoloc"

afgoulart:
fix createBoundingBoxFromHits
i included a check in the createBoundingBoxFromHits function when lists hits has one his or many hits with property "_geoloc" as array.

**Summary**

**Result**

### DIFF
--- a/packages/react-instantsearch-dom-maps/src/Provider.js
+++ b/packages/react-instantsearch-dom-maps/src/Provider.js
@@ -62,15 +62,9 @@ class Provider extends Component {
     const { google } = this.props;
 
     const latLngBounds = hits.reduce((acc, hit) => {
-      if (
-        !hit._geoloc ||
-        (Array.isArray(hit._geoloc) && hit._geoloc.lenght === 0)
-      ) {
-        return acc.extend({});
-      }
+      const geoLocs = Array.isArray(hit._geoloc) ? hit._geoloc : [hit._geoloc];
 
-      const _geoloc = Array.isArray(hit._geoloc) ? hit._geoloc[0] : hit._geoloc;
-      return acc.extend(_geoloc);
+      return geoLocs.reduce((bounds, current) => bounds.extend(current), acc);
     }, new google.maps.LatLngBounds());
 
     return {
--- a/packages/react-instantsearch-dom-maps/src/Provider.js
+++ b/packages/react-instantsearch-dom-maps/src/Provider.js
@@ -62,7 +62,12 @@ class Provider extends Component {
     const { google } = this.props;
 
     const latLngBounds = hits.reduce(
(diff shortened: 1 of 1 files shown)


afgoulart:
image

packages/react-instantsearch-dom-maps/src/Provider.js
```diff
@@ -61,10 +61,17 @@ class Provider extends Component {
...
-    );
+    const latLngBounds = hits.reduce((acc, hit) => {
+      if (
+        !hit._geoloc ||
+        (Array.isArray(hit._geoloc) && hit._geoloc.lenght === 0)
+      ) {
+        return acc.extend({});
+      }
+
+      const _geoloc = Array.isArray(hit._geoloc) ? hit._geoloc[0] : hit._geoloc;
```
Haroenv, line 72: you  likely want to extend for every item in _geoloc if it's an array

Haroenv:
it's expected that e2e tests don't run here, since you're an external contributor, and otherwise that api key would leak

packages/react-instantsearch-dom-maps/src/Provider.js
```diff
@@ -61,10 +61,17 @@ class Provider extends Component {
...
+      return acc.extend(_geoloc);
```
Haroenv, line 73:
```suggestion
      const geoLocs = Array.isArray(hit._geoloc) ? hit._geoloc : [hit._geoloc];

      return geoLocs.reduce((bounds, current) => bounds.extend(current), acc);
```

I think this will be more correct, can you check if it behaves as you expect for you?

afgoulart: Thanks! It's this that I want

Haroenv: perfect, I'll let you edit the pull request, and it would also be good if you  could add a test for this, thanks!

Haroenv:
Hey @afgoulart do you think you could write a new test case like the ones here:

https://github.com/algolia/react-instantsearch/blob/065208c1d75d794706faee7c5a3aef4d6e380322/packages/react-instantsearch-dom-maps/src/__tests__/Provider.js#L47

It would be the same as the existing ones, except with an array in geoloc, or a an array in one and a single in another element

Thanks!

SavageWilliam:
would love this to be merged :)

sarahdayan:
@afgoulart We're in the process of moving `algolia/react-instantsearch` to our InstantSearch monorepo. This PR won't be transferrable, so I'll close it here, but please feel free to reopen it in (algolia/instantsearch.js)[https://github.com/algolia/instantsearch.js].

PR #2921, opened 2020-04-27
feat(infinite-hits): support cache

eunjae-lee:
**Summary**

**Result**


packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -29,6 +29,27 @@ function getCurrentRefinement(props, searchState, context) {
...
+function getStateWithoutPage(state) {
+  const { page, ...rest } = state;
+  return rest;
+}
+
+function getInMemoryCache() {
+  let cachedHits = undefined;
+  let cachedState = undefined;
+  return {
+    read: ({ state }) => {
```
Haroenv, line 41:
```suggestion
    read({ state }) {
```

eunjae-lee: Any reason?

Haroenv: arrow functions add a layer of .bind in babel in some cases IIRC. At least as a regular function there's less characters

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/index.js
```diff
@@ -0,0 +1,3 @@
+export {
+  default as getInfiniteHitsSessionStorageCache,
```
eunjae-lee, line 2: Any better name? 😟

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2921/commits/0f8bf1ef036d53495b29645439ec2e27b91d9737

eunjae-lee:
I didn't know well about browser's scroll restoration.

If you take a look at this example,
- [code](https://codesandbox.io/s/funny-pasteur-pw5jj?file=/src/App.js)
- [try it!](https://pw5jj.csb.app/)

It's a very simple example and the scroll restoration doesn't work there.

```jsx
  useEffect(() => {
    setResult(true);
  }, []);

  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      {result && (
           ....
```

I need to research more, but from my understanding, the scroll is restored right after the first dom rendering, which breaks this current example which renders the big chunk of dom elements at the second render.

For the same reason, when going back to a page with InstantSearch, it renders after /queries API. So the scroll is not restored.

This might initiate a new RFC which saves the current search result at `beforeunload` or something like that, and later restore it without triggering API call. Without this, when going back to InstantSearch, the page will flash and the scroll restoration needs to be done manually.

At the moment, we ship this as is and document how to manually restore scroll position, which is still better than before.

What do you think?

eunjae-lee:
I talked to Haroen, and he said there's a render before the search api call. So I'll try to render the InfiniteHits with the cache at the first render. Then the scroll might be restored correctly.

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
```diff
@@ -0,0 +1,49 @@
...
+      try {
+        window.sessionStorage.setItem(
+          KEY,
+          JSON.stringify({
+            state: getStateWithoutPage(state),
+            hits,
+          })
+        );
+      } catch (error) {
+        console.error(error); // eslint-disable-line no-console
```
yannickcr, line 45: Logging the error in the console is not useful, the user can't do anything with it like catching it or handling it differently. He will just get an error message without any possibility to fix/suppress it.

Since in that case the behavior is like having no cache I'm more to do nothing here (no `console.error`).

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/index.js
yannickcr: Since the function instantiate our cache object I would have gone with the `create` prefix instead of `get`: `createInfiniteHitsSessionStorageCache`

francoischalifour: @yannickcr I thought the same at first until I realize that it doesn't create it, but _retrieve_ it. The `create` prefix gets confusing when you see the function getting called at each render, because it doesn't actually create the cache:

```jsx
function Hits() {
  return (
    <InfiniteHits
      hitComponent={Product}
      cache={createInfiniteHitsSessionStorageCache()}
    />
  );
}
```

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
```diff
@@ -0,0 +1,49 @@
...
+      if (!hasSessionStorage()) {
+        return null;
+      }
+      try {
+        const cache = JSON.parse(window.sessionStorage.getItem(KEY));
+        return isEqual(cache.state, getStateWithoutPage(state))
+          ? cache.hits
+          : null;
+      } catch (_error) {
+        return null;
```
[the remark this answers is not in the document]
Haroenv: it could also mean that `getItem` throws an error, so deleting from the cache is a good idea, however doing it in a finally or similar could throw an error again, right?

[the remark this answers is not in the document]
eunjae-lee: Similarly to what Haroen said, there can be an environment where sessionStorage is not available at all like incognito mode of Safari.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -29,6 +29,35 @@ function getCurrentRefinement(props, searchState, context) {
...
+        : null;
+    },
+    write({ state, hits }) {
+      cachedState = getStateWithoutPage(state);
+      cachedHits = hits;
+    },
+  };
+}
+
+function flattenHits(cachedHits) {
```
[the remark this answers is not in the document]
eunjae-lee: Agree that "flatten" is a bit misleading.

```js
this._cachedHits = cache.read({ state: searchState }) || {};
...
extractHitsFromCache(this._cachedHits)
```

It's actually not exactly extracting hits from "cache", but "cachedHits" which is an object form. Any better suggestion? Like `extractFromCachedHits`?

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -29,6 +29,35 @@ function getCurrentRefinement(props, searchState, context) {
   return currentRefinement;
 }
 
+function getStateWithoutPage(state) {
```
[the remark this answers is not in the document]
eunjae-lee: exporting helper functions
vs
expose them as parameters like `read({ state, getStateWithoutPage, isEqual }) { ... }`

What do you think of the latter?

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
```diff
@@ -0,0 +1,49 @@
+import isEqual from 'react-fast-compare';
+
+function getStateWithoutPage(state) {
+  const { page, ...rest } = state;
+  return rest;
+}
+
+const KEY = 'ais.infinitehits';
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2921/commits/d0daddb0cac2ae6a14deb2187f84a95ec3949b57

examples/infinite-hits-cache/README.md
```diff
@@ -0,0 +1,26 @@
+This example shows how to synchronize your instantsearch url if you are using react-router.
```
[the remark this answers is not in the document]
eunjae-lee: Nice catch https://github.com/algolia/react-instantsearch/pull/2921/commits/eaa8d29b041a456af10b2cc2cbc52e2523a20448

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
eunjae-lee: It totally makes sense. Thanks for the feedback!
https://github.com/algolia/react-instantsearch/pull/2921/commits/fcd9bea0c31dd4f12fa7e62b9491732a91238349

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/index.js
yannickcr: For the SessionStorage it does not create the cache itself (since it is handled by the browser), but this is IMO a detail. It will still recreate the object to manipulate it (`read`/`write`).

Same apply for `getInMemoryCache` where it will fully re-create the cache on call.

eunjae-lee: Interesting. That makes sense too.
I feel like by using `get` as a prefix, we imply this is a light operation. `create` is somewhat too heavy.

But do we want to guide users to do like the following? Or, is the current way `cache={getInfiniteHitsSessionStorageCache()}` is still fine?

```js
const cache = getInfiniteHitsSessionStorageCache();

function Hits() {
  return (
    <InfiniteHits
      hitComponent={Product}
      cache={cache}
    />
  );
}
```

What do you think?

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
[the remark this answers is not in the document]
francoischalifour: That can't happen at this stage because [we already check for `sessionStorage` in the lines above](https://github.com/algolia/react-instantsearch/blob/900519a568a3bc0457424d7c34d73f6655853fda/packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js#L20-L22).

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
francoischalifour: I wouldn't add them to the API because then it's hard to remove them since it becomes a feature.

We don't need to export them externally though, just share them internally for now.

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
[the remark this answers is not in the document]
eunjae-lee: Even if it exists, `QuotaExceededError` can be thrown: https://stackoverflow.com/a/18861484

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
eunjae-lee: you're right. makes sense.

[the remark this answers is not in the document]
eunjae-lee: Okay, now I remember why I didn't gather `getStateWithoutPage` in one file in the first place.

There is an in-memory cache implementation inside `connectInfiniteHits` which belongs to `-core` package.

There is a session storage implementation which is exposed as `getInfiniteHitsSessionStorageCache` from `-dom` package.

If we were to share `getStateWithoutPage`, then the place where it lives would be `-core`. Then we need to expose it so that it can be used in `-dom` package.

For a function with two lines, it looked like an overkill to me.

What do you think?

eunjae-lee:
> Is the new example supposed to be accessible at [this address](https://deploy-preview-2921--react-instantsearch.netlify.app/examples/infinite-hits-cache)?

@francoischalifour it's not published at the moment.
To try it out, you need to check out this branch and run the following:

```
yarn install && yarn build
yarn workspace example-infinite-hits-cache start
```

I thought of removing this example before merging this PR, because I made this as a simple PoC. Do you think it's something we can keep in the repo?

And I see there are bunch of examples in `/examples` folder but not published, for example `geo-search`, `multi-index`, etc. Do you know any history why we only publish only a few examples and not the others?

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/index.js
yannickcr: > I feel like by using get as a prefix, we imply this is a light operation. create is somewhat too heavy.

Well this is a very subjective feeling :wink: 

What I mean is that this `getX` function is still creating something new in all cases, not just getting something.

```js
getInfiniteHitsSessionStorageCache() === getInfiniteHitsSessionStorageCache() // false, for a `getX` function I would expect this to be true
```

My two cents, this is still a small naming question, not a big deal :slightly_smiling_face:

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
francoischalifour: Right, let's skip this then!

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
francoischalifour: Although long, I think that's explicit: `extractHitsFromCachedHits`.

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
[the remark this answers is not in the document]
francoischalifour: Got it! Should we still try to recover from a parsing issue here? Try to delete the cache, otherwise fail silently. Otherwise we might stay with a broken cache "forever".

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2921/commits/032318ab94d29dff38886003a1115992ccd0910e

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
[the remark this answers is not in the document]
eunjae-lee: I think JSON parse error won't likely happen unless user directly writes wrong data to sessionStorage with the correct key `ais.infiniteHits`. In our implementation, we write 

```js
window.sessionStorage.setItem(
  KEY,
  JSON.stringify({
    state: getStateWithoutPage(state),
    hits,
  })
)
```

I think it's okay we don't deal with error of low probability that we won't cause.

eunjae-lee:
fixing tests

eunjae-lee:
Is "argos" expected to run, or not?
Should I do something to force-run the task?

Haroenv:
We don't rely on Argos anymore, it was too flaky :)

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/__tests__/sessionStorage.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  it('returns null if the state is different', () => {
+    const cache = getInfiniteHitsSessionStorageCache();
+    const state = { q: 'hello' };
+    const newState = { q: 'world' };
+    const hits = { 1: ['a', 'b', 'c'] };
+    cache.write({ state, hits });
+    expect(cache.read({ state: newState })).toBeNull();
+  });
```
Haroenv, line 43: I'd like to see a test where the cache (mock) throws an error on creation or on access, do you think that makes sense?

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/index.js
Haroenv: Yes, I'd go with `create` too

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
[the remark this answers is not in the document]
Haroenv: JSON error might still happen if the local storage itself truncates the json for example. In that case we can consider the cache as corrupted and try as if the cache didn't exist

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -29,6 +29,35 @@ function getCurrentRefinement(props, searchState, context) {
...
+    write({ state, hits }) {
+      cachedState = getStateWithoutPage(state);
+      cachedHits = hits;
+    },
+  };
+}
+
+function extractHitsFromCachedHits(cachedHits) {
+  return Object.keys(cachedHits)
+    .sort()
```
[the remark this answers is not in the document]
eunjae-lee: That's a good point!

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/__tests__/sessionStorage.js
eunjae-lee: I was assuming user passes a correct cache object with correct methods (read, write) with exception handling. That's why I try - catched inside `sessionStorage`. Do you want to handle exceptions on the connector level?

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/sessionStorage.js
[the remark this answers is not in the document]
eunjae-lee: Huh, I didn't know sessionStorage truncates data (but not remove). https://github.com/algolia/react-instantsearch/pull/2921/commits/4f191a2b354469c27e9d3288632c09ca72ac45fc

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/index.js
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2921/commits/d705bd900320fa861d6e2c7462f1aa77281eb7be

packages/react-instantsearch-dom/src/lib/infiniteHitsCache/__tests__/sessionStorage.js
Haroenv: It's the cache which catches the error, right?

eunjae-lee: Yes, but I'm not sure if I understand your first comment.

> where the cache (mock) throws an error on creation or on access

```jsx
<InfiniteHits
  hitComponent={Product}
  cache={createInfiniteHitsSessionStorageCache()}
/>
```

On creation -> `createInfiniteHitsSessionStorageCache()` call?


```js
# connectInfiniteHits

this._cachedHits = cache.read({ state: searchState }) || {};
```
On access -> `.read()` or `.write()` ?

Haroenv: What I meant is to write some tests for cases where the underlying session storage fails at different moments (creation, read, invalid json) with this cache wrapper

eunjae-lee: ah you meant sessionStorage. got it.

eunjae-lee: wdyt?
https://github.com/algolia/react-instantsearch/pull/2921/commits/89aedc6a4ca5a2e4442a8746a18fd1ee737e6435

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -120,8 +156,13 @@ export default createConnector({
   },
 
   refine(props, searchState, event, index) {
-    if (index === undefined && this._lastReceivedPage !== undefined) {
-      index = this._lastReceivedPage + 1;
+    const pages = Object.keys(this._cachedHits || {}).map(Number);
```
Haroenv, line 159: If we are storing elements with a number index in an object, what about using an array to store that instead? I guess it's no big difference either way

eunjae-lee: I initially started with an object because user could've reached the page with params like `page=3`. Still in that case, an array can be used, though.

PR #2927, opened 2020-05-18
chore: add codesandbox ci

eunjae-lee:
**Summary**

This PR configures [CodeSandbox CI app](https://github.com/apps/codesandbox). 

When a PR is created, the app prepares a sandbox with the build from the branch.
image
The comment gets updated with the new url as the PR is updated.

image

This app doesn't give us a way to install the temporary build on other environments whereas Pika CI does.

image

In my opinion, we heavily use CodeSandbox and might not need to `yarn install` out of CodeSandbox.

What do you think? If you want to see Pika CI in action, I can open another PR to demonstrate it. Or, if you all are satisfied with CodeSandbox CI, then we can merge this PR 🙂

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,0 +1,3 @@
+{
+  "packages": ["packages/*"]
+}
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,6 @@
 {
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "sandboxes": [
+    "github/algolia/create-instantsearch-app/tree/templates/react-instantsearch"
+  ]
 }


Haroenv:
@eunjae-lee can we customise which sandbox gets shown? maybe one of the doc-code-samples (getting started) or Create InstantSearch App would be good?

eunjae-lee:
@Haroenv yes, it's possible and I'm going to try to configure it :)

eunjae-lee:
image

CompuIves:
First of all, awesome to see that you're using CodeSandbox CI! 🎉 

One remark is that you can also install the builds from CodeSandbox CI locally! For this build you can do:
```
yarn add https://pkg.csb.dev/algolia/react-instantsearch/commit/ecd04305/react-instantsearch-core
``` 

Let me know if you have any other questions!

eunjae-lee:
@CompuIves Thanks! I didn't know it was possible.
I think you could also add that command to the comment so that whenever a pr is updated, we can just copy the updated command. (because the hash will change) What do you think? 🙂

CompuIves:
If you click the "here" link you can copy the install instructions from the details page here (under "Local Install Instructions"): 

image

Does that also work? I want to keep the comment as small as possible, to not pollute the thread too much 😄.

eunjae-lee:
Ahh I totally missed it. Thanks. It's enough for me 😁

francoischalifour:
The CodeSandbox CI gets the `version` from `package.json`, which results in sending the User Agent `react-instantsearch (6.4.0)` in the PR sandbox. This can cause wrong analytics. Is there a way to pass the commit hash as the version?

eunjae-lee:
@francoischalifour since we read the version from `packages/react-instantsearch-core/src/core/version.js`, and use it to build user agent string, we should modify `version.js` to avoid it. We could add a script and configure CodeSandbox CI to run something like `yarn run set_test_ver && yarn build`.

But I don't think this will have any meaningful impact on the analytics. It's just us trying things a couple of times.

francoischalifour:
This new setup will allow us the share fixes and custom builds to customers, which is why I was raising this point. That can definitely wait though!

PR #2941, opened 2020-06-16
WIP: remove second instance of "query" in the response "params" for SSR

### LINKED ISSUE #2940: Hydration of search client uses `query` from response instead of `request`
**Describe the bug 🐛**

This isn't usually a problem, except when a query rule changes the query (e.g. `voltaren` to `voltaren®`). This causes the cache to be missed & a request on the frontend to me made

I'd still need to check this, but I think this is only relevant for combining with `algoliasearch@4`

**To Reproduce 🔍**

1. go to an index which has an edited query (e.g. latency/instant_search for "edit_me")
2. enter that query
3. reload
4. see a query was done frontend, while it shouldn't be

This can be checked in the official next demo: https://codesandbox.io/s/github/algolia/react-instantsearch/tree/master/examples/next

**Expected behavior 💭**

No query should be done, even if the query done by the user doesn't match up with the query in the response / from the engine

**Screenshots 🖥**

Screenshot 2020-06-15 at 17 39 23

Screenshot 2020-06-15 at 17 39 10

**Additional context**

Relevant code is:

https://github.com/algolia/react-instantsearch/blob/7e883df4b39144efe7da9d2a713658075d337028/packages/react-instantsearch-core/src/core/createInstantSearchManager.js#L322-L474

Haroenv:
fixes #2940

So far this is a pretty ugly option. I wonder if we can maybe make the cache ignore a second parameter? (that's probably not possible, since the key is a string, not an object)

- [ ] try out in more cases
- [ ] test performance (params might be LARGE)
- [ ] write tests

### DIFF
--- a/packages/react-instantsearch-core/src/core/createInstantSearchManager.js
+++ b/packages/react-instantsearch-core/src/core/createInstantSearchManager.js
@@ -319,6 +319,19 @@ export default function createInstantSearchManager({
     }
   }
 
+  function removeDuplicateQuery(params) {
+    let previousIndex = undefined;
+    const queryParamRegex = /&?query=[^&]*/g;
+    return params.replace(queryParamRegex, function replacer(match, offset) {
+      if (previousIndex && offset > previousIndex) {
+        previousIndex = offset;
+        return '';
+      }
+      previousIndex = offset;
+      return match;
+    });
+  }
+
   function hydrateSearchClient(client, results) {
     if (!results) {
       return;
@@ -385,7 +398,7 @@ export default function createInstantSearchManager({
                 acc.concat(
                   result.rawResults.map(request => ({
                     indexName: request.index,
-                    params: request.params,
+                    params: removeDuplicateQuery(request.params),
                   }))
                 ),
               []
@@ -414,7 +427,7 @@ export default function createInstantSearchManager({
           acc.concat(
             result.rawResults.map(request => ({
               indexName: request.index,
-              params: request.params,
+              params: removeDuplicateQuery(request.params),
             }))
           ),
         []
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -319,6 +319,19 @@ export default function createInstantSearchManager({
     }
   }
 
+  function removeDuplicateQuery(params) {
+    let previousIndex = undefined;
+    const queryParamRegex = /&?query=[^&]*/g;
+    return params.replace(queryParamRegex, function replacer(match, offset) {
+      if (previousIndex && offset > previousIndex) {
```
8kto, line 326:
```suggestion
      if (typeof previousIndex !== 'undefined' && offset > previousIndex) {
```
Prevents from sneaked zero index (`params = "query=test"`)

Haroenv:
```suggestion
      if (previousIndex !== undefined && offset > previousIndex) {
```

denpolischuk:
Checked it on our project - seems to solve the issue.

Haroenv:
Thanks for testing this @denpolischuk, I'd advice for now, since this is a big issue for your project that you stay on the version with this fix, so that we can fix this issue for everyone at a non-rushed pace. WDYT?

denpolischuk:
@Haroenv what is the ETA of this fix rollout in "non-rushed" mode?

Haroenv:
So far this is something that we are doing between priorities, and are still discussing with the backend team if there's a better longterm solution, so it could unfortunately take a month or so. In the mean time, if you don't want to use a fork, I've had really good experiences with [patch-package](https://github.com/ds300/patch-package), we're not planning to do any different changes to this part of the code, so maintaining that will be very straightforward

PR #2945, opened 2020-06-23
fix(ssr): remove second instance of "query" in the response "params" for SSR

### LINKED ISSUE #2940: Hydration of search client uses `query` from response instead of `request`

Haroenv:
**Summary**

We use the `params` of the response given by the Algolia query to craft a cache key for the query done by server-side rendering. However, when the engine has e.g. a query rule which edits the query, this `params` won't be the same as the one actually queried. It will have `query=original...&query=edited`, which means that we should remove the second instance of the query to make the cache not miss

**Result**

When rendering a query which is edited by a query rule, no query gets done frontend. This can be tested by the query `edit_me` on the `instant_search` index.

closes #2941
fixes #2940

### DIFF
--- a/packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
+++ b/packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
@@ -13,11 +13,14 @@ import { findResultsState } from '../createInstantSearchServer';
 Enzyme.configure({ adapter: new Adapter() });
 
 describe('findResultsState', () => {
-  const createSearchClient = () => ({
+  const createSearchClient = responseParams => ({
     search: requests =>
       Promise.resolve({
         results: requests.map(({ indexName, params: { query } }) => ({
           query,
+          params: responseParams
+            ? responseParams(query)
+            : `query=${encodeURIComponent(query)}`,
           index: indexName,
         })),
       }),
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -13,11 +13,14 @@ import { findResultsState } from '../createInstantSearchServer';
 Enzyme.configure({ adapter: new Adapter() });
 
 describe('findResultsState', () => {
-  const createSearchClient = () => ({
+  const createSearchClient = responseParams => ({
```
francoischalifour, line 16: Having a signature like the following would make it easier to understand:

```js
createSearchClient({ transformResponseParams } = {})
```

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -59,9 +59,36 @@ const getSearchParameters = (indexName, searchParameters) => {
...
+/**
+ * The engine can return params: "query=xxx&query=yyy" if e.g. a query rule modifies it.
+ * This however will cause us to miss the cache hydration, so we make sure to keep
+ * only the first query (always the one from the parameters).
+ */
+function removeDuplicateQuery(params) {
+  let previousIndex = undefined;
+  const queryParamRegex = /&?query=[^&]*/g;
+  return params.replace(queryParamRegex, function replacer(match, offset) {
+    if (previousIndex !== undefined && offset > previousIndex) {
```
eunjae-lee, line 71: Can `offset > previousIndex` be false when `previousIndex !== undefined`?

Haroenv: if there's three instances of `query` that would be possible, so I left that condition to be sure

eunjae-lee: When you meet the first instance of `query`, `previousIndex` gets assigned (no longer undefined). After that, no matter how many more instances of `query` occurs, we can just return an empty string, IIUC.
Meaning, we can have just a boolean instead of `previousIndex` to store the offset.

Haroenv: good call, fixed

PR #2946, opened 2020-06-24
Add auto-updating minzip size to Readme

websocket98765:


Haroenv:
Thanks for your PR, but we don't think adding the size in a badge is the most efficient use of space, as there are other places (bundlephobia for example) that give more detailed information rather than just a number

Issue #2875, opened 2019-11-27
The connectStateResults hoc injected props allSearchResults behaviour is different between v5.7.0 and v6.0.0


hongchanhk01:
**Context**

We have a component will render no results placeholder when all the all of the indexes have no results.

We achieve the feature by using `connectStateResults` HOC. After upgrade to v6 from v5.7, we can't access index search results.

**Describe the bug 🐛**

The `connectStateResults` hoc injected props `allSearchResults` behavior is different between v5.7.0 and v6.0.0

In v5, the `allSearchResults` a new object reference will assigned when ever a new index be fetched. 

In v6, the `allSearchResults` object will updated by mutated

**To Reproduce 🔍**

checking out the repo that I prepare for this issue. Switch between branches to test out behavior.

A live example helps a lot! We have a simple online template for you to use for your explanations:

https://github.com/hongchanhk01/test-react-instantsearch

- master branch (v6.0.0)
- react-instantsearch-v5  (v5.7.0)

**Screenshot**
v5
![image](https://user-images.githubusercontent.com/44665484/69702450-37c24a80-112a-11ea-9af9-5ac1a91d6be8.png)

v6
![image](https://user-images.githubusercontent.com/44665484/69702369-0c3f6000-112a-11ea-9d32-a9ef93e17083.png)

PS: if you encounter file not found problem by the bundler when switching branch. Trying remove the `.cache` and start the server again

**Expected behavior 💭**

The `allSearchResults` props should not have been mutated. Instead it should given a new Object reference every time an index fetched. 

**Environment:**

 - OS: MacOS
 - Browser Chrome
 - Version 78

This happen in all browsers as well

Haroenv:
So far I indeed have been able to confirm this in your environment, but have not yet found what causes it. Thanks for the bug report and reproduction!

hongchanhk01:
@Haroenv 

I have a dirty quick fix on this. what do think  ?

https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/connectors/connectStateResults.js#L65 

```js
    return {
      searchState,
      searchResults: results,
      allSearchResults: Object.assign({}, searchResults.results), // Quick fixes here
      searching: searchResults.searching,
      isSearchStalled: searchResults.isSearchStalled,
      error: searchResults.error,
      searchingForFacetValues: searchResults.searchingForFacetValues,
      props,
    };
```

[...]

PR #2953, opened 2020-07-16
fix(core): appending successful index search results by returning new object reference

SecretBase:
Fix https://github.com/algolia/react-instantsearch/issues/2875

So basically,

The index search results are from the `store` and because the new search result added by mutating the results object.

At the `createConnector.tsx`

https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/core/createConnector.tsx#L342

The `providedProps.allSearchResults` spread into the `Composed` component but since the object `providedProps. allSearchResults` reference did not change. React did not trigger re-render the `Composed` Component.

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js
@@ -145,6 +145,97 @@ describe('createInstantSearchManager with multi index', () => {
     );
   });
 
+  // https://github.com/algolia/react-instantsearch/issues/2875
+  it('update resultsState without mutate object', async () => {
+    const ism = createInstantSearchManager({
+      indexName: 'first',
+      initialState: {},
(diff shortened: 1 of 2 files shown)


SecretBase:
@Haroenv 

Hope you can help on reviewing this PR when you are free.

Haroenv:
This is very interesting, thanks! I'd like to have a test which avoids this issue in the future, do you have an idea how to write that?

SecretBase:
> This is very interesting, thanks! I'd like to have a test which avoids this issue in the future, do you have an idea how to write that?

I might need some time to understand how `createInstantSearchManager.js` works.

Just an idea.
using `Object.freeze` to freeze the initial state and pass it to the store.
Then asserting no error throw when changing the state ? I guess ?

SecretBase:
@Haroenv 

Adding unit test for this case. Hope the test make sense.

PR #2961, opened 2020-08-03
fix(ssr): allow "params" to be optional in custom clients

### LINKED ISSUE #2958: Next.js SSR - TypeError: Cannot read property 'replace' of undefined - version: 6.7.0
**Describe the bug 🐛**

Seems like the version `6.7.0` has a bug related to results state when implementing it by server-side rendering in Next.js applications. https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/

**To Reproduce 🔍**

1. Update the package to `6.7.0`.
2. Implement `findResultsState` on `getInitialProps` or `getServerSideProps` functions.

```
TypeError: Cannot read property 'replace' of undefined
    at removeDuplicateQuery (/home/paneladm/projects/paneladm-dev/medina-app/node_modules/react-instantsearch-dom/dist/cjs/core/createInstantSearchServer.js:83:17)
    at /home/paneladm/projects/paneladm-dev/medina-app/node_modules/react-instantsearch-dom/dist/cjs/core/createInstantSearchServer.js:96:15
```

## Environment that buged

- node: 10.20.1
- npm: 6.14.4
- algoliasearch: `4.3.1`
- react-instantsearch: `6.7.0`

### Environment that worked

- node: 10.20.1
- npm: 6.14.4
- algoliasearch: `4.3.0`
- react-instantsearch: `6.6.0`

### Live example on Codesandbox.io

A live example working: https://codesandbox.io/s/nextjs-instantsearch-ohphp

Haroenv:
**Summary**

A custom client could _not_  return `params`. In this case we can keep `params` as-is instead of trying to `replace`.

**Result**

If `params`  is falsy, it doesn't get transformed. This will also add a  small boost for empty query, although a replace on empty string already probably was very fast.

fix #2958

### DIFF
--- a/packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
+++ b/packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
@@ -262,6 +262,32 @@ describe('findResultsState', () => {
         );
       });
 
+      it('with custom client, without "params"', async () => {
+        const Connected = createWidget();
+
+        const App = props => (
+          <InstantSearch {...props}>
+            <Connected />
+          </InstantSearch>
+        );
+
+        const props = {
+          ...requiredProps,
+          searchClient: createSearchClient({
+            transformResponseParams() {
+              return undefined;
+            },
+          }),
+          searchState: {
+            query: 'iPhone',
+          },
+        };
+
+        const data = await findResultsState(App, props);
+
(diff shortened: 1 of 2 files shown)


Haroenv:
This is the sandbox that actually is fixed by this PR: https://codesandbox.io/s/nextjs-instantsearch-2jozv?file=/modules/instantsearch/index.js

PR #2973, opened 2020-10-02
fix(ssr): make sure metadata is available on initial render

### LINKED ISSUE #2972: CurrentRefinements widget doesn't render during ssr
**To Reproduce 🔍**

1. load https://ploxl.sse.codesandbox.io/?refinementList%5Bcategories%5D%5B0%5D=Appliances&page=1&configure%5BhitsPerPage%5D=12

https://codesandbox.io/s/crazy-wright-ploxl

This seems to happen because `metadata` is empty in any render on the server (inside `connectCurrentRefinements > getProvidedProps`)

**Expected behavior 💭**

current refinements shows in the html, like it does once the results are in

**Screenshots 🖥**

Untitled 2020-10-02 11_07_05

Haroenv:
**Summary**

Metadata isn't available on initial render/ server rendered page. This is because `registerWidget` is added only in componentDidMount, as well as only registering its change to instantsearchManager asynchronously (to prevent many requests if multiple widgets are added at once, which is normal on the first render).

This is preventing CurrentRefinements from rendering notably (only widget which uses metadata)

**Result**

metadata is caught using the same method as searchParameters, and injected via resultsState.

CurrentRefinements renders as expected in SSR.

fixes #2972

### DIFF
--- a/packages/react-instantsearch-core/src/core/createConnector.tsx
+++ b/packages/react-instantsearch-core/src/core/createConnector.tsx
@@ -113,6 +113,12 @@ export function createConnectorWithoutContext(
             this.props
           );
         }
+
+        if (isWidget) {
+          this.unregisterWidget = this.props.contextValue.widgetsManager.registerWidget(
+            this
+          );
+        }
       }
 
       componentDidMount() {
@@ -123,12 +129,6 @@ export function createConnectorWithoutContext(
             });
           }
         });
-
-        if (isWidget) {
-          this.unregisterWidget = this.props.contextValue.widgetsManager.registerWidget(
-            this
-          );
-        }
       }
 
       shouldComponentUpdate(nextProps, nextState) {
--- a/packages/react-instantsearch-core/src/core/createConnector.tsx
+++ b/packages/react-instantsearch-core/src/core/createConnector.tsx
@@ -115,8 +115,9 @@ export function createConnectorWithoutContext(
         }
 
         if (isWidget) {
+          const isServer = true; // TODO: find a real value for this
           this.unregisterWidget = this.props.contextValue.widgetsManager.registerWidget(
-            this
+            this, isServer
           );
         }
       }
--- a/packages/react-instantsearch-core/src/core/createConnector.tsx
+++ b/packages/react-instantsearch-core/src/core/createConnector.tsx
@@ -110,14 +110,8 @@ export function createConnectorWithoutContext(
               ais: this.props.contextValue,
               multiIndexContext: this.props.indexContextValue,
             },
-            this.props
-          );
-        }
-
-        if (isWidget) {
-          const isServer = true; // TODO: find a real value for this
-          this.unregisterWidget = this.props.contextValue.widgetsManager.registerWidget(
-            this, isServer
+            this.props,
+            connectorDesc.getMetadata && connectorDesc.getMetadata.bind(this)
           );
         }
       }
@@ -130,6 +124,12 @@ export function createConnectorWithoutContext(
             });
           }
         });
+
+        if (isWidget) {
+          this.unregisterWidget = this.props.contextValue.widgetsManager.registerWidget(
+            this
+          );
+        }
       }
 
       shouldComponentUpdate(nextProps, nextState) {
--- a/packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
+++ b/packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
@@ -2,7 +2,11 @@ import React, { Component, Children } from 'react';
 import isEqual from 'react-fast-compare';
 import PropTypes from 'prop-types';
 import createInstantSearchManager from '../core/createInstantSearchManager';
-import { InstantSearchProvider, InstantSearchContext } from '../core/context';
+import {
+  InstantSearchProvider,
+  InstantSearchContext,
+  IndexContext,
+} from '../core/context';
 import { Store } from '../core/createStore';
 import { PlainSearchParameters, SearchParameters } from 'algoliasearch-helper';
 import { MultiResponse } from '../types/algoliasearch';
@@ -17,20 +21,17 @@ type ResultsState = {
 type InstantSearchManager = {
   store: Store;
   widgetsManager: any;
-  getWidgetsIds: any;
-  getSearchParameters: (
+  getWidgetsIds(): any;
+  getSearchParameters(
     ...args: any[]
-  ) => {
-    mainParameters: SearchParameters;
-    derivedParameters: SearchParameters;
-  };
-  onSearchForFacetValues: (...args: any[]) => any;
-  onExternalStateUpdate: (...args: any[]) => any;
+  ): { mainParameters: SearchParameters; derivedParameters: SearchParameters };
+  onSearchForFacetValues(...args: any[]): any;
+  onExternalStateUpdate(...args: any[]): any;
   transitionState: any;
   updateClient: any;
   updateIndex: any;
-  clearCache: () => void;
-  skipSearch: any;
+  clearCache(): void;
+  skipSearch(...args: any[]): any;
 };
 
 type SearchClient = {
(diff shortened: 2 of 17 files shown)


packages/react-instantsearch-core/src/core/createConnector.tsx
```diff
@@ -113,6 +113,13 @@ export function createConnectorWithoutContext(
             this.props
           );
         }
+
+        if (isWidget) {
+          const isServer = true; // TODO: find a real value for this
```
Haroenv, line 118: this condition might be something related to `process`, but I'm not sure what bundle size impacts that will have

packages/react-instantsearch-dom/src/components/CurrentRefinements.js
```diff
@@ -49,7 +49,7 @@ export const CurrentRefinements = ({
 const itemPropTypes = PropTypes.arrayOf(
   PropTypes.shape({
     label: PropTypes.string.isRequired,
-    value: PropTypes.func.isRequired,
+    value: PropTypes.func,
```
Haroenv, line 52: initial render on client won't include `value`, since that's a function

eunjae-lee:
The changes make sense to me, but I'm not proficient enough with react ssr to approve this PR.
I'd love to get others' eyes here.

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -22,40 +23,43 @@ const createSearchParametersCollector = accumulator => {
   };
 };
 
-const getSearchParameters = (indexName, searchParameters) => {
-  const sharedParameters = searchParameters
-    .filter(searchParameter => !hasMultipleIndices(searchParameter.context))
+const getSearchParameters = (indexName, widgets) => {
```
samouss, line 26: The function used to only deal with `SearchParameters`.  We have introduced the computation of the metadata which doesn't make sense. We could create a dedicated function to derive the metadata from the collected values.

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -180,5 +189,8 @@ export const findResultsState = function(App, props) {
     helper,
     sharedParameters,
     derivedParameters
-  );
+  ).then(results => {
+    results[0].metadata = metadata;
+    return results;
+  });
```
samouss, line 195: IIRC this shape is opaque for the consumer of `react-instantsearch` i.e. we are the only once that interact with the value. In that case I would use a more appropriate structure (i.e. an object) rather than assigning the value to the first element of the array.

packages/react-instantsearch-dom/src/components/CurrentRefinements.js
samouss: Rather than lower the validation we could assign a noop function to the metadata when we hydrate them.

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -246,14 +248,15 @@ class InstantSearch extends Component<Props, State> {
     }
   }
 
-  onSearchParameters(getSearchParameters, context, props) {
+  onSearchParameters(getSearchParameters, context, props, getMetadata) {
```
samouss, line 251: The signature becomes a bit ugly. IMO it would be simpler to introduce a new prop/function with a simpler signature. We can then adapt the current implementation on the new one without BC. Here is an example:

```js
class InstantSearch {
  onSearchParameters(widget, context, props) {
    if (this.props.onSearchParameters) {
      this.props.onSearchParameters(widget.getSearchParameters, context, props, searchState);
    }
  }

  onFindBetterNameThanThis(widget, context, props) {
    if (this.props.onFindBetterNameThanThis) {
      this.props.onFindBetterNameThanThis(widget, context, props, searchState);
    }
  }
}
```

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
Haroenv: that's an interesting option. I think I remember seeing people loop over the results since they always knew it's an array, but usually those use cases didn't really make sense to me

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
Haroenv: I didn't rename it because there's a _lot_ of people who get stuck up on forgetting to pass `onSearchParameters` down to React InstantSearch, I feel like if we change it again, it might be even worse

eunjae-lee: cc @yannickcr @francoischalifour

eunjae-lee: What if we warn or throw if `onSearchParameters` is missing in the context of SSR?

Haroenv: these detections of server side rendering often cause accidental polyfills, not sure how to detect a prop is not passed only then

eunjae-lee: Hmm I didn't think of accidentail polyfills. But that snippet might be okay, because it's directly coming from the internal code of React: https://github.com/facebook/react/blob/1eb2b892dfffedfdb57039ef715fc7049f4d538a/packages/shared/ExecutionEnvironment.js
but again, I'm not too sure.

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -150,28 +161,38 @@ export const findResultsState = function(App, props) {
...
-    <App
-      {...props}
-      onSearchParameters={createSearchParametersCollector(searchParameters)}
-    />
+    <App {...props} widgetsCollector={createWidgetsCollector(widgets)} />
   );
 
+  if (widgets.length === 0) {
+    throw new Error(
+      '[ssr]: no widgets were added, you likely did not pass the `widgetsCollector` down to the InstantSearch component.'
```
eunjae-lee, line 172: Nice. I like this 👍

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
Haroenv: https://github.com/algolia/react-instantsearch/pull/2977 here's that alternative

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
```diff
@@ -246,7 +256,15 @@ class InstantSearch extends Component<Props, State> {
...
 
-  onSearchParameters(getSearchParameters, context, props) {
+  onSearchParameters(
+    getSearchParameters: ConnectorDescription['getMetadata'],
+    context: {
+      ais: InstantSearchContext;
+      multiIndexContext: IndexContext;
+    },
+    props: object,
+    getMetadata?: ConnectorDescription['getMetadata']
```
eunjae-lee, line 266: I see `getMetadata` being optional here, but `widgetsCollector` is expecting it to be required.

Haroenv: good call, it's optional in both places

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -150,28 +161,38 @@ export const findResultsState = function(App, props) {
...
 
-  const searchParameters = [];
+  const widgets = [];
 
   renderToString(
-    <App
-      {...props}
-      onSearchParameters={createSearchParametersCollector(searchParameters)}
-    />
+    <App {...props} widgetsCollector={createWidgetsCollector(widgets)} />
```
eunjae-lee, line 167: Don't forget to update [this doc](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/) to adapt to `widgetsCollector`.

eunjae-lee (approved):
Looks good to me!

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
Haroenv: https://github.com/algolia/doc/pull/4992

packages/react-instantsearch-core/src/widgets/InstantSearch.tsx
Haroenv: actually ConnectorDescription already makes it optional, so this was duplicate

ItsWendell:
@Haroenv  Great work but sadly this broke SSR in my codebase using NextJS, this should have been marked as a breaking change according to semver but is not.

Also the example should have been updated too. https://github.com/algolia/react-instantsearch/tree/master/examples/next/pages

PR #2974, opened 2020-10-07
poc: add hooks

eunjae-lee:
**Summary**

This is a dirty PoC just to play with hooks.

This doesn't even work as intended.
`useSearch()` returns an outdated things, but I'm not intending to dig into that, at least for now.

To try this,

- open [this storybook](https://5f7dba6c2a449700073c7f6f--react-instantsearch.netlify.app/storybook/?path=/story/hooks--default).
- type any character in the search box, then you will see `results` displayed.
- if you click the `Search for 'Amazon'` button, it will refine with `amazon` query.

### DIFF
--- a/packages/react-instantsearch-dom/src/hooks.js
+++ b/packages/react-instantsearch-dom/src/hooks.js
@@ -1,0 +1,16 @@
+import { useContext } from 'react';
+import { InstantSearchContext } from 'react-instantsearch-core';
+
+export function useSearch() {
+  const { store, widgetsManager } = useContext(InstantSearchContext);
+  const getResults = () => store.getState().results;
+  const searchBox = widgetsManager
+    .getWidgets()
+    .find(widget => widget.name.startsWith('AlgoliaSearchBox'));
+  const setQuery = query => searchBox && searchBox.refine(query);
+
+  return {
+    getResults,
+    setQuery,
+  };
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch-dom/src/hooks.js
```diff
@@ -0,0 +1,16 @@
+import { useContext } from 'react';
+import { InstantSearchContext } from 'react-instantsearch-core';
+
+export function useSearch() {
+  const { store, widgetsManager } = useContext(InstantSearchContext);
+  const getResults = () => store.getState().results;
+  const searchBox = widgetsManager
+    .getWidgets()
+    .find(widget => widget.name.startsWith('AlgoliaSearchBox'));
+  const setQuery = query => searchBox && searchBox.refine(query);
```
Haroenv, line 10: I wonder if there should be a setSearchState where you can add query yourself; this doesn't use multi-index yet I think

eunjae-lee: Yeah, we should expose `searchState` and `setSearchState`. This was just to see if it works :)

eunjae-lee:
I'm planning to work on the hook support. I'm closing this PR because it was a quick & dirty branch back then. I'll open a new one.

PR #2975, opened 2020-10-07
fix: add missing dependencies

merceyz:
**Summary**

Added `react-fast-compare` as a dependency to `react-instantsearch-dom` because it's using it without declaring it and `algoliasearch` as a peer dependency to `react-instantsearch-core` because it inherits it from `algoliasearch-helper` so it needs to be declared

**Result**

Building using Yarn 2's PnP mode succeeds.
Been declared like this for months in https://github.com/yarnpkg/berry/blob/9e5ea238cf4363dd95c43111494db8bfbdb06ccd/.yarnrc.yml#L60-L66

### DIFF
--- a/packages/react-instantsearch-core/package.json
+++ b/packages/react-instantsearch-core/package.json
@@ -44,6 +44,7 @@
     "react-fast-compare": "^3.0.0"
   },
   "peerDependencies": {
+    "algoliasearch": ">= 3.1 < 5",
     "react": ">= 16.3.0 < 17"
   }
 }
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
this makes sense. Really algoliasearch should be an optional peer, but I don't think it's possible to declare it like that. For now it indeed sounds better as a peer dependency

merceyz:
If the intent is for it to be optional, I can change it to that

Haroenv:
if you mark it as optional, we'd need to change that in the helper too I think

francoischalifour (approved):
Looks good, thanks!

We need to find a solution for the `import/no-extraneous-dependencies` ESLint rule's limitations that would have caught the missing `react-fast-compare` dep: https://github.com/algolia/react-instantsearch/blob/14622260f6789ff96b1f08b72bc984a6b7a85bac/.eslintrc.js#L14-L17

merceyz:
You can run your tests through Yarn PnP, it would have (and did) catch this

PR #2976, opened 2020-10-14
feat. request: expose some utilities in createInstantSearchServer for wider SSR-usage

jrmattila:
**Summary**
This is more like a feature request than finalized PR.

Currently findResultsState can be used only by giving it a component tree to render. Often you already have similar kind of server-side rendering logic to gather data for hydration. For example, with Apollo Client you might be using getDataFromTree to extract cache.

By exposing these functions you can get the results state at the same time with other such renderings.

**Result**

```
const searchParams = [];
const searchParamsCollector = createSearchParametersCollector(acc);
const data = await getDataFromTree(
  <App
    {...appProps}
    searchParamsCollector={searchParamsCollector}
    Component={Component}
    apolloState={apolloState}
    apollo={apollo}
  />
 );
algoliaResultsState = await getResultsStateFromSearchParams(
  indexName,
  client,
  searchParams
);
```

### DIFF
--- a/packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
+++ b/packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
@@ -10,7 +10,7 @@ const getIndexId = context =>
     ? context.multiIndexContext.targetedIndex
     : context.ais.mainTargetedIndex;
 
-const createSearchParametersCollector = accumulator => {
+export const createSearchParametersCollector = accumulator => {
   return (getWidgetSearchParameters, context, props, searchState) => {
     accumulator.push({
       getSearchParameters: getWidgetSearchParameters,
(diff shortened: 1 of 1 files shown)


Haroenv:
This is a very interesting idea, rendering multiple times on server when you also have eg Apollo is indeed a waste. I'll wait until the consensus on #2973 is found whether we change the name of onSearchParameters or not

joonasmattila:
> This is a very interesting idea, rendering multiple times on server when you also have eg Apollo is indeed a waste. I'll wait until the consensus on #2973 is found whether we change the name of onSearchParameters or not

Thanks for taking this into consideration.

Also, if we could have some utility-function that generates resultsState based on search-parameters and indexName, could we live without renderToString?

Haroenv:
We would always need to know which widgets are mounted to know what the search parameters + other widget state is, how do you see the translation from search parameters?

joonasmattila:
Yeah, I figured that widgets have their effect on initial results state. Just to clarify, with search-parameters I was actually meaning url query-paremeters like in this [long url to demo page](https://react-instantsearch.netlify.app/examples/default-theme/?refinementList%5Bcategories%5D%5B0%5D=Audio&refinementList%5Bcategories%5D%5B1%5D=Cell%20Phone%20Accessories&page=1&configure%5BhitsPerPage%5D=16&hierarchicalMenu%5BhierarchicalCategories.lvl0%5D=Cell%20Phones%20%3E%20Cell%20Phone%20Accessories%20%3E%20Data%20Cables&range%5Brating%5D%5Bmin%5D=3&range%5Brating%5D%5Bmax%5D=5)

We are eventually aiming to lazy-load most of the refining widgets anyway. So they would be dynamically loaded not until on client and also excluded from server-side rendering. I don't know yet if this is a good idea at all(?) 

Anyway, it would be nice to be able to translate query-parameters directly to resultsState so we could show items based on that. Then ultimately a new query would be made on client when all the widgets are mounted.

Haroenv:
hmm, the widgets would need to be mounted for that, since they are the ones translating from searchState to search parameters

PR #2977, opened 2020-10-14
fix(ssr): cleaner result for multi-index

Haroenv:
**Summary**

storing metadata in the first result is confusing, so we change the shape to always be an object; be it multi-index or single index

stacked PR on top of #2973

**Result**

result of findResultsState is now an object with the keys `{ metadata, results }` instead of `results` and metadata being in the first element

### DIFF
--- a/packages/react-instantsearch-core/src/core/createInstantSearchManager.js
+++ b/packages/react-instantsearch-core/src/core/createInstantSearchManager.js
@@ -364,8 +364,8 @@ export default function createInstantSearchManager({
       };
     }
 
-    if (Array.isArray(results)) {
-      hydrateSearchClientWithMultiIndexRequest(client, results);
+    if (Array.isArray(results.results)) {
+      hydrateSearchClientWithMultiIndexRequest(client, results.results);
       return;
     }
 
@@ -478,8 +478,8 @@ export default function createInstantSearchManager({
       return null;
     }
 
-    if (Array.isArray(results)) {
-      return results.reduce(
+    if (Array.isArray(results.results)) {
+      return results.results.reduce(
         (acc, result) => ({
           ...acc,
           [result._internalIndexId]: new algoliasearchHelper.SearchResults(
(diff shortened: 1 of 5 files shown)


packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -611,18 +611,12 @@ export default function createInstantSearchManager({
...
   }
-  if (Array.isArray(resultsState)) {
-    return resultsState[0].metadata;
-  }
-  return resultsState.metadata;
-}
 
-function hydrateMetadata(resultsState) {
-  return extractMetadata(resultsState).map(datum => ({
+  return resultsState.metadata.map(datum => ({
```
eunjae-lee, line 619: nice ❤️

packages/react-instantsearch-dom/src/core/createInstantSearchServer.js
```diff
@@ -202,7 +202,6 @@ export const findResultsState = function(App, props) {
     sharedParameters,
     derivedParameters
   ).then(results => {
-    results[0].metadata = metadata;
-    return results;
+    return { metadata, results };
```
eunjae-lee, line 205: 👍

eunjae-lee (approved):
Looks great to me. This looks way better than before.

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
```diff
@@ -364,8 +364,8 @@ export default function createInstantSearchManager({
       };
     }
 
-    if (Array.isArray(results)) {
-      hydrateSearchClientWithMultiIndexRequest(client, results);
+    if (Array.isArray(results.results)) {
```
francoischalifour, line 367: This duplication reflects that the naming is wrong, perhaps we should go with `stateResult.results` (like we return `{results} = await findResultState()`.

francoischalifour (commented):
Nice!

packages/react-instantsearch-core/src/core/createInstantSearchManager.js
Haroenv: ah, I didn't think of renaming the variable, that would be a good idea in the future

Issue #2964, opened 2020-08-11
Instantsearch server-side rendering is also fetching on the client


francisrod01:
**Describe the bug 🐛**

Maybe is because the mock request I'm doing for empty query. I don't know the cause.

**To Reproduce 🔍**

1. Go to https://sgk4r.sse.codesandbox.io/?query=amazon&page=1&configure%5BhitsPerPage%5D=4
2. Open the Chrome console and refresh the page
3. Go to `Network` tab and see the `queries` request made on the client

**A live example helps a lot! :100:**

> Your application on codesandbox does not seem to include the code from the guide, are you saying it's solved or not? thanks
_Originally posted by @Haroenv in https://github.com/algolia/react-instantsearch/issues/2960#issuecomment-671830911_

Here's a codesandbox: https://codesandbox.io/s/nextjs-algolia-instantsearch-sgk4r

**Expected behavior 💭**

A clean rendering ONLY on the server-side and see the HTML appropriately on the page.

**Screenshots 🖥**

![image](https://user-images.githubusercontent.com/4691768/89908659-83d3f300-dbc4-11ea-9a95-eebf52a66052.png)

**Environment:**

 - OS: `Linux debian 4.9.0-13-amd64`
 - Google Chrome Browser `Version 84.0.4147.125 (Official Build) (64-bit)`
- NPM: `6.14.4`
- Node.js: `10.20.1`
- next: `9.3.5`
- react: `16.13.1`
- algoliasearch: `4.2.0`
- react-instantsearch-dom: `6.6.0`

**Additional context**

- Here's the doc I followed: [algolia.com - building search UI - conditional requests][1]
- Related: issue #2960 (Error serializing `.resultsState.state` returned from `getServerSideProps` in "/search")

[1]: https://www.algolia.com/doc/guides/building-search-ui/going-further/conditional-requests/react/#mocking-a-search-request

Haroenv:
Hmm, I originally thought that this could be due to the fact that the custom client for conditional requests no longer has the cache functionality, and can't be hydrated. If that was the case, the solution would be:

```js
const searchClientSearch = searchClient.search;
searchClient.search = function search(requests) {
  if (requests.every(({ params }) => !params.query)) {
    return Promise.resolve({
      results: requests.map(() => ({
        hits: [],
        nbHits: 0,
        nbPages: 0,
        processingTimeMS: 0
      }))
    });
  }

  return searchClientSearch(requests);
};
```

[...]

PR #2987, opened 2020-11-09
Re #2964

callmephilip:
re https://github.com/algolia/react-instantsearch/issues/2964

Changes to the original Next example:

- updated deps
- filtering UI moved to https://github.com/algolia/react-instantsearch/compare/master...bakery:extra-api-call?expand=1#diff-9b8869972d19a24a39c1c70f7f97a746d277c27d869026fc21fb5a1baa1ea81c, wrapped with `connectStats`

### DIFF
--- a/examples/next/components/filters.js
+++ b/examples/next/components/filters.js
@@ -1,0 +1,13 @@
+import React from 'react';
+import {
+  RefinementList,
+  connectStats,
+} from 'react-instantsearch-dom';
+
+function Filters() {
+  return (
+    <RefinementList attribute="categories" />
+  );
+}
+
+export default connectStats(Filters);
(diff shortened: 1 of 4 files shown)


examples/next/components/filters.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {
+  RefinementList,
+  connectStats,
+} from 'react-instantsearch-dom';
+
+function Filters() {
+  return (
+    <RefinementList attribute="categories" />
```
tkrugg, line 9: Hey @callmephilip thanks for the contribution.
It's not clear to me why wrapping Filters with connectStats needed. Can you explain to us your usecase?

Nesting connectors (in this case `connecStats` & `connectRefinementList`) is in general not advisable because they can cause too many renderings.

sarahdayan:
@callmephilip We're in the process of moving `algolia/react-instantsearch` to our InstantSearch monorepo. This PR won't be transferrable, so I'll close it here, but please feel free to reopen it once the migration is done.

PR #2996, opened 2020-12-04
feat(answers): add `EXPERIMENTAL_Answers` widget

eunjae-lee:
**Summary**

This PR adds `answers` widget.
Since it's experimental, it's written in JS (not in TS) without a dedicated connector.
We will improve it later on once we figure out how to integrate `findAnswers` method better.

- example: https://codesandbox.io/s/react-instantsearch-app-forked-521nw?file=/src/App.js

**TODO**
- [x] Add basic tests

**In the upcoming PR**
- [ ] Applying findAnswers from helper
- [ ] Applying facets
- [ ] Remove `searchClient` from props
- [ ] Use index name from parent index widget
- [ ] loading state (`'idle' | 'loading' | 'stalled'`)


packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,81 @@
...
+    <p>loading...</p>
+  ) : (
+    hits && hits.length > 0 && <pre>{JSON.stringify(hits, null, 2)}</pre>
+  );
+}
+
+export default function Answers({
+  searchClient,
+  attributesForPrediction = ['*'],
+  queryLanguages = ['en'],
```
dustincoates, line 18: @eunjae-lee I suspect this may have come from the documentation? If so, the docs were incorrect on this point. We want `queryLanguages` to be set each time and not default to English.

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,81 @@
...
+import { instantSearchContext } from 'react-instantsearch-core';
+import { createConcurrentSafePromise } from '../lib/createConcurrentSafePromise';
+import { debounce } from '../lib/debounce';
+
+/* eslint-disable react/prop-types */
+function DefaultAnswersComponent({ isLoading, hits = [] }) {
+  return isLoading ? (
+    <p>loading...</p>
+  ) : (
+    hits && hits.length > 0 && <pre>{JSON.stringify(hits, null, 2)}</pre>
```
shortcuts, line 11: Doesn't it throw an error in case `(!hits || !hits.length) && !isLoading` ?

packages/react-instantsearch-dom/src/widgets/Answers.js
eunjae-lee: @dustincoates oh I see.
I got it from https://www.algolia.com/doc/rest-api/answers/#method-param-querylanguages
What about `attributesForPrediction = ['*']`. Is it correct?

packages/react-instantsearch-dom/src/widgets/Answers.js
eunjae-lee: If it's not loading (!isLoading) and there is no hits, then it will not render anything.
image

shortcuts: Oh yeah make sense!

packages/react-instantsearch-dom/src/widgets/Answers.js
dustincoates: @eunjae-lee we want it to be set to a specific language if at all possible

eunjae-lee:
Hey @Haroenv can you review this PR since you're already reviewing Answers widget in JS repo.
I couldn't find a proper access to helper or client, so I ended up making it like how it is now.

* How it's made without connector, subscribing context directly to get updates
* If it's okay to accept one `answersComponent` or if it should accept `headerComponent`, `loaderComponent`, ...
* What are the things you think are required for this PR to go live? (With the prefix `EXPERIMENTAL_` at the moment)

Haroenv:
that would be a new pattern in our library, experimental means more *this could change* than *this could be removed*. I see how that could be confusing, but prefixing with beta seems odd to me for a widget name

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,83 @@
...
+  );
+  const searchIndex = useMemo(() => searchClient.initIndex(index), [
+    searchClient,
+    index,
+  ]);
+  useEffect(() => {
+    // eslint-disable-next-line no-warning-comments
+    // FIXME: remove this customization once the engine accepts url encoded query params
+    if (searchClient.transporter) {
+      searchClient.transporter.userAgent.value = 'answers-test';
```
Haroenv, line 40: this needs to be set for just the answers query, not global like this

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,83 @@
...
+    const unsubcribe = context.store.subscribe(() => {
+      const { widgets, results } = context.store.getState();
+      setQuery(widgets.query);
+      setIndex(results && results.index);
+    });
+    return unsubcribe;
+  }, [context]);
+  const setDebouncedResult = useMemo(
+    () =>
+      debounce(result => {
```
Haroenv, line 53: it's the query that debounce, not the rendering, no?

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,83 @@
...
+  );
+}
+
+export default function Answers({
+  searchClient,
+  queryLanguages,
+  attributesForPrediction = ['*'],
+  nbHits = 1,
+  extraParameters = {},
+  answersComponent: AnswersComponent = DefaultAnswersComponent,
```
Haroenv, line 21:
```suggestion
  answersComponent: AnswersComponent = DefaultAnswersComponent,
  ...extraParameters = {},
```

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,83 @@
...
+    <p>loading...</p>
+  ) : (
+    hits && hits.length > 0 && <pre>{JSON.stringify(hits, null, 2)}</pre>
+  );
+}
+
+export default function Answers({
+  searchClient,
+  queryLanguages,
+  attributesForPrediction = ['*'],
```
Haroenv, line 18:
```suggestion
  attributesForPrediction,
```

Haroenv: let the engine set this default

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,83 @@
...
+  nbHits = 1,
+  extraParameters = {},
+  answersComponent: AnswersComponent = DefaultAnswersComponent,
+}) {
+  const context = useContext(instantSearchContext);
```
Haroenv, line 23: let's bundle up all these effects, and expose `useAnswers`

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,83 @@
...
+/* eslint-disable react/prop-types */
+function DefaultAnswersComponent({ isLoading, hits = [] }) {
+  return isLoading ? (
+    <p>loading...</p>
+  ) : (
+    hits && hits.length > 0 && <pre>{JSON.stringify(hits, null, 2)}</pre>
+  );
+}
+
+export default function Answers({
```
Haroenv, line 15: there's no tests at all?

packages/react-instantsearch-dom/src/widgets/Answers.js
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581#discussion_r556519074

what do you think?

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -64,6 +65,7 @@ export default function Answers({
     setIsLoading(true);
     runConcurrentSafePromise(
       searchIndex.findAnswers(_query, queryLanguages, {
+        ...extraParameters,
```
RaseMotte, line 68: @eunjae-lee  I wonder if this is not a bit confusing to have some parameters top level (`nbHits`, `query`, ect) and some nested in `extraParameter` (eg `params` and `threshold`). I wonder if it would make more sens to have them all at the same level ?

eunjae-lee: @RaseMotte 
The yesterday's version I sent you was 

```jsx
<Answers extraParameters={{
  threshold: xxx,
  params: {
    ...
  }
}} />
```

I'm thinking to change it to

```jsx
<Answers threshold={xxx} params={{
  filters: ...
  , whatever search parameters, ...
}} />
```

I guess this is also what you're talking about. What do you think?

packages/react-instantsearch-dom/src/widgets/Answers.js
eunjae-lee: Initially I wasn't sure and wanted to get some feedback on how this component is made (no connector, directly using instantSearchContext, ...). So I didn't add tests at that time. Do you see this component being released looking like this? If so, I can add tests now.

packages/react-instantsearch-dom/src/widgets/Answers.js
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2996/commits/194cddb48f2fb9b9a789a185e5efc4610c65759f

packages/react-instantsearch-dom/src/widgets/Answers.js
eunjae-lee: c0d2d573

packages/react-instantsearch-dom/src/widgets/Answers.js
Haroenv: I don't see huge blockers, as long as the hook also gets released :)

eunjae-lee:
> at the moment, the filter values are missing right?

Yeah right. I'd like to release them without filters first. And then later apply `findAnswers` method from helper which takes care of filters. (Same in IS.js)

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -12,13 +12,12 @@ function DefaultAnswersComponent({ isLoading, hits = [] }) {
...
-export default function Answers({
+function useAnswers({
   searchClient,
   queryLanguages,
   attributesForPrediction,
-  nbHits = 1,
-  answersComponent: AnswersComponent = DefaultAnswersComponent,
-  ...extraParameters
+  nbHits,
+  extraParameters,
```
Haroenv, line 31: I don't think we should expose "extraParameters", especially considering how it's `extraParameters.params` for search options to be passed. Let's spread here too 👍

eunjae-lee: Ah you're right! https://github.com/algolia/react-instantsearch/pull/2996/commits/cd530682856d8ffd29db202a61fe83f52710ee05

Haroenv:
let's at least add some basic tests here, but for me this is good (ask someone else on the team to review as well, since this are some big changes)

We should keep it as EXPERIMENTAL for now if we don't add the filters yet

dustincoates:
@Haroenv @eunjae-lee quick confirmation: when you say "if we don't add the filters yet" do you mean that you can't filter or that we aren't returning facet values? I'm guessing the latter since you can put filters in to the params, but want to be sure.

eunjae-lee:
> @Haroenv @eunjae-lee quick confirmation: when you say "if we don't add the filters yet" do you mean that you can't filter or that we aren't returning facet values? I'm guessing the latter since you can put filters in to the params, but want to be sure.

You can already pass filters to the `params` right now. What I meant was to pass the filters (that are currently applied to the search) automatically to answers API call.

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,70 @@
+import React, { useState, useEffect, useMemo, useContext } from 'react';
```
francoischalifour, line 1: This is the first time that we introduce React Hooks in the codebase and they're compatible from React 16.8.0. Our `react*` peer dependencies start from 16.3.0, which would throw here: https://github.com/algolia/react-instantsearch/blob/cd530682856d8ffd29db202a61fe83f52710ee05/packages/react-instantsearch-dom/package.json#L51-L52

Bumping our `react*` peer dependencies would be a major bump in React InstantSearch.

How do you plan to make it work in the current React InstantSearch version?

Haroenv: could we give a warning if this new widget is used in react pre 16.8? Since this is an experimental widget, there might be different tradeoffs

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -8,15 +8,17 @@ const cx = createClassNames('Answers');
...
+      {isLoading && <div className={cx('loader')}></div>}
+      {!isLoading && (
+        <ul className={cx('list')}>
+          {hits.map((hit, index) => (
+            <li key={index} className={cx('item')}>
+              {JSON.stringify(hit)}
+            </li>
+          ))}
+        </ul>
+      )}
```
shortcuts, line 21: I think one of these two implementations would be easier to read

```suggestion
      {(isLoading && <div className={cx('loader')}></div>) || (
        <ul className={cx('list')}>
          {hits.map((hit, index) => (
            <li key={index} className={cx('item')}>
              {JSON.stringify(hit)}
            </li>
          ))}
        </ul>
      )
```

```suggestion
      {isLoading ? (
        <div className={cx('loader')}></div>
      ) : (
        <ul className={cx('list')}>
          {hits.map((hit, index) => (
            <li key={index} className={cx('item')}>
              {JSON.stringify(hit)}
            </li>
          ))}
        </ul>
      )}
```

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2996/commits/2f8a79f618e2c57a3de732d456b95888f05f3861

what do you think?

packages/react-instantsearch-dom/src/widgets/__tests__/Answers.js
```diff
@@ -0,0 +1,101 @@
...
+import { render, fireEvent, waitFor } from '@testing-library/react';
+import MutationObserver from '@sheerun/mutationobserver-shim';
+import Answers from '../Answers';
+
+describe('Answers', () => {
+  let searchClient;
+  beforeEach(() => {
+    searchClient = algoliasearch(
+      'CKOEQ4XGMU',
+      '6560d3886292a5aec86d63b9a2cba447'
```
Haroenv, line 13: why not test this with a fake search client? that way it can't be flaky

eunjae-lee: Yeah I wanted to make sure if the test cases really work well. So I went with the real appId. Once I confirmed both the code and the test code work well, I wondered if I should replace it with the fake client. Maybe I should, otherwise it may fail.

Haroenv: sounds good for testing, but the final test should be a fake client I think

eunjae-lee: yup just updated it https://github.com/algolia/react-instantsearch/pull/2996/commits/3b5f980689bf728e7c114b1aabd90175b4088308

packages/react-instantsearch-dom/src/widgets/__tests__/Answers.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+    expect(container.querySelector('ul.ais-Answers-list')).toBeNull();
+  });
+
+  it('renders an answer', async () => {
+    // It's required for `waitFor` method.
+    // https://github.com/testing-library/dom-testing-library/releases/tag/v7.0.0
+    window.MutationObserver = MutationObserver;
+
+    const { container, getByText, getByPlaceholderText } = render(
```
shortcuts, line 83:
```suggestion
    const { getByText, getByPlaceholderText } = render(
```

This variable doesn't seems to be used

packages/react-instantsearch-dom/src/lib/debounce.ts
```diff
@@ -0,0 +1,39 @@
+// Debounce the function call to the trailing edge.
```
shortcuts, line 1: We could now use the typing from https://github.com/algolia/instantsearch.js/commit/3df3e8102bf72e310cd1bfbe4aa3823c30d24dbe

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,108 @@
...
+
+function hasReactHooks() {
+  // >= 16.8.0
+  const [major, minor] = React.version.split('.').map(Number);
+  if (major >= 17) {
+    return true;
+  }
+  if (major === 16 && minor >= 8) {
+    return true;
+  }
```
shortcuts, line 14:
```suggestion
  return Boolean(major >= 17 || (major === 16 && minor >= 8));
```

Not sure if that's better but it's a cute one-liner

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,108 @@
...
+  return false;
```
shortcuts, line 15: (repost because wrong lines were commented)

```suggestion
  return Boolean(major >= 17 || (major === 16 && minor >= 8));
```

Not sure if that's better but it's a cute one-liner.

packages/react-instantsearch-dom/src/lib/debounce.ts
eunjae-lee: done https://github.com/algolia/react-instantsearch/pull/2996/commits/65bda7dea20f1cfbc2d2fd22d22e188c2ad0dd10

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,102 @@
...
+  const debouncedSearch = useMemo(() => {
+    if (!searchIndex) {
+      // eslint-disable-next-line prefer-promise-reject-errors
+      return () => Promise.reject();
+    }
+    if (!searchIndex.findAnswers) {
+      throw new Error(
+        '`Answers` component and `useAnswers` hook require `algoliasearch` to be 4.8.0 or higher.'
+      );
+    }
```
Haroenv, line 48: this error should be thrown at line 39 instead of in the search call itself I think

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,102 @@
...
+    return debounceAsync(searchIndex.findAnswers, searchDebounceTime);
+  }, [searchIndex]);
+  useEffect(() => {
```
Haroenv, line 51: these hook calls all look a little cramped without newline between them, wdyt?

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,102 @@
...
+    setIndex(context.mainTargetedIndex);
+
+    const unsubcribe = context.store.subscribe(() => {
```
Haroenv, line 54: could directly return this

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,102 @@
...
+      const { widgets } = context.store.getState();
+      setQuery(widgets.query);
+    });
+    return unsubcribe;
+  }, [context]);
+  const setDebouncedResult = useMemo(
+    () =>
+      debounce(result => {
```
Haroenv, line 62: could you use debounceAsync here to only have one helper?

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,102 @@
...
+    if (!_query) {
+      setIsLoading(false);
+      setHits([]);
+      return;
+    }
+    setIsLoading(true);
+    runConcurrentSafePromise(
+      debouncedSearch(_query, queryLanguages, {
+        ...extraParameters,
+        nbHits,
```
Haroenv, line 78: in theory these parameters could be outdated. I don't think it's a huge issue, but you probably need to deal with it like with _query

packages/react-instantsearch-dom/src/index.js
```diff
@@ -66,6 +66,10 @@ export { default as Stats } from './widgets/Stats';
 export { default as ToggleRefinement } from './widgets/ToggleRefinement';
 export { default as VoiceSearch } from './widgets/VoiceSearch';
 export { default as QueryRuleCustomData } from './widgets/QueryRuleCustomData';
+export { default as EXPERIMENTAL_Answers } from './widgets/Answers';
+
+// hooks
+export { default as EXPERIMENTAL_useAnswers } from './hooks/useAnswers';
```
Haroenv, line 72: I think hooks can't start with anything else than `use`, or is all-caps prefix allowed?

packages/react-instantsearch-dom/src/widgets/__tests__/Answers.js
```diff
@@ -0,0 +1,111 @@
...
+        <Answers
+          searchClient={searchClient}
+          queryLanguages={['en']}
+          answersComponent={({ isLoading }) =>
+            isLoading ? <p>is loading</p> : <p>not loading</p>
+          }
+        />
+      </InstantSearch>
+    );
+    getByText('not loading');
```
Haroenv, line 58: you are only relying on this throwing an error if it's not found, not on an assertion. I'm not sure if we have that pattern anywhere else. It initially confused me, but seems fine actually

packages/react-instantsearch-dom/src/widgets/__tests__/Answers.js
```diff
@@ -0,0 +1,111 @@
...
+      </InstantSearch>
+    );
+    expect(() => {
+      getByText('hits received');
+    }).toThrow();
+
+    fireEvent.change(getByPlaceholderText('Search here…'), {
+      target: { value: 'sarah' },
+    });
+    await waitFor(() => getByText('hits received'));
```
Haroenv, line 109: what about using `runAllMicrotasks` like here

https://github.com/algolia/react-instantsearch/blob/0a711a7533449330cd27b23a17b85b094b4be645/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js#L3

https://github.com/algolia/react-instantsearch/blob/0a711a7533449330cd27b23a17b85b094b4be645/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.derived.js#L99

(that way you don't need MutationObserver)

package.json
```diff
@@ -72,7 +73,7 @@
     "@wdio/selenium-standalone-service": "5.16.10",
     "@wdio/spec-reporter": "5.16.11",
     "@wdio/static-server-service": "5.16.10",
-    "algoliasearch": "4.0.0-beta.15",
+    "algoliasearch": "4.8.2",
```
Haroenv, line 76: you probably need to update bundlesize

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: Can you explain how they could be outdated? It's a function freshly made at every render. And `_query` is used immediately inside the function with no delay.

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2996/commits/df47dc3db50b1456a8da3fa54f2f5f82ea1a6d22

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2996/commits/29fb4a0485fef3bd39f0467a119144f5ece67f23

packages/react-instantsearch-dom/src/widgets/__tests__/Answers.js
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2996/commits/5dffc128369a87f42197e69ebbb35c99b8d10c70

shortcuts (approved):

packages/react-instantsearch-dom/src/widgets/__tests__/Answers.js
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+import { InstantSearch, SearchBox } from 'react-instantsearch-dom';
+import { render, fireEvent } from '@testing-library/react';
+import Answers from '../Answers';
+
+const runAllMicroTasks = (time = 0) =>
+  new Promise(resolve => setTimeout(resolve, time));
```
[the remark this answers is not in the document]
Haroenv: it's only used in tests, so maybe a root `/testutils`?

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,99 @@
+import React, { useState, useEffect, useMemo, useContext } from 'react';
```
francoischalifour, line 1: Any reasons for not having this new file in TypeScript?

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,99 @@
...
+  if (!hasReactHooks()) {
+    throw new Error(
+      `\`Answers\` component and \`useAnswers\` hook require all React packages to be 16.8.0 or higher.`
+    );
+  }
+  const context = useContext(instantSearchContext);
+  const [query, setQuery] = useState();
+  const [index, setIndex] = useState();
+  const [isLoading, setIsLoading] = useState();
+  const [hits, setHits] = useState();
```
francoischalifour, line 30: We should default this to `[]` to not introduce another type and conditional in user code:

image

```suggestion
  const [hits, setHits] = useState([]);
```

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,99 @@
...
+  searchDebounceTime = 100,
+  ...extraParameters
+}) {
+  if (!hasReactHooks()) {
+    throw new Error(
+      `\`Answers\` component and \`useAnswers\` hook require all React packages to be 16.8.0 or higher.`
+    );
+  }
+  const context = useContext(instantSearchContext);
+  const [query, setQuery] = useState();
```
francoischalifour, line 27: Why don't we set the initial value to `context.store.getState().widgets.query` here?

Right now we deal with two data types: `undefined | string`.

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,99 @@
...
+  const [index, setIndex] = useState();
```
francoischalifour, line 28: Why don't we set the initial value to `context.mainTargetedIndex` here?

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,99 @@
...
+  const [isLoading, setIsLoading] = useState();
```
francoischalifour, line 29: We should give it an initial value too.

francoischalifour, line 29: How about a status state instead of a loading state?

packages/react-instantsearch-dom/src/widgets/Answers.js
```diff
@@ -0,0 +1,49 @@
...
+    </div>
+  );
+}
+
+export default function Answers({
+  searchClient,
```
francoischalifour, line 28: Is the Answers search client always the same as InstantSearch's index? If not, would it make sense to get it from the parent `Index`?

eunjae-lee: It makes sense, and I will address it in the next PR, but how would you access it?

packages/react-instantsearch-dom/src/widgets/__tests__/Answers.js
```diff
@@ -0,0 +1,112 @@
...
+    );
+    expect(() => {
+      getByText('hits received');
+    }).toThrow();
+
+    fireEvent.change(getByPlaceholderText('Search here…'), {
+      target: { value: 'sarah' },
+    });
+    await runAllMicroTasks(10);
+    getByText('hits received');
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/react-instantsearch/pull/2996/commits/494a836003a47157824d652d01286f4ed2a76d9f

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: It makes sense https://github.com/algolia/react-instantsearch/pull/2996/commits/0228c51c2d65e477fbce6471e04a85597f5bfda6

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: It makes sense https://github.com/algolia/react-instantsearch/pull/2996/commits/0228c51c2d65e477fbce6471e04a85597f5bfda6

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: It makes sense https://github.com/algolia/react-instantsearch/pull/2996/commits/0228c51c2d65e477fbce6471e04a85597f5bfda6

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: It makes sense https://github.com/algolia/react-instantsearch/pull/2996/commits/0228c51c2d65e477fbce6471e04a85597f5bfda6

eunjae-lee: normally status state would be appropriate, but here we don't need any other state, so isLoading should be sufficient. If we need to know about other state, or need to introduce something like `isLoaded`, then we can convert it to status state.

francoischalifour: In the context of Answers I think it makes perfect sense to introduce multiple status that reflects the current state of the search. Answers is tricky in terms of UX because it's slower than regular Algolia indices. IMO it would make sense to also have a "stalled" state to allow better rendering of the component.

Example:

- Wait for input (`idle`): display nothing
- Start typing (`loading`): display nothing
- Network is slow (`stalled`): display a scaffold

With the current solution it's not trivial to do.

Possible TypeScript definition:

```ts
type AnswersStatus = 'idle' | 'loading' | 'stalled';
```

We also use a similar concept for the Voice widget I believe. If this is too much of a change compared to how React InstantSearch functions, that's fair! Since the widget is experimental we can still change this in a minor.

packages/react-instantsearch-dom/src/widgets/Answers.js
francoischalifour: Probably by storing the search client in the context?

packages/react-instantsearch-dom/src/hooks/useAnswers.js
Haroenv: if the function is made fresh, why do we need the _query variable if there's already `query`?

packages/react-instantsearch-dom/src/hooks/useAnswers.js
Haroenv: shouldn't that take IndexContext in account too?

francoischalifour: Yes, indeed!

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: you're right!

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: IIUC, it's related to https://github.com/algolia/react-instantsearch/pull/2996#discussion_r565490874

In the next PR, I will expose `indexContext` as I exposed `instantSearchContext` and I will take the indexName from either of contexts. Does it make sense to you?

francoischalifour: Yes, as you prefer!

packages/react-instantsearch-dom/src/hooks/useAnswers.js
eunjae-lee: It's a good idea to split it into `loading` and `stalled` so that they can display a scaffold only on `stalled`. Since we're updating the vanilla PR and the React one at the same time, it will slow down the process at the moment. I'd like to get these two released first, so that customers can start using, and then I'll take a look at how to introduce these statuses.

packages/react-instantsearch-dom/src/hooks/useAnswers.js
```diff
@@ -0,0 +1,94 @@
...
+import { debounce } from '../lib/debounce';
+
+function hasReactHooks() {
+  // >= 16.8.0
+  const [major, minor] = React.version.split('.').map(Number);
+  return major >= 17 || (major === 16 && minor >= 8);
+}
+
+export default function useAnswers({
+  searchClient,
```
Haroenv, line 13: this is accessible from the InstantSearch context I think (otherwise should get exposed that way)

eunjae-lee: I just checked and the context doesn't have it. I'll continue that part in a subsequent PR, if you think it's okay.

packages/react-instantsearch-dom/src/widgets/Answers.js
Haroenv: Ok, let's do it in a separate PR

yarn.lock
```diff
@@ -46,6 +92,15 @@
...
+"@algolia/client-recommendation@4.8.2":
+  version "4.8.2"
+  resolved "https://registry.npmjs.org/@algolia/client-recommendation/-/client-recommendation-4.8.2.tgz#f8483adca6ce829414d8e8c5b58420b22d160d49"
+  integrity sha512-evngF6Odrw93gXkXrOYPXxTWwDQ2K01sadB3Xpa1hQb+vjiBwcA/54w6nKyE4aiII1loT5q+Uj+G1f8HwBuksw==
+  dependencies:
+    "@algolia/client-common" "4.8.2"
+    "@algolia/requester-common" "4.8.2"
+    "@algolia/transporter" "4.8.2"
+
 "@algolia/client-search@4.0.0-beta.15":
```
Haroenv, line 104: something still uses this beta, maybe an example?

Haroenv: can be done in next pr

Haroenv (approved):
there's some loose ends to fix, but this can work for a first version for me

PR #3005, opened 2021-01-25
fix(types): fix type errors regarding SpeechRecognition

eunjae-lee:
**Summary**

Since the module `typescript` has been upgraded to `3.8.3` from #2996, it started to throw new errors regarding SpeechRecognition.
However the correct types are not released yet (https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/924), so this PR copies and pastes the related type and adds some `@ts-ignore` comments to pass the `yarn type-check`.

### DIFF
--- a/packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
+++ b/packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
@@ -1,5 +1,24 @@
 // copied from https://github.com/algolia/instantsearch.js/blob/688e36a67bb4c63d008d8abc02257a7b7c04e513/src/lib/voiceSearchHelper/index.ts
 
+// begin - wrong SpeechRecognition-related types
+// This is not released in typescript yet, so we're copy&pasting the type definition from
+// https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/924
+export type SpeechRecognitionErrorCode =
+  | 'aborted'
+  | 'audio-capture'
+  | 'bad-grammar'
+  | 'language-not-supported'
+  | 'network'
+  | 'no-speech'
+  | 'not-allowed'
+  | 'service-not-allowed';
+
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
```diff
@@ -1,5 +1,24 @@
 // copied from https://github.com/algolia/instantsearch.js/blob/688e36a67bb4c63d008d8abc02257a7b7c04e513/src/lib/voiceSearchHelper/index.ts
 
+// begin - wrong SpeechRecognition-related types
```
Haroenv, line 3:
```suggestion
// #region wrong SpeechRecognition-related types
```

packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
```diff
@@ -1,5 +1,24 @@
...
+  | 'network'
+  | 'no-speech'
+  | 'not-allowed'
+  | 'service-not-allowed';
+
+interface SpeechRecognitionErrorEvent extends Event {
+  readonly error: SpeechRecognitionErrorCode;
+  readonly message: string;
+}
+// end - wrong SpeechRecognition-related types
```
Haroenv, line 20:
```suggestion
// #endregion wrong SpeechRecognition-related types
```

packages/react-instantsearch-dom/src/lib/voiceSearchHelper/index.ts
Haroenv: this should allow it to collapse, but not 100% sure

eunjae-lee: Yup, it works!

Haroenv (approved):

PR #3009, opened 2021-02-16
feat(smartSort): add widget

shortcuts:
Implements `smartSort` widget

related: https://github.com/algolia/instantsearch.js/pull/4648

PR #3010, opened 2021-02-16
feat(stats): apply nbSortedHits

shortcuts:
**Summary**

- Updates `stats` widget to reflect `nbSortedHits` and `isSmartSorted`
- Add related tests
- Move files to typescript

related: https://github.com/algolia/react-instantsearch/pull/3009


packages/react-instantsearch-dom/src/widgets/Stats.ts
```diff
@@ -2,12 +2,12 @@ import { connectStats } from 'react-instantsearch-core';
...
 
 /**
- * The Stats component displays the total number of matching hits and the time it took to get them (time spent in the Algolia server).
+ * The Stats component displays the total number of matching hits and the time it took to get them (time spent in the Algolia server) if `relevancyStrictness` is `undefined`. Displays the total number of matching sorted hits out of the matching hits and the time it took to get them (time spent in the Algolia server) otherwise.
  * @name Stats
  * @kind widget
  * @themeKey ais-Stats - the root div of the widget
  * @themeKey ais-Stats-text - the text of the widget - the count of items for each item
- * @translationkey stats - The string displayed by the stats widget. You get function(n, ms) and you need to return a string. n is a number of hits retrieved, ms is a processed time.
+ * @translationkey stats - The string displayed by the stats widget. You get function({ n, nSorted, isSmartSorted, ms }) and you need to return a string. `n` is a number of hits retrieved, `nSorted` is a number of sorted hits retrieved, `isSmartSorted` is a boolean translating a `relevancyStrictness` set between 0 and 100, `ms` is a processed time.
```
shortcuts, line 10: I'm not sure if this change will be accepted but I feel like it's better to have an object to avoid retrieving unused params

Haroenv: unfortunately that's a breaking change. Can you make the signature `(n, ms, {nbHits, nbSortedHits, isSmartSorted})` to avoid the breakage?

shortcuts: Sure. As `n` is the `nbHits`, `{nbSortedHits, isSmartSorted}` should be enough

Haroenv: I put nbHits in the third argument so you can use only that one, but on reflection, a regular argument list might be most consistent: `(nbHits, ms, nbSortedHits, isSmartSorted)`, even though it's odd that ms is in the middle

shortcuts: Got it!

shortcuts: Done in https://github.com/algolia/react-instantsearch/pull/3010/commits/b1886239778a3be9f1c5ff2bed5c57030de0fe32

packages/react-instantsearch-dom/src/components/Stats.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export default translatable({
+  stats: (
+    n: number,
+    ms: number,
+    nSorted?: number,
+    isSmartSorted?: boolean
+  ): string =>
+    isSmartSorted && nSorted !== undefined && n !== nSorted
```
Haroenv, line 63:
```suggestion
    isSmartSorted && n !== nSorted
```

is this part of the condition needed?

packages/react-instantsearch-dom/src/components/Stats.tsx
```diff
@@ -0,0 +1,66 @@
...
+  nbHits: PropTypes.number.isRequired,
+  nbSortedHits: PropTypes.number,
+  processingTimeMS: PropTypes.number.isRequired,
+  translate: PropTypes.func.isRequired,
+};
+
+export default translatable({
+  stats: (
+    n: number,
+    ms: number,
```
Haroenv, line 59: I only just realised ms is a number and not the ms module, it makes sense like this!

packages/react-instantsearch-dom/src/components/Stats.tsx
```diff
@@ -0,0 +1,66 @@
...
+  );
+};
+
+Stats.propTypes = {
+  className: PropTypes.string,
+  isSmartSorted: PropTypes.bool.isRequired,
+  nbHits: PropTypes.number.isRequired,
+  nbSortedHits: PropTypes.number,
+  processingTimeMS: PropTypes.number.isRequired,
+  translate: PropTypes.func.isRequired,
```
Haroenv, line 53: is it really a function and not an object with functions? it was like this before already though

packages/react-instantsearch-dom/src/components/Stats.tsx
shortcuts: As `nSorted` and `isSmartSorted` are optional, TypeScript will yet at line 64 if this condition is not here

packages/react-instantsearch-dom/src/components/__tests__/Stats.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const tree = renderer.create(
+      <Stats
+        isSmartSorted={true}
+        nbHits={42}
+        nbSortedHits={42}
+        processingTimeMS={0}
+      />
+    );
+
+    expect(tree.toJSON()).toMatchSnapshot();
```
Haroenv, line 50: since these are new tests, what do you think about using inline snapshots? they seem small enough

packages/react-instantsearch-dom/src/components/Stats.tsx
Haroenv: annoying, really isSmartSorted already guarantees that they're not undefined, so I'd use `!` on like 64

packages/react-instantsearch-dom/src/components/Stats.tsx
shortcuts: From my POV, `translate` is a function, but `translatable` is an object with functions

packages/react-instantsearch-dom/src/components/Stats.tsx
shortcuts: Make sense! https://github.com/algolia/react-instantsearch/pull/3010/commits/c2e526bfababc9b42557c0f212a81e1a750b73a7

packages/react-instantsearch-dom/src/components/__tests__/Stats.tsx
shortcuts: Sure, done https://github.com/algolia/react-instantsearch/pull/3010/commits/bf9509333d9a655e14411a46953c8ea8b1505b4b. Do you think it would make sense to do it for https://github.com/algolia/react-instantsearch/pull/3009 ?

eunjae-lee: Yes, if not too big.

Haroenv (approved):

Issue #2742, opened 2019-08-06
Refresh does not request pages in infinite hits again


Haroenv:
**Describe the bug 🐛**

When updating the `refresh` token, infinite hits will not react to the refetch.

**To Reproduce 🔍**

1. go to the example
2. trigger next page
3. refresh
4. see no changes in the infinite hits, even if the index would have changed

The example's index can be changed to an index you control where you make changes to make the effect clearer.

> https://codesandbox.io/s/react-instantsearch-app-dq2ro

**Expected behavior 💭**

Infinite hits reacts to refresh, either by:

1. clearing the cache completely and restarting on the current page
3. clearing the cache and the state and restarting on page 0

Likely we should go for option 2, unless there's a showPrevious.

**Additional context**

When refresh happens, we need to make sure the infinite hits internal cache is also invalidated.

On React Native, where the list component itself is stateful, we can not rely on the "key" hack, because it rerenders with an empty state when we simply clear the cache. What could be an option is:

1. clear cache
2. redo current search
3. save in cache
4. rerender

The problem is that you can't do that as a user reasonably, since you don't have access to the helper state.

A possible solution is:

In the function `refresh` in InstantSearchManager, emit an event to all widgets. Then in InfiniteHits, listen to that event, and clear the internal cache as we expect.

Another potential solution is to return a promise from the `search` which happens in refresh. This should allow people to rerender the InfiniteHits component manually.

Relevant pieces of code: 

https://github.com/algolia/react-instantsearch/blob/ec9e0fbd6106d1c3e47f1dbfa6eaac3e20af6bd5/packages/react-instantsearch-core/src/core/createInstantSearchManager.js#L69-L72

Relevant issues:

https://github.com/algolia/react-instantsearch/issues/2464

alexpchin:
Hi all,

Just to give you a demonstration of what happens here if you refresh using a key. Unfortunately, you momentarily receive hits of length 0.

![](https://ephmedia.giphy.com/65ccc16d-299e-4432-975c-91c2e55c8d43.gif)

I have created an example to use ClearCache and a searchClient object - which does not work. https://snack.expo.io/@alexpchin/aglolia-refresh-problem
Please run on iOS and look out for the red flash.

The full code is here:  https://github.com/alexpchin/algolia-refresh-issue/

The docs say this method is (browserOnly)? 
https://www.algolia.com/doc/api-client/advanced/cache-browser-only/javascript/?language=javascript

[...]

PR #3011, opened 2021-02-18
fix(infiniteHits): do not cache the cached hits

eunjae-lee:
**Summary**

This PR fixes the bug where InfiniteHits stored the cached hits, so clearing cache didn't make any effect.

* related
  * https://github.com/algolia/react-instantsearch/issues/2742
  * https://github.com/algolia/instantsearch.js/pull/4534

**Result**

InfiniteHits reads cache object everytime and do not store it inside the connector itself.

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
+++ b/packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
@@ -80,14 +80,12 @@ export default createConnector({
 
     this._prevState = this._prevState || {};
 
-    const cache = props.cache || getInMemoryCache();
-    if (this._cachedHits === undefined) {
-      this._cachedHits = cache.read({ state: searchState }) || {};
-    }
+    this._cache = props.cache ? props.cache : this._cache || getInMemoryCache();
+    let cachedHits = this._cache.read({ state: searchState }) || {};
 
     if (!results) {
       return {
-        hits: extractHitsFromCachedHits(this._cachedHits),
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -80,14 +80,12 @@ export default createConnector({
 
     this._prevState = this._prevState || {};
 
-    const cache = props.cache || getInMemoryCache();
-    if (this._cachedHits === undefined) {
-      this._cachedHits = cache.read({ state: searchState }) || {};
-    }
+    this._cache = props.cache ? props.cache : this._cache || getInMemoryCache();
```
Haroenv, line 83: this is a very important fix!

Haroenv (approved):
I'm a fan, thanks!

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
eunjae-lee: Yes, it especially makes sure to take `cache` prop into account even when user passes a different object!

Haroenv: and makes sure we don't recreate the memory cache on every render

alexpchin:
@Haroenv Do you know when this will be released?

eunjae-lee:
@alexpchin It's just released in 6.10.0

PR #3012, opened 2021-02-23
fix(examples): use react-instantsearch-native in native examples

shortcuts:
This PR solve the `text:examples` fails


examples/react-native-query-suggestions/App.js
```diff
@@ -12,12 +12,14 @@ import {
...
+  InstantSearch,
+  Configure,
+  Index,
   connectSearchBox,
   connectInfiniteHits,
   connectHits,
   connectRefinementList,
-} from 'react-instantsearch/connectors';
-import { InstantSearch, Configure, Index } from 'react-instantsearch/native';
+} from 'react-instantsearch-native';
```
eunjae-lee, line 22: Do you know why it doesn't work as before?

shortcuts: Not really, neither do I know why we were using `react-instantsearch/...` instead of `react-instantsearch-native` or `react-instantsearch-dom`.  Was it just outdated?

eunjae-lee: we have https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/native.js
so it should've worked without this change.

shortcuts: True but the other `react-native` test use `react-instantsearch-native` so in either way I think it should be better to make them use the same package. I'll try to see why the other way doesn't work

shortcuts: I just realized that this is the only part where this kind of import is used. Has it ever worked outside of this example?

francoischalifour: The `react-instantsearch-native` import is the most up-to-date way to import the package. It allows better tree shaking IIRC.

francoischalifour (approved):

eunjae-lee (approved):

PR #3014, opened 2021-02-25
fix(SmartSort): make `textComponent` and `buttonTextComponent` optional

shortcuts:
**Summary**

To match the `InstantSearch.js` implementation (https://github.com/algolia/instantsearch.js/pull/4648), we made the `textComponent` and `buttonTextComponent` props optional and provided a default `ButtonTextComponent`

### DIFF
--- a/packages/react-instantsearch-dom/src/components/SmartSort.tsx
+++ b/packages/react-instantsearch-dom/src/components/SmartSort.tsx
@@ -21,7 +21,7 @@ export type SmartSortProps = {
 const DefaultButtonTextComponent = ({
   isSmartSorted,
 }: SmartSortComponentProps) => (
-  <div>{isSmartSorted ? 'See all results' : 'See relevant results'}</div>
+  <span>{isSmartSorted ? 'See all results' : 'See relevant results'}</span>
 );
 
 const SmartSort: React.FC<SmartSortProps> = ({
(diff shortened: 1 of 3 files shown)


eunjae-lee:
Can you make it compliant with the spec? I understand `FunctionComponent` cannot return a string but would like to find a solution. If not, we should update the spec and IS.js so that all of the flavors return the same DOM.

packages/react-instantsearch-dom/src/components/SmartSort.tsx
```diff
@@ -13,23 +13,29 @@ export type SmartSortProps = {
...
-  textComponent: React.FunctionComponent<SmartSortComponentProps>;
+  buttonTextComponent?: React.FunctionComponent<SmartSortComponentProps>;
+  textComponent?: React.FunctionComponent<SmartSortComponentProps>;
   refine(relevancyStrictness: number | undefined): void;
 };
 
+const DefaultButtonTextComponent = ({
+  isSmartSorted,
+}: SmartSortComponentProps) => (
+  <div>{isSmartSorted ? 'See all results' : 'See relevant results'}</div>
```
eunjae-lee, line 24:
```suggestion
  <span>{isSmartSorted ? 'See all results' : 'See relevant results'}</span>
```

I think we can still change it to be a `span`.

shortcuts: Sure!

eunjae-lee (approved):

Haroenv (approved):

PR #3021, opened 2021-03-01
fix(infiniteHits): fix stale hits issue

### LINKED ISSUE #3018: Search old results by one state
Hello!
I have an issue, which happens just on stage and production servers. On localhost everything works fine.
I’m using react-instantsearch-dom inside Next.js project.
The problem is that results, visible for a user, are old, they were fetched by previous search state, not by current one. In other words:

1) you see initial search page with all results
2) you type search query for a 1st time - nothing changed (but in this step results should be different)
3) you change search param/query again and see results from step 2 (in this step you already have 3rd SearchState)
4) you change something - create 4th SearchState, but see results from step 3

Like one step behind, hope I clearly explained it.
To my mind the problem is in connectInfiniteHits, because in console I see that Algolia made a correct request (by new SearchState, also url is updating correctly), in response there is a new portion of data, but inside connectInfiniteHits component I receive an old portion of data (by previous search state, not by current). It’s soooo bad for user experience and I don’t understand what is wrong.

Will be appreciate for any suggestion and help

eunjae-lee:
**Summary**

This PR fixes the stale hits issue with InfiniteHits.

fixes #3018

**Details**

* `searchState` is initially an empty object until it receives the first search API response.
* When `searchState` changes, it triggers a search API, but inside the connector there's a time when `searchState` is new, but `searchResults` is still stale because the response hasn't arrived yet.
* `results._state` (state from the helper) is always synced with the hits from `results` object. So it makes sense to use `results._state` as a key for cache.

---

- [x] ~It should store cache based on `searchState`, not the one from `searchResults` because we want to display hits from cache before calling API request if the cache exists~
- [x] ~`searchState` changes right after user interaction but the hits from `searchResults` can be still the ones from the previous state. We need to distinguish this part~
- [x] https://codesandbox.io/s/spring-frost-1o91f → fixes the stale hits?
- [x] https://codesandbox.io/s/reverent-wiles-fqmr3?file=/src/App.js → works with url routing?

**Result**


packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -318,14 +318,14 @@ describe('connectInfiniteHits', () => {
...
-        'theQueryID_2',
-        'theQueryID_2',
+        'theQueryID_2_',
+        'theQueryID_2_',
+        'theQueryID_2_',
+        'theQueryID_2_',
+        'theQueryID_2_',
+        'theQueryID_2_',
+        'theQueryID_2_',
+        'theQueryID_2_',
```
Haroenv, line 328: what's this change for?

eunjae-lee: https://github.com/algolia/react-instantsearch/blob/ad9a17e5beff2794e0d4c4f1d75595ee7f421af4/packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js#L265-L274

retrying with a different results updates the hits being rendered (override the cache)

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -498,17 +498,6 @@ describe('connectInfiniteHits', () => {
       expect(instance.refine).toHaveBeenLastCalledWith(event, 1);
     });
 
-    it('adds 1 to page when calling refine', () => {
```
eunjae-lee, line 501: `refine` is not a public method, so we don't need to test it. And we don't need this behavior either.

packages/react-instantsearch-core/src/connectors/__tests__/connectInfiniteHits.js
```diff
@@ -520,14 +509,6 @@ describe('connectInfiniteHits', () => {
       expect(state1).toEqual({ page: 6 });
     });
 
-    it('automatically converts String state to Number', () => {
```
eunjae-lee, line 523: no need to test `refine` and this behavior doesn't need to be tested because refineNext and refinePrevious already pass a number to it.

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -78,14 +78,17 @@ export default createConnector({
...
+    if (results === null) {
+      cachedHits = {};
+    } else {
+      cachedHits = this._cache.read({ state: results._state }) || {};
+    }
 
     if (!results) {
       return {
-        hits: extractHitsFromCachedHits(cachedHits),
+        hits: [],
```
Haroenv, line 91: this isn't perfect, but I don't see how we can get something that's sync with the results when there are results, but also exists when there's no results yet (searchState is "eager", which means we shouldn't use it to cache yet, and it has a different shape to _state)

Haroenv (approved):

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -94,29 +89,20 @@ export default createConnector({
...
 
-    if (!isEqual(currentState, this._prevState)) {
-      cachedHits = this._cache.read({ state: searchState }) || {};
-    }
-    if (cachedHits[page] === undefined) {
-      cachedHits[page] = hitsWithPositionsAndQueryID;
-      this._cache.write({ state: searchState, hits: cachedHits });
-    }
+    cachedHits[page] = hitsWithPositionsAndQueryID;
+    this._cache.write({ state: results._state, hits: cachedHits });
```
Haroenv, line 104:
```suggestion
    this._cache.write({ state, hits: cachedHits });
```

packages/react-instantsearch-core/src/connectors/connectInfiniteHits.js
```diff
@@ -94,29 +89,20 @@ export default createConnector({
...
-      page,
-      hits,
-      hitsPerPage,
-      nbPages,
-      _state: { page: p, ...currentState } = {},
-    } = results;
+    const { page, hits, hitsPerPage, nbPages, _state } = results;
+
+    this._cache = props.cache ? props.cache : this._cache || getInMemoryCache();
+    const cachedHits = this._cache.read({ state: _state }) || {};
```
Haroenv, line 95:
```suggestion
    const { page, hits, hitsPerPage, nbPages, _state: state } = results;

    this._cache = props.cache ? props.cache : this._cache || getInMemoryCache();
    const cachedHits = this._cache.read({ state }) || {};
```

yannickcr (approved):

PR #3037, opened 2021-03-23
fix(SearchBox): RTL support

ShaMan123:
`dir="auto"` for rtl auto support
## Algolia is AMAZING!!
Thanks

## Summary
There is no access to the underlying `input` element, not able to pass props.
`dir="auto"` is safe for all use cases

## Result
a
b


ShaMan123:
workaround
```js
 document.getElementsByClassName('ais-SearchBox-input')[0].setAttribute('dir', 'auto');
```

Haroenv:
this is an interesting PR, we need to do some more checks to see if this is sufficient in every case, but I think in the mean time a different workaround is: `.ais-SearchBox-input { direction: auto }` via css. Do you confirm that also works?

ShaMan123:
unfortunately not.
css doesn't support `auto` value

Haroenv:
Sorry for the delay here, we were testing this in other flavours. I'm in favour of adding this attribute by default. Do you think you can update the relevant tests @ShaMan123? Thanks!

ShaMan123:
Maybe. Could you point me to the relevant tests?

Haroenv:
The snapshot tests are here https://app.circleci.com/pipelines/github/algolia/react-instantsearch/1077/workflows/2876d94c-986f-485d-b16c-ea4073c9c131/jobs/44752 @ShaMan123 :)

ShaMan123:
@Haroenv I think it's done.

Haroenv (approved):
looks great! just checking for a second approval in the team

shortcuts (approved):
Looks good!

Haroenv:
We are thinking slightly about whether just this input change is enough @ShaMan123, would you in hebrew expect the icons to be positions reversely too or not?

shortcuts:
> Would you in hebrew expect the icons to be positions reversely too or not?

@Haroenv  I didn't found much [search resources](https://www.maariv.co.il/#/search;query=test) but it looks like only the text position itself changes

ShaMan123:
I would expect the text to be positioned to the right and the icons to left.

בתאריך יום ה׳, 6 במאי 2021, 14:33, מאת Clément Vannicatte ‏<
***@***.***>:

sarahdayan:
@ShaMan123 We're in the process of moving `algolia/react-instantsearch` to our InstantSearch monorepo. This PR won't be transferrable, so I'll close it here, but please feel free to reopen it once the migration is done.

Issue #3044, opened 2021-04-20
Feature Request: PoweredBy dark background logo


luciobenini:
Hello!
As told in the title, please add the negative versione of the logo.

Haroenv:
thanks for opening this issue, it seems like React InstantSearch is actually still using the old logo here, so that needs to be updated!

Haroenv:
In the mean time, the correct dom for dark mode is:

```html
[...]

PR #3045, opened 2021-04-21
feat(poweredBy): add `theme` property

shortcuts:
**Summary**

Similar implementation as in [InstantSearch.js](https://github.com/algolia/instantsearch.js/blob/master/src/components/PoweredBy/PoweredBy.tsx)

- Adds a new `theme` prop to the `poweredBy` component to handle dark mode. 
- Updates the Algolia logo

**Result**

Changes the logo/text color when `theme="dark"` is passed to the component

closes https://github.com/algolia/react-instantsearch/issues/3044


packages/react-instantsearch-dom/src/components/PoweredBy.js
```diff
@@ -7,70 +7,73 @@ import { createClassNames } from '../core/utils';
...
 /* eslint-disable max-len */
-const AlgoliaLogo = () => (
+const AlgoliaLogo = ({ theme }) => (
   <svg
+    width="168"
     xmlns="http://www.w3.org/2000/svg"
-    baseProfile="basic"
-    viewBox="0 0 1366 362"
-    width="100"
-    height="27"
```
Haroenv, line 16: the width and height should probably stay the same as before, to avoid breaking change?

shortcuts: I think it is the same size as before but the SVG isn't, I'll try to see if I can set the same width/height

Haroenv: I think the height matches, but the new width is a tiny bit more. Not sure whether it's worth worrying about

francoischalifour (commented):
Do we really need to change the API to support dark themes? Can't this be solved with CSS?

The API will be different from other InstantSearch flavors now.

shortcuts:
> Do we really need to change the API to support dark themes? Can't this be solved with CSS?
> 
> The API will be different from other InstantSearch flavors now.

This implementation is the same as in the [InstantSearch.js PoweredBy component](https://github.com/algolia/instantsearch.js/blob/master/src/components/PoweredBy/PoweredBy.tsx)

shortcuts:
> It doesn't seem to work as expected on dark mode:
> See [story](https://deploy-preview-3045--react-instantsearch.netlify.app/storybook/?path=/story/poweredby--with-dark-theme).

@francoischalifour @Haroenv It seems like the issue is related to storybook not picking the changes locally.

Importing the component with `import { PoweredBy } from '../packages/react-instantsearch-dom'` in `PoweredBy.stories.js` actually show the changes.


https://user-images.githubusercontent.com/20689156/120993692-16f99d00-c784-11eb-9a84-135371e4928e.mov

sarahdayan:
@shortcuts We're in the process of moving `algolia/react-instantsearch` to our InstantSearch monorepo. This PR won't be transferrable, so I'll close it here, but please feel free to reopen it once the migration is done.

PR #3047, opened 2021-04-28
feat(connectNumericMenu): add support for floating point values

VincentRoest:
**Summary**

I was trying to figure out why our numeric menu with statically defined percentages could not filter on <= 0.5%, which it couldn't.

**Result**

Valid searches not showing up after filtering -> nowhere in the docs is it specified that numeric menu is INTEGER only. 

Simple fix: parseInt -> parseFloat. I'm not missing the radix argument, I doubt others will have different numeric bases than 10.


Haroenv:
I think this makes sense, could you add some tests related to this change? Thanks!

VincentRoest:
Hey! Would you require any more tests? I added the props correctly derived + filters created.

Haroenv:
I'll release this Monday!

PR #3050, opened 2021-05-05
fix(maps): leave zoom in place if the bounds did not change

anikolskiy:
Fix map-zoom after browser-zoom

**Summary**

We encountered an issue with a map and connected list of hits. When browser zoom is set to 125% (or any other value really) and we tried to zoom in the map (with a "+" button), map did zoom in, but immediately zoomed out again. This issue happened on almost (but not every) press of the "+" button in map controls. And it happens in Chrome 90.0.4430.93 on Mac OS X, but was also reported by colleagues on Windows with the same Chrome version.

**Result**

After debugging I figured out that fitBounds() is being called multiple times after "+" button is pressed. Some of these calls were setting bounds to the same value that was already set. By adding a condition around fitBounds() to not execute it unless bounds are different from the current bounds, problem with zooming-out immediately after zooming-in is solved.

Please note that there are no additional tests for this fix.

**Guess of what the root cause is**

My guess is that the issue is related to "refine" functionality and perhaps timing.

I can imagine the scenario where first fitBounds() call with original wider bounds (before zooming in) generates a call to "refine" which in turn sends a request (1) to fetch data. Immediately after that another fitBounds() call with narrower area is executed, which in turn executes "refine" functionality again and that sends another request (2) to fetch data.

Now, because request(2) will probably have less data, it could finish earlier than request(1). So then bounds could be reset to the area before zooming-in.

This is a wild guess though and I'll let poeple who are more familiar with react-instantsearch code than me to perhaps investigate further.

### DIFF
--- a/packages/react-instantsearch-dom-maps/src/GoogleMaps.js
+++ b/packages/react-instantsearch-dom-maps/src/GoogleMaps.js
@@ -68,13 +68,23 @@ class GoogleMaps extends Component {
 
     if (boundingBox) {
       this.lockUserInteration(() => {
-        this.instance.fitBounds(
-          new google.maps.LatLngBounds(
-            boundingBox.southWest,
-            boundingBox.northEast
-          ),
-          boundingBoxPadding
+        const oldBounds = this.instance.getBounds();
+        const newBounds = new google.maps.LatLngBounds(
+          boundingBox.southWest,
+          boundingBox.northEast
         );
+
+        if (
+          !oldBounds ||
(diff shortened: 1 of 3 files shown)


Haroenv (approved):
makes sense to me now, @anikolskiy, do you mind testing my version again, in case anything changed?

anikolskiy:
@Haroenv sure, I'll test it today or tomorrow.

anikolskiy:
Just tested it and it works perfectly.

PR #3056, opened 2021-05-20
feat(DynamicWidgets): add implementation

Haroenv:
**Summary**

Adds a new widget _ExperimentalDynamicWidgets_ that can be used to conditionally render other widgets. This condition is based on the search results, by default results.facetOrdering.facet.values, but can be overridden using transformItems

**Result**

```jsx
<ExperimentalDynamicWidgets
  transformItems={(_attributes, { results }) => {
    // if you want to use userData instead of facet ordering
    return results.userData[0].ordering;
  }}
>
  <HierarchicalMenu
    attributes={[
      'hierarchicalCategories.lvl2',
      'hierarchicalCategories.lvl3',
    ]}
  />
  <Panel>
    <RefinementList attribute="brand" />
  </Panel>
  <Menu attribute="categories" />
</ExperimentalDynamicWidgets >
```

Note that the widgets to be rendered have to have an _attribute_ prop, you can't "wrap" a widget like this:

```jsx
// does not work
const Wrapper1 = ({ attr })  => <RefinementList attribute={attr} />

// does work
const Wrapper2 = ({ attribute })  => <RefinementList attribute={attribute} />

<ExperimentalDynamicWidgets>
  <Wrapper1 attr='test'/> // does not work
  <Wrapper2 attribute='test'/> // does work
</ExperimentalDynamicWidgets>
```

See also:

- https://github.com/algolia/instantsearch.js/pull/4687
- https://github.com/algolia/vue-instantsearch/pull/922
- DX-1668

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectDynamicWidgets.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectDynamicWidgets.js
@@ -58,18 +58,6 @@ describe('connectDynamicWidgets', () => {
         expect(actual).toEqual(expectation);
       });
 
-      it('error without transformItems', () => {
-        const props = { contextValue };
-        const searchState = {};
-        const searchResults = createSingleIndexSearchResults();
-
-        expect(() => {
-          connector.getProvidedProps(props, searchState, searchResults);
-        }).toThrowErrorMatchingInlineSnapshot(
-          `"props.transformItems is not a function"`
-        );
-      });
-
       it('attributesToRender is the return value of transformItems', () => {
         const returnValue = ['test1', 'test2'];
         const props = { contextValue, transformItems: () => returnValue };
@@ -97,6 +85,27 @@ describe('connectDynamicWidgets', () => {
         );
 
         expect(actual.attributesToRender).toEqual([]);
+      });
+
+      it('reads from facetOrdering by default', () => {
+        const props = {
+          contextValue,
+          ...connector.defaultProps,
+        };
+        const searchState = {};
+        const searchResults = createSingleIndexSearchResults({
+          renderingContent: {
+            facetOrdering: { facet: { order: ['one', 'two'] } },
+          },
+        });
+
+        const actual = connector.getProvidedProps(
+          props,
+          searchState,
+          searchResults
+        );
+
+        expect(actual.attributesToRender).toEqual(['one', 'two']);
       });
 
       it('transformItems gets called with results', () => {
@@ -234,6 +243,28 @@ describe('connectDynamicWidgets', () => {
         );
 
         expect(actual.attributesToRender).toEqual([]);
+      });
+
+      it('reads from facetOrdering by default', () => {
+        const props = {
+          contextValue,
+          indexContextValue,
+          transformItems: items => items,
+        };
+        const searchState = createMultiIndexSearchState();
+        const searchResults = createMultiIndexSearchResults({
+          renderingContent: {
+            facetOrdering: { facet: { order: ['one', 'two'] } },
+          },
+        });
+
+        const actual = connector.getProvidedProps(
+          props,
+          searchState,
+          searchResults
+        );
+
+        expect(actual.attributesToRender).toEqual(['one', 'two']);
       });
 
       it('transformItems gets called with results', () => {
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectDynamicWidgets.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectDynamicWidgets.js
@@ -195,19 +195,6 @@ describe('connectDynamicWidgets', () => {
         expect(actual).toEqual(expectation);
       });
 
-      it('error without transformItems', () => {
-        const props = {
-          contextValue,
-          indexContextValue,
-        };
-        const searchState = createMultiIndexSearchState();
-        const searchResults = createMultiIndexSearchResults();
-
-        expect(() =>
-          connector.getProvidedProps(props, searchState, searchResults)
-        ).toMatchInlineSnapshot(`[Function]`);
-      });
-
       it('attributesToRender is the return value of transformItems', () => {
         const returnValue = ['one', 'two'];
         const props = {
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectDynamicWidgets.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectDynamicWidgets.js
@@ -1,0 +1,284 @@
+import { SearchResults, SearchParameters } from 'algoliasearch-helper';
+import connector from '../connectDynamicWidgets';
+
+jest.mock('../../core/createConnector', () => x => x);
+
+const EMPTY_RESPONSE = {
+  results: [
+    {
+      hits: [],
+      nbHits: 0,
+      page: 0,
+      nbPages: 0,
+      hitsPerPage: 20,
+      exhaustiveNbHits: true,
+      query: '',
+      queryAfterRemoval: '',
+      params:
+        'highlightPreTag=%3Cais-highlight-0000000000%3E&highlightPostTag=%3C%2Fais-highlight-0000000000%3E&query=&facets=%5B%5D&tagFilters=',
+      index: 'instant_search',
+      processingTimeMS: 2,
+    },
+  ],
+};
+
+describe('connectDynamicWidgets', () => {
+  const empty = {};
+
+  describe('single index', () => {
+    const contextValue = {
+      mainTargetedIndex: 'index',
+    };
+
+    const createSingleIndexSearchResults = (result = {}, state) => ({
+      results: new SearchResults(new SearchParameters(state), [
+        {
+          ...EMPTY_RESPONSE.results[0],
+          ...result,
+        },
+      ]),
+    });
+
+    describe('getProvidedProps', () => {
+      it('empty results', () => {
+        const props = { contextValue };
+        const searchState = {};
+        const searchResults = empty;
+
+        const actual = connector.getProvidedProps(
+          props,
+          searchState,
+          searchResults
+        );
+
+        const expectation = {
+          attributesToRender: [],
+        };
+
+        expect(actual).toEqual(expectation);
+      });
+
+      it('error without transformItems', () => {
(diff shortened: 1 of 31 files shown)


packages/react-instantsearch-core/src/widgets/__tests__/DynamicWidgets.test.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+const RefinementList = connectRefinementList(
+  ({ attribute }) => `RefinementList(${attribute})`
+);
+
+const HierarchicalMenu = connectHierarchicalMenu(
+  ({ attributes }) => `HierarchicalMenu(${attributes.join(',')})`
+);
+
+describe('DynamicWidgets', () => {
```
francoischalifour, line 39: Perhaps add a test with deeply nested components so that we can align and track that behavior?

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
```diff
@@ -0,0 +1,52 @@
...
+  if (typeof component !== 'object') {
+    return undefined;
+  }
+
+  if (component.props.attribute) {
+    return component.props.attribute;
+  }
+  if (Array.isArray(component.props.attributes)) {
+    return component.props.attributes[0];
+  }
```
francoischalifour, line 15: I'm thinking about cases where perhaps using the first attribute could be limiting.

If we process the component if _any_ attributes match the `attributes` prop, it might become more flexible for users—especially in the context of the "scoped hierarchical menu" à la Amazon where the root changes (or is the `root` prop enough in that use case?).

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (!attribute) {
+      throw new Error(
+        `attribute could not be found for ${getDisplayName(child)}`
+      );
+    }
+    widgets.set(attribute, child);
+  });
+
+  // on initial render this will be empty, but React InstantSearch keeps
+  // search state for unmounted components in place, so routing works.
```
eunjae-lee, line 42: can you point me to which part of the code enables this?

packages/react-instantsearch-core/src/connectors/__tests__/connectDynamicWidgets.js
```diff
@@ -0,0 +1,284 @@
...
+
+      it('error without transformItems', () => {
+        const props = { contextValue };
+        const searchState = {};
+        const searchResults = createSingleIndexSearchResults();
+
+        expect(() => {
+          connector.getProvidedProps(props, searchState, searchResults);
+        }).toThrowErrorMatchingInlineSnapshot(
+          `"props.transformItems is not a function"`
```
eunjae-lee, line 69: Until we get data from the engine, we'd better provide a concrete error message for missing `transformItems`.

packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (!results) {
+      return { attributesToRender: [] };
+    }
+
+    // retrieve the facet order out of the results:
+    // results.facetOrder.map(facet => facet.attribute)
+    const facetOrder = [];
+
+    return {
+      attributesToRender: props.transformItems(facetOrder, { results }),
```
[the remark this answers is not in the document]
francoischalifour: `transformItems` won't be required once the engine implements it. So if we make it optional right now (defaulted to the identity function), we won't have to change our code and APIs to support it, it'll be transparent for both us and users.

Users won't have to upgrade the library for their code and index configuration to work.

So I think we should make `transformItems` optional right now.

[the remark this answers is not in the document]
Haroenv: If it's optional now, we can't change the default, since that's a breaking change, however _not having a default_ changing to having a default later isn't a breaking change, right?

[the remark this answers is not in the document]
francoischalifour: The default will always be to return the facet ordering coming from the engine, right? (meaning the breaking change will happen in the index configuration, which is outside of InstantSearch's responsibility.)

I see that we hardcode `const facetOrder = [];`, isn't the engine response specs final yet? (Can't we retrieve e.g., `response.facetOrder || []`?)

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
Haroenv: I assumed for hierarchical menus that people would set the first/root attribute as the one one to render hierarchically. If the menu is only displayed from a certain level onwards, that probably would be done either with a refinement and ignoring the levels that aren't shown or using `rootPath` for this. Concatenating or an array of attributes seems wrong, since for example in current refinement we use rootPath too

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <>
```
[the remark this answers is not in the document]
Haroenv: If we add a div, it has to move from -core to -dom, not sure if that's a big deal, but just wanted to make sure you think that's worth it (not being able to use the component directly without writing the rendering in react native for example)

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
Haroenv: React InstantSearch keeps searchState always as it is, unless cleanUp of a component has been called (when it unmounts). That means that whatever state is available over refinementList for example, it will not be removed until that component unmounts. Initially the componeent doesn't _need_ to be visible

eunjae-lee: Thanks for clarification.

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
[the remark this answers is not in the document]
eunjae-lee: That's a good point. I guess we can leave it at -core. There's nothing specific style to apply to this widget. And if users want, they can wrap it with a div to stack them horizontally or to do whatever they want.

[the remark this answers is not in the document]
francoischalifour: That makes sense, although that's inconsistent with our other flavors.

How about re-exporting `DynamicWidgets` from the DOM package that wraps it in a `div` with the `className`?

Shipow:
couldn't import **DynamicWidgets**.
`'react-instantsearch-dom' does not contain an export named 'DynamicWidgets'.`
tried with
`"react-instantsearch-dom": "https://pkg.csb.dev/algolia/react-instantsearch/commit/b5916e8d/react-instantsearch-dom",`

Haroenv:
It's called ExperimentalDynamicWidgets @Shipow

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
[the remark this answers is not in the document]
Haroenv: added a wrapper in 3be50ad

eunjae-lee (approved):
Looks good to me!

packages/react-instantsearch-core/src/widgets/__tests__/DynamicWidgets.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+              class="ais-Panel-body"
+            >
+              RefinementList(test3)
+            </div>
+          </div>
+        </div>
+      `);
+    });
+
+    test("does not render items that aren't directly in children", () => {
```
francoischalifour, line 261: Can we add a limitation section in the PR description that mentions that it doesn't support "proxy components"?

```jsx
function ProxyComponent({ attribute }) {
  return <RefinementList attribute={attribute} />;
}

function App() {
  return (
    <DynamicWidgets>
      <ProxyComponent attribute="brand" />
    </DynamicWidgets>
  );
}
```

It'll be easier to track later.

francoischalifour (approved):

packages/react-instantsearch-dom/src/widgets/__tests__/DynamicWidgets.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+const createSearchClient = () => ({
+  search: jest.fn(() => Promise.resolve(EMPTY_RESPONSE)),
+  searchForFacetValues: jest.fn(() => Promise.resolve({})),
+});
+
+const RefinementList = connectRefinementList(
+  ({ attribute }) => `RefinementList(${attribute})`
+);
+
+describe('DynamicWidgets', () => {
```
francoischalifour, line 41: We can add a test for the `className` prop.

packages/react-instantsearch-dom/src/widgets/__tests__/DynamicWidgets.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+
+const createSearchClient = () => ({
+  search: jest.fn(() => Promise.resolve(EMPTY_RESPONSE)),
+  searchForFacetValues: jest.fn(() => Promise.resolve({})),
+});
+
+const RefinementList = connectRefinementList(
+  ({ attribute }) => `RefinementList(${attribute})`
+);
+
```
francoischalifour, line 40: Wondering if the whole test suite makes sense here since we rely on the core `DynamicWidgets` which is also part of this repo.

Haroenv: So I understand it right, you suggest to only test the class part in the dom tests?

francoischalifour: `className`, and that it forwards the props and the children to the core component maybe?

Haroenv:
Thanks for your thorough reviews!

PR #3059, opened 2021-06-08
fix(RefinementList): prevent searchable component to refine on empty list

shortcuts:
**Summary**

Prevent the `RefinementList` component to refine when no items are in the list.

**Reproduction**
- Go to https://react-instantsearch.netlify.app/storybook/iframe.html?id=refinementlist--with-search-inside-items
- Search for anything that renders the `noResults` component
- Press <kbd>enter</kbd>

The component tries to refine even though the list is empty (`Cannot read property 'value' of undefined`), forcing the user to type again to make the facets reappear.

**Result**

Pressing <kbd>enter</kbd> when the list is empty won't do anything.

**Test**

_Following [this comment](https://github.com/algolia/react-instantsearch/pull/3045#issuecomment-855806807), it seems that our storybook don't pick local changes, so you need to change the import in the story._

`yarn storybook`

### DIFF
--- a/packages/react-instantsearch-dom/src/components/List.js
+++ b/packages/react-instantsearch-dom/src/components/List.js
@@ -131,7 +131,7 @@ class List extends Component {
           onSubmit={e => {
             e.preventDefault();
             e.stopPropagation();
-            if (isFromSearch) {
+            if (isFromSearch && items.length > 0) {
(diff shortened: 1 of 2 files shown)


francoischalifour (approved):

Haroenv (approved):
great catch!

PR #3066, opened 2021-06-24
feat(DynamicWidgets): add fallbackComponent

Haroenv:
This was brought up by @Shipow as a possible use case: you want to have dynamic widgets, but don't know what the facets could be in advance.

The fallbackWidget prop caters for this use case, allowing you to special-case some widgets (like hierarchical for example), but use the same widget for the rest.

```jsx
const Defaults = () => (
  <ExperimentalDynamicWidgets fallbackComponent={RefinementList}>
    <HierarchicalMenu attributes={['hierarchy', 'subHierarchy']} />
  </ExperimentalDynamicWidgets>
);

const Custom = () => (
  <ExperimentalDynamicWidgets
    fallback fallbackComponent={({ attribute }) => <CustomWidget facet={attribute} />}
  >
    <HierarchicalMenu attributes={['hierarchy', 'subHierarchy']} />
  </ExperimentalDynamicWidgets>
);
```

## Alternatives

1. fallback
2. fallbackWidget
2. fallbackComponent
3. function as a child
4. widgetComponent
5. component
6. itemComponent
7. attributeComponent

### DIFF
--- a/packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
+++ b/packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment, ReactChild, ReactNode } from 'react';
+import React, { Fragment, ReactChild, ComponentType, ReactNode } from 'react';
 import { getDisplayName } from '../core/utils';
 import connectDynamicWidgets from '../connectors/connectDynamicWidgets';
 
@@ -23,9 +23,14 @@ function getAttribute(component: ReactChild): string | undefined {
 type DynamicWidgetsProps = {
   children: ReactNode;
   attributesToRender: string[];
+  fallbackWidget?: ComponentType<{ attribute: string }>;
 };
 
-function DynamicWidgets({ children, attributesToRender }: DynamicWidgetsProps) {
+function DynamicWidgets({
+  children,
+  attributesToRender,
+  fallbackWidget: Fallback = () => null,
+}: DynamicWidgetsProps) {
   const widgets: Map<string, ReactChild> = new Map();
 
   React.Children.forEach(children, child => {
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-core/src/widgets/__tests__/DynamicWidgets.test.tsx
```diff
@@ -39,7 +40,12 @@ const RefinementList = connectRefinementList(
 );
 
 const HierarchicalMenu = connectHierarchicalMenu(
-  ({ attributes }) => `HierarchicalMenu(${attributes.join(',')})`
+  ({ attributes }) => `HierarchicalMenu([${attributes.join(', ')}])`
```
eunjae-lee, line 43: I really like this pattern for tests!

packages/react-instantsearch-core/src/widgets/DynamicWidgets.tsx
```diff
@@ -23,9 +23,14 @@ function getAttribute(component: ReactChild): string | undefined {
...
   children: ReactNode;
   attributesToRender: string[];
+  fallbackWidget?: ComponentType<{ attribute: string }>;
 };
 
-function DynamicWidgets({ children, attributesToRender }: DynamicWidgetsProps) {
+function DynamicWidgets({
+  children,
+  attributesToRender,
+  fallbackWidget: Fallback = () => null,
```
eunjae-lee, line 32: `widget` doesn't sound like React. Maybe just `fallback` or `fallbackComponent` (as in `hitComponent` in `Hits`)?

Haroenv: good point! I put this in the description as well, since I'm not sure what's best. At the moment I'm edging towards fallbackComponent

eunjae-lee (approved):

francoischalifour (approved):

PR #3067, opened 2021-06-25
feat(facets): add a new option "facetOrdering" to Menu, RefinementList & HierarchicalMenu

Haroenv:
If `facetOrdering` is enabled (the default behaviour), before the default sortBy is used, the result from renderingContent.facetOrdering.values is first checked. If that's present, it will be used to sort the items.

You can still change that ordering afterwards with the existing transformItems, so if you are sorting in transformItems, you actually override the sorting done by facet ordering, and won't see the effect. To use facetOrdering, you thus need to remove any sorting done in transformItems.

If there is a facetOrdering present in the index, but you don't want to use it for a certain widget, you need to explicitly pass `facetOrdering: false` to the widget or connector

References:
- [RFC 45](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/flexible-facet-values.md)
- https://github.com/algolia/instantsearch.js/pull/4784
- https://github.com/algolia/algoliasearch-helper-js/pull/822

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectHierarchicalMenu.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectHierarchicalMenu.js
@@ -1,4 +1,4 @@
-import { SearchParameters } from 'algoliasearch-helper';
+import { SearchResults, SearchParameters } from 'algoliasearch-helper';
 import connect from '../connectHierarchicalMenu';
 
 jest.mock('../../core/createConnector', () => x => x);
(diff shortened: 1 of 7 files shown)


eunjae-lee:
sorry, i'll review this tomorrow morning!

packages/react-instantsearch-core/src/connectors/__tests__/connectHierarchicalMenu.js
```diff
@@ -174,6 +174,158 @@ describe('connectHierarchicalMenu', () => {
...
+          count: 10,
+          isRefined: false,
+          items: null,
+        },
+        {
+          label: 'wat',
+          value: undefined,
+          count: 20,
+          isRefined: true,
+          items: [
```
eunjae-lee, line 309:
```suggestion
          items: [ // sorted by count:desc by default
```

WDYT? That context is missing in either comment or in the test name, so I had to think hard to remember it.

eunjae-lee (approved):
This looks good to me 🙌🏻

francoischalifour (approved):

packages/react-instantsearch-core/src/connectors/__tests__/connectHierarchicalMenu.js
Haroenv: it's actually alphabetical, but in this small set it's also count

eunjae-lee (approved):

PR #3072, opened 2021-06-30
fix(HitsPerPage): Adds `id` prop to `HitsPerPage`, `Select` components

joshpensky:
**Summary**

This PR adds an `id` prop to the `HitsPerPage` component, in order to associate the element with a `label`. To accomplish this, the PR also adds an `id` prop to the `Select` component

This PR is a continuation of the work from #3068 

**Result**

1. Ran `npm link` in `packages/react-instantsearch-dom` and root directory to sync package update
2. Added an `id` to the default `HitsPerPage` Storybook story
3. Confirmed the `id` prop appears in DOM

Screen Shot 2021-06-30 at 9 07 09 AM

### DIFF
--- a/packages/react-instantsearch-dom/src/components/HitsPerPage.js
+++ b/packages/react-instantsearch-dom/src/components/HitsPerPage.js
@@ -8,6 +8,7 @@ const cx = createClassNames('HitsPerPage');
 
 class HitsPerPage extends Component {
   static propTypes = {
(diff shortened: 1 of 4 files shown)


Haroenv (approved):
looks good! will probably need to fix conflicts once we merge

Haroenv (approved):

PR #3078, opened 2021-07-07
fix(server side rendering): return a value from mock currentRefinement/metadata

Haroenv:
This is required, as transitionState gets called with the return of value in a createURL. 

Without this fix, an error happens if there is:
-  dynamic configure
- currentRefinements + createURL
- something is refined
- SSR
- first client render

The error is `TypeError: Cannot read properties of undefined (reading 'configure')` as the transitionState argument is given the result of createURL/currentRefinements, but that's undefined instead of an object. The state isn't actually transitioned, so an empty object works as a solution.

Here's a sandbox showing the existing state: https://codesandbox.io/s/charming-williamson-bifbf?file=/components/app.js:1158-1167
And that same sandbox fixed in this PR: https://codesandbox.io/s/bold-brook-gdyri

### DIFF
--- a/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
+++ b/packages/react-instantsearch-core/src/core/__tests__/createInstantSearchManager.js
@@ -388,7 +388,7 @@ describe('createInstantSearchManager', () => {
   });
 
   describe('metadata hydration', () => {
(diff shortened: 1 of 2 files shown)


shortcuts (approved):
This fix make sense to me

PR #3087, opened 2021-08-12
fix(props): accept objects for hitComponent

### LINKED ISSUE #3086: Type checking of hitComponent props of Hits and InfiniteHits should include elementType
**Describe the bug 🐛**

Passing a `hitComponent` that is not a functional component results in a type check warning.  Should the type be [elementType] (introduced in [15.7.0])?

[elementType]: https://github.com/facebook/prop-types/blob/d62a775b78085e38805d34f27295798bcf54987c/README.md#L88
[15.7.0]: https://github.com/facebook/prop-types/blob/d62a775b78085e38805d34f27295798bcf54987c/CHANGELOG.md#user-content-1570

**To Reproduce 🔍**

1. Create a non-functional component.  For example by using `React.memo(...)` which returns an object:

    ```javascript
   const MyHit = React.memo((props => (<div>{props.hit.name}</div>)));
   ```

2. Use the non-functional component in Hits or InfiniteHits:

   ```javascript
   <InfiniteHits hitComponent={MyHit} />
   ```

3. Perform search
4. Open console in dev toolbar to see warning.  For example:

   ```
   Warning: Failed prop type: Invalid prop `hitComponent` supplied to `InfiniteHits`.
       in InfiniteHits (created by Translatable(InfiniteHits))
       in Translatable(InfiniteHits) (created by AlgoliaInfiniteHits(Translatable(InfiniteHits)))
       in AlgoliaInfiniteHits(Translatable(InfiniteHits)) (created by Context.Consumer)
       in ConnectorWrapper (at Demo.js:62)
       in InstantSearch (at Demo.js:51)
       in section (at Demo.js:41)
       in Demo (at App.js:164)
   ```

A live example helps a lot! We have a simple online template for you to use for your explanations:

> https://codesandbox.io/s/github/algolia/create-instantsearch-app/tree/templates/react-instantsearch

https://codesandbox.io/s/vigilant-hofstadter-kzsib?file=/src/App.js

```
Warning: Failed prop type: Invalid prop `hitComponent` of type `object` supplied to `Hits`, expected `function`.
    in Hits (created by AlgoliaHits(Hits))
    in AlgoliaHits(Hits) (created by Context.Consumer)
    in ConnectorWrapper (at App.js:48)
    in div (at App.js:41)
    in div (at App.js:36)
    in InstantSearch (at App.js:35)
    in div (at App.js:34)
    in div (at App.js:21)
    in App (at src/index.js:6)
```

**Expected behavior 💭**

I expect there to be no warnings.

**Environment:**

All environments where warnings in console are visible.

**Additional context**

The current implementation in [InfiniteHits] also includes `PropTypes.string` which seems incorrect.

[InfiniteHits]: https://github.com/algolia/react-instantsearch/blob/6f868ee388bb47ba0ef1cea3e15e658f1859fe9d/packages/react-instantsearch-dom/src/components/InfiniteHits.js#L65

Haroenv:
**Summary**

hitComponent accepts all react elements, things like memo are actually objects, not functions.

**Result**

No console errors for memo components

fixes #3086

### DIFF
--- a/packages/react-instantsearch-dom/src/components/Hits.tsx
+++ b/packages/react-instantsearch-dom/src/components/Hits.tsx
@@ -62,11 +62,7 @@ Hits.propTypes = {
   className: PropTypes.string,
 
   // this is actually PropTypes.elementType, but our prop-types version is outdated
-  hitComponent: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.func,
-    PropTypes.any, // TypeScript won't accept PropTypes.object as a react element (memo)
-  ]),
+  hitComponent: PropTypes.any,
 };
 
 export default Hits;
--- a/packages/react-instantsearch-dom/src/components/Hits.tsx
+++ b/packages/react-instantsearch-dom/src/components/Hits.tsx
@@ -57,7 +57,11 @@ const HitPropTypes = PropTypes.shape({
 Hits.propTypes = {
   hits: PropTypes.arrayOf(HitPropTypes.isRequired).isRequired,
   className: PropTypes.string,
-  hitComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.func, PropTypes.object]),
+  hitComponent: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.func,
+    PropTypes.object
+  ]),
 };
 
 export default Hits;
--- a/packages/react-instantsearch-dom/src/components/Hits.tsx
+++ b/packages/react-instantsearch-dom/src/components/Hits.tsx
@@ -14,7 +14,10 @@ type HitProps = {
 type Props = {
   hits: Hit[];
   className?: string;
-  hitComponent?: React.FunctionComponent<HitProps>;
+  hitComponent?:
+    | string
+    | React.ReactType<HitProps>
+    | React.ExoticComponent<HitProps>;
 };
 
 const cx = createClassNames('Hits');
@@ -57,10 +60,12 @@ const HitPropTypes = PropTypes.shape({
 Hits.propTypes = {
   hits: PropTypes.arrayOf(HitPropTypes.isRequired).isRequired,
   className: PropTypes.string,
+  // PropTypes.elementType would be more correct, but it has more dependencies
+  // and bumps up the bundle size too much.
   hitComponent: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.func,
-    PropTypes.object,
+    PropTypes.any,
   ]),
 };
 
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-dom/src/components/InfiniteHits.js
```diff
@@ -62,7 +62,11 @@ InfiniteHits.propTypes = {
   refineNext: PropTypes.func.isRequired,
   translate: PropTypes.func.isRequired,
   className: PropTypes.string,
-  hitComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  hitComponent: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.func,
+    PropTypes.object
```
tkrugg, line 68:
```suggestion
    PropTypes.object // support components wrapped with React.memo
```

packages/react-instantsearch-dom/src/components/Hits.tsx
```diff
@@ -57,7 +57,11 @@ const HitPropTypes = PropTypes.shape({
 Hits.propTypes = {
   hits: PropTypes.arrayOf(HitPropTypes.isRequired).isRequired,
   className: PropTypes.string,
-  hitComponent: PropTypes.func,
+  hitComponent: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.func,
+    PropTypes.object
```
tkrugg, line 63:
```suggestion
    PropTypes.object // support components wrapped with React.memo
```

packages/react-instantsearch-dom/src/components/InfiniteHits.js
Haroenv: is this comment needed? Alternatively I also thought naming it something like ReactElement so it can be reused

packages/react-instantsearch-dom/src/components/InfiniteHits.js
```diff
@@ -62,7 +62,11 @@ InfiniteHits.propTypes = {
   refineNext: PropTypes.func.isRequired,
   translate: PropTypes.func.isRequired,
   className: PropTypes.string,
-  hitComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  hitComponent: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.func,
+    PropTypes.object,
```
tkrugg, line 68: Why is it `PropTypes.object` for infiniteHits and `PropTypes.any` in hits?

Haroenv: object isn't actually the right type (it needs to be a specific object with $$typeof), which isn't describable in PropTypes as well as TypeScript at the same time. As the prop types aren't really that important in this case, I went with the strictest way that still worked with both TS in hits, but a slightly stricter way in the .js one. 

As tried in a previous commit, the real right way is using elementType, but it increased the bundle size significantly

tkrugg: so why not put the same (though non ideal) type for both, just to keep it consistent? Is it possible?

Haroenv: sure, they could both be `any`

tkrugg (approved):

PR #3090, opened 2021-08-26
feat(DynamicWidgets): release as stable

Haroenv:
This widget is fully available now, and thus no longer experimental.

DX-2420

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,7 +150,7 @@
     },
     {
       "path": "packages/react-instantsearch/dist/umd/Dom.min.js",
-      "maxSize": "40 kB"
+      "maxSize": "41.50 kB"
     },
     {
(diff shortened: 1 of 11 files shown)


eunjae-lee (approved):
Let's go!

dhayab (approved):

package.json
```diff
@@ -122,8 +122,6 @@
     "react-autosuggest": "9.4.3",
     "react-dom": "16.14.0",
     "react-error-boundary": "3.1.3",
-    "react-instantsearch-dom": "6.12.0",
-    "react-instantsearch-dom-maps": "6.12.0",
```
francoischalifour, line 126: Why this change?

Haroenv: they didn't seem needed, but it caused the stories to import from node_modules instead of the local packages @francoischalifour

francoischalifour (approved):

PR #3128, opened 2021-09-10
feat(server): allow components to request an initial search

Haroenv:
**Summary**

dynamic widgets requires a search results to know which widgets to mount. This means that the initial search results are only used to *add widgets*, not to actually refine them for example.

**Result**

server-side rendering now will check if there's a dynamic widget added, and do an initial search after render, and pass those results to findResultsState again. This will do an additional query to ensure the results can be used to refine the search

Things to consider still:
- add analytics: false to the initial search
- add hitsPerPage: 1 or 0 to the initial search
- write tests etc.


examples/next/components/app.js
```diff
@@ -62,11 +72,14 @@ export default class extends React.Component {
         <Configure hitsPerPage={12} />
         <header>
           <h1>React InstantSearch + Next.Js</h1>
+          <a href="/?query=%20&page=1&configure%5BhitsPerPage%5D=12&refinementList%5Bbrand%5D%5B0%5D=Insignia%E2%84%A2">
+            refined link
+          </a>
```
Haroenv, line 77: remove before merging

examples/next/components/app.js
```diff
@@ -66,7 +65,7 @@ export default class extends React.Component {
         </header>
         <main>
           <div className="menu">
-            <RefinementList attribute="categories" />
+            <RefinementList attribute="brand" />
```
Haroenv, line 68:
```suggestion
            <RefinementList attribute="categories" />
```

packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
```diff
@@ -13,6 +13,11 @@ export default createConnector({
     transformItems: PropTypes.func,
   },
 
+  // count as a widget for server-side rendering
```
francoischalifour, line 16: Here's a suggestion, if I understood correctly:

```suggestion
  // This is required to flag as a widget for SSR to pick up its search parameters.
```

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -920,6 +953,40 @@ describe('findResultsState', () => {
...
+        getSearchParameters,
+      });
+      const Dynamic = createWidget({
+        getSearchParameters: getSearchParameters2,
+        requiresInitialResults: true,
+      });
+
+      const App = props => (
+        <InstantSearch {...props}>
+          <Dynamic />
```
francoischalifour, line 970: Why do we keep a `Dynamic` component in the root index in this test?

packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
```diff
@@ -34,4 +39,6 @@ export default createConnector({
       attributesToRender: props.transformItems(facetOrder, { results }),
     };
   },
+
+  requiresInitialResults: true,
```
francoischalifour, line 43: Couldn't we use the existing Metadata API to flag a widget that needs initial results to render? Wondering if we could avoid transporting the connector description in the whole data flow.

packages/react-instantsearch-core/src/widgets/Index.tsx
```diff
@@ -83,7 +83,8 @@ class Index extends Component<InnerProps, State> {
         multiIndexContext: this.state.indexContext,
       },
       this.props,
-      undefined
+      undefined,
+      { displayName: 'AlgoliaIndex', getProvidedProps() {} }
```
francoischalifour, line 87: This dummy data reveals a flaw in the solution IMO. See my comment about using the Metadata API for this.

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
Haroenv: to check it doesn't do three requests

packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
Haroenv: possible to put it in metadata, but that's used in current refinements, so I'll need to check a way to avoid it showing up there

packages/react-instantsearch-core/src/widgets/Index.tsx
Haroenv: yes, this doesn't really have a connector, could also make this key optional in the function

Haroenv: what do you have as alternative in mind? I also thought instead of the widget requesting the rerender, it could be an argument to findResultsState (something like "renderPasses: 2")

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
sarahdayan: Not sure I understand why it would do three requests and why having it at the root prevents it.

packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
sarahdayan: Why would it be a problem to have the flag exposed?

packages/react-instantsearch-core/src/core/createConnector.tsx
```diff
@@ -46,6 +46,12 @@ export type ConnectorDescription = {
    */
   propTypes?: {}; // I can't find a definition for a propTypes object
   defaultProps?: {};
+
+  /**
+   * this is true for conditional widgets that require search results before
+   * knowing the search parameters required. Example: dynamic widgets.
```
sarahdayan, line 52:
```suggestion
   * Necessary for conditional widgets that need search results
   * before knowing the required search parameters (e.g., `DynamicWidgets`).
```

packages/react-instantsearch-core/src/widgets/Index.tsx
sarahdayan: I like the approach of moving this to `findResultsState`, it feels less invasive and easier to control for the user.

sarahdayan: I like the approach of moving the logic to `findResultsState`. It feels less invasive and easier to control for users.

packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
Haroenv: I'm testing that dynamic widgets does 1 extra request, and that there's no mismatch and trying to do one extra request per dynamic widget

packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
Haroenv: because it's not a current refinement, so it shouldn't be an option in current refinements (that's what metadata is used for)

Haroenv:
This seems to controversial and no customers have asked, so I'll pause this work for now. In the mean time you can call `findResultsState` twice, like this: https://codesandbox.io/s/hungry-gagarin-eyim1?file=/pages/index.js:1171-1663

PR #3132, opened 2021-09-17
feat(v7): bootstrap Core package

francoischalifour:
## Description

This prepares the packages for the next version of React InstantSearch, which will expose hooks.

## What's included

- ESLint configuration
- Jest configuration
- Build configuration
- Dependency updates. The Babel upgrade has increased the bundle size of the existing packages. I think we can do a bundle optimization pass later on.

PR #3133, opened 2021-09-17
feat(v7): add InstantSearch and Index components

francoischalifour:
## Description

This implements the foundation of the React InstantSearch API: the `InstantSearch` and `Index` components.

## Usage

The usage is similar to React InstantSearch v6, except that it supports options from the [InstantSearch core](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/) widget.

```jsx
import { searchClient } from './searchClient';

function App() {
  return (
    <InstantSearch searchClient={searchClient} indexName="parentIndex">
      {/* Widgets */}

      <Index indexName="childIndex">{/* Widgets */}</Index>
    </InstantSearch>
  );
}
```

#### `InstantSearch`

The `InstantSearch` component is a React Context provider that provides the InstantSearch instance context value.

#### `Index`

The `Index` component is a React Context provider that provides the parent index context value.

### Data sharing

Data sharing via React Context


package.json
```diff
@@ -155,7 +155,7 @@
     },
     {
       "path": "packages/react-instantsearch-core-next/dist/umd/ReactInstantSearchCore.min.js",
-      "maxSize": "28.25 kB"
+      "maxSize": "40 kB"
```
francoischalifour, line 158: We need to do some work to reduce the InstantSearch Core package bundle size.

packages/react-instantsearch-core-next/src/useIndex.ts
```diff
@@ -0,0 +1,23 @@
...
+import { IndexWidgetParams } from 'instantsearch.js/es/widgets/index/index';
+import { useEffect, useMemo } from 'react';
+
+import { useIndexContext } from './useIndexContext';
+
+export type UseIndexProps = IndexWidgetParams;
+
+export function useIndex(props: UseIndexProps) {
+  const searchIndex = useIndexContext();
+  const widget = useMemo(() => index(props), [props]);
```
francoischalifour, line 12: We'll take care of using stable values to avoid infinite loops in another PR.

test/mock/createSearchClient.js
```diff
@@ -0,0 +1,26 @@
+import {
```
francoischalifour, line 1: The current Jest configuration doesn't support TypeScript files from this folder.

package.json
sarahdayan: Have you already identified what the culprits are in InstantSearch Core?

packages/react-instantsearch-core-next/src/SearchIndex.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
sarahdayan, line 1: Why call this file `SearchIndex` and return `Index`? Do we envision other kind of indices (e.g., `RecommendIndex`)?

If yes, I'd rename the exported component. If not, i'd rename the file.

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { IndexContext } from '../IndexContext';
+import { InstantSearch } from '../InstantSearch';
+import { Index } from '../SearchIndex';
+
+describe('Index', () => {
+  test('renders children', () => {
+    const searchClient = createSearchClient();
+
+    const { container } = render(
+      <InstantSearch indexName="indexName" searchClient={searchClient}>
```
sarahdayan, line 16: Do we still want to keep this pattern where we can provide an index name via a prop on `<InstantSearch>` or should we move to always using `<Index>`? Is it a requirement of InstantSearch Core to pass an `indexName`?

If there's a way around it, it could be nice to move to only using `Index` widgets in RIS. It's a lot clearer and makes a lot of sense with visually nested components.

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <Index indexName="childIndex">Children</Index>
```
sarahdayan, line 17: What about `name`? We're already on an `Index` component so it's implied.

packages/react-instantsearch-core-next/src/useIndex.ts
sarahdayan:
```suggestion
  const indexWidget = useMemo(() => index(props), [props]);
```

packages/react-instantsearch-core-next/src/useIndex.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  useEffect(() => {
+    searchIndex.addWidgets([widget]);
+
+    return () => {
+      searchIndex.removeWidgets([widget]);
+    };
+  }, [searchIndex, widget]);
+
+  return widget;
```
sarahdayan, line 22:
```suggestion
  useEffect(() => {
    searchIndex.addWidgets([indexWidget]);

    return () => {
      searchIndex.removeWidgets([indexWidget]);
    };
  }, [searchIndex, indexWidget]);

  return indexWidget;
```

test/mock/createSearchClient.js
francoischalifour: Handled in #3132.

test/mock/createSearchClient.js
```diff
@@ -0,0 +1,26 @@
...
+      Promise.resolve(
+        createMultiSearchResponse(
+          ...requests.map(() => createSingleSearchResponse())
+        )
+      )
+    ),
+    searchForFacetValues: jest.fn(() =>
+      Promise.resolve([createSFFVResponse()])
+    ),
+    ...args,
```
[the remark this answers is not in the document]
francoischalifour: Handled in #3132.

test/mock/createAPIResponse.js
```diff
@@ -0,0 +1,48 @@
...
+    hitsPerPage = 20,
+    hits = [],
+    nbHits = hits.length,
+    nbPages = Math.ceil(nbHits / hitsPerPage),
+    params = '',
+    exhaustiveNbHits = true,
+    exhaustiveFacetsCount = true,
+    processingTimeMS = 0,
+    ...rest
+  } = subset;
```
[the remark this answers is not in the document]
francoischalifour: Handled in #3132.

packages/react-instantsearch-core-next/src/SearchIndex.tsx
francoischalifour: Otherwise we get `index.ts` and `Index.tsx` in the same folder and it quickly gets confusing for the tooling. In fact, it triggers an error.

It's only the file name though, the component remains `Index`.

Another solution is not to use `index.ts` in the same folder. But the file architecture might change so I think we can talk about it later on.

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
francoischalifour: It is a requirement of InstantSearch Core, and it would be a consequent change.

Indeed, in the future it would be nice to remove this default index.

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
francoischalifour: It's using the same API as InstantSearch Core for alignment.

packages/react-instantsearch-core-next/src/SearchIndex.tsx
sarahdayan: No that makes sense to keep it like that indeed, I didn't think of filename case-sensitivity.

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
sarahdayan: Yes, and it makes sense to do it all at once in every InstantSearch libraries for consistency.

It isn't an emergency but we should keep this in mind. I opened a [ticket](https://algolia.atlassian.net/browse/FX-122) so we don't forget.

package.json
francoischalifour: My bad, we used both ESM and CJS imports, resulting in a duplicated InstantSearch.js bundle being used here.

So we're now down from 40KB to 22 KB, which is smaller than the current RIS Core (28KB). It gives us room to implement all the other connectors!

Bundle visualizer

test/mock/createAPIResponse.js
```diff
@@ -0,0 +1,48 @@
...
+    results: args.map(createSingleSearchResponse),
+  };
+};
+
+export const createSFFVResponse = (args = {}) => ({
+  facetHits: [],
+  exhaustiveFacetsCount: true,
+  processingTimeMS: 1,
+  ...args,
+});
```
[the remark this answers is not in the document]
francoischalifour: Handled in #3132.

packages/react-instantsearch-core-next/src/useIndex.ts
francoischalifour: See https://github.com/algolia/react-instantsearch/pull/3133/commits/ae7a19bac061cf366faaefa6b5f958ab1d2d5ac8.

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
sarahdayan: Gotcha. It should be easier to fix this one globally though, we might want to tackle it in the future majors.

test/mock/__tests__/createAPIResponse.test.js
```diff
@@ -0,0 +1,36 @@
+import { createSingleSearchResponse } from '../createAPIResponse';
+
+describe('createSingleSearchResponse', () => {
```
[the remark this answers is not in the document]
francoischalifour: Handled in #3132.

packages/react-instantsearch-core-next/src/IndexContext.ts
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+
+import type { IndexWidget } from 'instantsearch.js/es/widgets/index/index';
+
+export const IndexContext = createContext<null | IndexWidget>(null);
```
Haroenv, line 5: could we want to store anything else than the widget on the context? maybe in an object in that case?

Although YAGNI, it could be separate contexts so this is good!

packages/react-instantsearch-core-next/src/InstantSearch.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useInstantSearch } from './useInstantSearch';
+
+import type { UseInstantSearchProps } from './useInstantSearch';
+
+type InstantSearchProps = UseInstantSearchProps & {
+  children?: React.ReactNode;
+};
+
+export function InstantSearch({ children, ...props }: InstantSearchProps) {
+  const search = useInstantSearch(props);
```
Haroenv, line 14: do we have a use case for useInstantSearch and useIndex outside of this widget?

packages/react-instantsearch-core-next/src/InstantSearch.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <InstantSearchContext.Provider value={search}>
+      <IndexContext.Provider value={search.mainIndex}>
```
Haroenv, line 18: this is genius! it could be implemented in Vue InstantSearch too so the inject is no longer needing to fall back to mainIndex, but it's InstantSearch providing it!

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
Haroenv: it's interesting, but having a root index can be confusing, what results would eg. a hits on the root get? it would need logic added to make sure everything can render in some way without results as well, and a warning for the cases where you add no index at all

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
Haroenv: I don't think we need to shorten it, indexName is a single concept, the name sounds like a label instead of the Algolia concept imo

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    );
+
+    expect(indexContext).toEqual(
+      expect.objectContaining({
+        $$type: 'ais.index',
+        addWidgets: expect.any(Function),
+        removeWidgets: expect.any(Function),
+      })
+    );
+    expect(indexContext!.getIndexName()).toEqual('subchildIndex');
```
Haroenv, line 81: grandchild :trollface:

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+        removeWidgets: expect.any(Function),
+      })
+    );
+  });
+
+  test('attaches users agents', () => {
+    const searchClient = createSearchClient();
+
+    render(<InstantSearch indexName="indexName" searchClient={searchClient} />);
+
```
Haroenv, line 83:
```suggestion
    expect(searchClient.addAlgoliaAgent).toHaveBeenCalledTimes(2);
```

or is it called more times?

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+    expect(searchClient.addAlgoliaAgent).toHaveBeenCalledWith(
+      `react (${ReactVersion})`
+    );
+    expect(searchClient.addAlgoliaAgent).toHaveBeenCalledWith(
+      `react-instantsearch-core (${version})`
```
Haroenv, line 88: I think there's some regexes that require react-instantsearch, worth having the same as v6 for simplicity

packages/react-instantsearch-core-next/src/useIndex.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { useIndexContext } from './useIndexContext';
+
+import type { IndexWidgetParams } from 'instantsearch.js/es/widgets/index/index';
+
+export type UseIndexProps = IndexWidgetParams;
+
+export function useIndex(props: UseIndexProps) {
+  const searchIndex = useIndexContext();
+  const indexWidget = useMemo(() => index(props), [props]);
```
Haroenv, line 12: should we leave a comment that the memo is using a non-stable parameter? Are we fixing this somewhere consistently?

packages/react-instantsearch-core-next/src/useIndex.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useEffect, useMemo } from 'react';
+
+import { useIndexContext } from './useIndexContext';
+
+import type { IndexWidgetParams } from 'instantsearch.js/es/widgets/index/index';
+
+export type UseIndexProps = IndexWidgetParams;
+
+export function useIndex(props: UseIndexProps) {
+  const searchIndex = useIndexContext();
```
Haroenv, line 11:
```suggestion
  const parentIndex = useIndexContext();
```

this makes more sense to me

packages/react-instantsearch-core-next/src/useIndex.ts
```diff
@@ -0,0 +1,23 @@
+import index from 'instantsearch.js/es/widgets/index/index';
+import { useEffect, useMemo } from 'react';
+
+import { useIndexContext } from './useIndexContext';
+
+import type { IndexWidgetParams } from 'instantsearch.js/es/widgets/index/index';
```
Haroenv, line 6: should the imports be grouped somehow? is there good eslint rules for it (if I remember it correctly recently there were some added in the algolia config)

packages/react-instantsearch-core-next/src/useIndexContext.ts
```diff
@@ -0,0 +1,15 @@
+import { useContext } from 'react';
+
+import { IndexContext } from './IndexContext';
+
+export function useIndexContext() {
+  const context = useContext(IndexContext);
+
+  if (context === null) {
+    throw new Error(
+      '`useIndexContext` must be used within `IndexContext.Provider`.'
```
Haroenv, line 10: maybe this error should be "an algolia widget is mounted outside of an InstantSearch provider" (different wording probably) as that's most likely what's wrong? this error seems cryptic for an outsider

packages/react-instantsearch-core-next/src/useInstantSearch.ts
```diff
@@ -0,0 +1,31 @@
...
+import { useEffect, useMemo, version as ReactVersion } from 'react';
+
+import version from './version';
+
+import type { InstantSearchOptions } from 'instantsearch.js';
+
+export type UseInstantSearchProps = InstantSearchOptions;
+
+export function useInstantSearch(props: UseInstantSearchProps) {
+  const search = useMemo(() => instantsearch(props), [props]);
```
Haroenv, line 11: same question here, but the props here aren't only strings / stable objects 🤔

packages/react-instantsearch-core-next/src/InstantSearch.tsx
Haroenv: ah for useInstantSearch it could be for refreshing

packages/react-instantsearch-core-next/src/useInstantSearch.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (typeof props.searchClient.addAlgoliaAgent === 'function') {
+      props.searchClient.addAlgoliaAgent(`react (${ReactVersion})`);
+      props.searchClient.addAlgoliaAgent(
+        `react-instantsearch-core (${version})`
+      );
+    }
+  }, [props.searchClient]);
+
+  useEffect(() => {
+    search.start();
```
Haroenv, line 23: future: we could have a prop/way to *not* start automatically, but only once people request it, so that search interfaces that don't display anything on empty query don't have to use the conditional requests guide

Haroenv (commented):
great push!

packages/react-instantsearch-core-next/src/useIndex.ts
```diff
@@ -1,4 +1,4 @@
-import index from 'instantsearch.js/cjs/widgets/index/index';
+import index from 'instantsearch.js/es/widgets/index/index';
```
Haroenv, line 1: we should add a local eslint rule to only use es imports so this can't happen by accident (although bundle size would show it I guess)

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
sarahdayan: The way I saw it, we'd drop `indexName` on the `<InstantSearch>` and have the same behavior by having a root `<Index>`. If you add a `<Hits>` widget not nested in an `<Index>` one it wouldn't work, just like right now you need to pass an `indexName`.

If you want to do multi-index you'd have multiple parallel `<Index>` widgets. You could imagine having authorized widgets outside of `<Index>` (e.g., `<Configure>`) that would be inherited, or overriden by local ones.

packages/react-instantsearch-core-next/src/__tests__/Index.test.tsx
sarahdayan: Even if it's a concept, I think there's redundancy in having the term while you're on an `<Index>` widget. We might just bee too used to calling it like that.

packages/react-instantsearch-core-next/src/useInstantSearch.ts
sarahdayan: So basically the user would have to manually start on the first input (either from a route or an input) but do it conditionally (based on `search.started`) right?

How would you see it exposed? In lifecycles?

packages/react-instantsearch-core-next/src/IndexContext.ts
francoischalifour: Yeah I think it would be separate contexts. Let's see how it evolves during the alpha!

packages/react-instantsearch-core-next/src/InstantSearch.tsx
francoischalifour: I don't think so. Users would use `useInstantSearchContext` and `useIndexContext` (not exported yet).

For refreshing, I think we can expose its own hook like `const { refresh } = useInstantSearchCache()`.

packages/react-instantsearch-core-next/src/InstantSearch.tsx
francoischalifour: Yeah let's use the framework!

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
francoischalifour: For now it's called more because props are not stable (handled in the next PR!).

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
francoischalifour: Ah I just realized in v6 we add `react-instantsearch`, and not `react-instantsearch-core`. I'll also add a specific one for this package, that I'll call `react-instantsearch-hooks` for now.

packages/react-instantsearch-core-next/src/useIndex.ts
francoischalifour: I already addressed this in the PR that's coming.

packages/react-instantsearch-core-next/src/useIndex.ts
francoischalifour: That's confusing to me since you're in `useIndex`, and the "parent index" here makes me think that's the parent index of this index.

packages/react-instantsearch-core-next/src/useIndex.ts
francoischalifour: They're already grouped following our ESLint rule. I'll revisit the config though because it's weird here indeed.

packages/react-instantsearch-core-next/src/useIndexContext.ts
francoischalifour: Right now `useIndexContext` is not exported, but I agree that once it is we need to gather the most frequent misuses to have an appropriate error.

packages/react-instantsearch-core-next/src/useInstantSearch.ts
Haroenv: It can definitely be done at a later point, but I saw it as "manual instantsearch" (component or hook) without the effect

packages/react-instantsearch-core-next/src/useIndex.ts
Haroenv: isn't it the parent of the index?

packages/react-instantsearch-core-next/src/useIndexContext.ts
Haroenv: ah yes, it's only an index that has this error, not a widget

packages/react-instantsearch-core-next/src/useIndex.ts
francoischalifour: Ugh, you're right!

packages/react-instantsearch-core-next/src/useInstantSearch.ts
sarahdayan: Yes, I can definitely see an exposed `useInstantSearch` hook (probably with a different name) used along with a `ControlledInstantSearch` widget.

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
francoischalifour: Added in TODO in fa71236b725a27cba12e4588a31a3adf57676aba.

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
francoischalifour: See afcd66a7c974e03e8d7914753f4711dcbdb156e1 and ba62c3516fac363e83942552a3200dc71654018d.

packages/react-instantsearch-core-next/src/useInstantSearch.ts
francoischalifour: I already addressed this in the PR that's coming.

packages/react-instantsearch-core-next/src/useIndex.ts
francoischalifour: See 0b0b82bfa339ec35f389d539841b6f6e7c509130.

packages/react-instantsearch-core-next/src/useIndex.ts
francoischalifour: Good idea! See b2e23190855afa617b3e09f56e2330e1c17e11c7.

Haroenv (approved):

sarahdayan (approved):

PR #3137, opened 2021-09-27
feat(hooks): introduce `useConnector`

francoischalifour:
## Description

This introduces `useConnector` which creates the bridge between InstantSearch.js connectors and this new React InstantSearch version.

This is the bare minimum for users to start using this version.

## Usage

Here's a basic version of a `useSearchBox` hook, as we'll provide in next PRs:

```tsx
import connectSearchBox from 'instantsearch.js/es/connectors/search-box/connectSearchBox';
import { useConnector } from 'react-instantsearch-hooks';

function useSearchBox(props) {
  return useConnector(connectSearchBox, props, {
    query: '',
    refine: () => {},
    clear: () => {},
    isSearchStalled: false,
  });
}

function SearchBox(props) {
  const { query, refine, clear, isSearchStalled } = useSearchBox(props);

  return <div>{/* ... */}</div>;
}
```

In our implementation, we'll rely on the `helper` state to compute the initial state.

### DIFF
--- a/packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
+++ b/packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
@@ -81,7 +81,7 @@ describe('InstantSearch', () => {
 
     render(<InstantSearch indexName="indexName" searchClient={searchClient} />);
 
-    // @TODO once props are stable:
+    // @TODO: figure out why it's called 6 times instead of 3.
     // expect(searchClient.addAlgoliaAgent).toHaveBeenCalledTimes(3);
     expect(searchClient.addAlgoliaAgent).toHaveBeenCalledWith(
       `react (${ReactVersion})`
--- a/packages/react-instantsearch-core-next/src/__tests__/useConnector.test.tsx
+++ b/packages/react-instantsearch-core-next/src/__tests__/useConnector.test.tsx
@@ -1,0 +1,70 @@
+import { renderHook } from '@testing-library/react-hooks';
+import React from 'react';
+
+import { createSearchClient } from '../../../../test/mock';
+import { InstantSearch } from '../InstantSearch';
+import { useConnector } from '../useConnector';
+import { noop } from '../utils';
+
+import type { Connector } from 'instantsearch.js';
+
+type CustomSearchBoxConnector = {
+  $$type: 'test.searchBox';
+  renderState: {
+    query: string;
+  };
+};
+
+const connectCustomSearchBox: Connector<
+  CustomSearchBoxConnector,
+  Record<string, never>
+> =
+  (renderFn, unmountFn = noop) =>
+  (widgetParams) => ({
+    $$type: 'test.searchBox',
+    init({ instantSearchInstance }) {
+      renderFn(
+        { query: 'query at init', instantSearchInstance, widgetParams },
+        true
+      );
+    },
+    render({ instantSearchInstance }) {
+      renderFn({ query: 'query', instantSearchInstance, widgetParams }, false);
+    },
+    dispose() {
+      unmountFn();
+    },
+  });
+
+describe('useConnector', () => {
+  test('returns the connector render state', async () => {
+    const searchClient = createSearchClient();
+    const wrapper = ({ children }) => (
+      <InstantSearch searchClient={searchClient} indexName="indexName">
+        {children}
+      </InstantSearch>
+    );
+    const { result, waitForNextUpdate } = renderHook(
+      () =>
+        useConnector(
+          connectCustomSearchBox,
+          {},
+          {
+            query: '',
+          }
+        ),
+      { wrapper }
+    );
+
+    // Initial render state
+    expect(result.current).toEqual({ query: '' });
+
+    await waitForNextUpdate();
+
+    // It should never be "query at init" because we skip the `init` step.
+    expect(result.current).not.toEqual({ query: 'query at init' });
+
+    // Render state provided by InstantSearch Core during `render`.
+    expect(result.current).toEqual({ query: 'query' });
+  });
+});
--- a/packages/react-instantsearch-core-next/src/useConnector.ts
+++ b/packages/react-instantsearch-core-next/src/useConnector.ts
@@ -1,0 +1,54 @@
+import { useEffect, useState } from 'react';
+
+import { useIndexContext } from './useIndexContext';
+import { useStableValue } from './useStableValue';
+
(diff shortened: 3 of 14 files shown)


packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
```diff
@@ -81,7 +81,7 @@ describe('InstantSearch', () => {
 
     render(<InstantSearch indexName="indexName" searchClient={searchClient} />);
 
-    // @TODO once props are stable:
+    // @TODO: figure out why it's called 6 times instead of 3.
     // expect(searchClient.addAlgoliaAgent).toHaveBeenCalledTimes(3);
```
francoischalifour, line 85: I need to figure out why it's called 6 times instead of 3. Any ideas yet?

packages/react-instantsearch-core-next/src/useStableValue.ts
```diff
@@ -0,0 +1,15 @@
+import { dequal } from 'dequal/lite';
+import { useEffect, useState } from 'react';
+
+export function useStableValue<TValue>(value: TValue) {
```
francoischalifour, line 4: We use the same implementation as in Recommend UI: https://github.com/algolia/recommend/blob/b3195f9c07d09a3f97b9ca980cfadedafd3a9913/packages/recommend-react/src/useStableValue.ts

packages/react-instantsearch-core-next/src/useIndex.ts
```diff
@@ -2,14 +2,16 @@ import index from 'instantsearch.js/es/widgets/index/index';
...
 export type UseIndexProps = IndexWidgetParams;
 
 export function useIndex(props: UseIndexProps) {
   const parentIndex = useIndexContext();
-  const indexWidget = useMemo(() => index(props), [props]);
+  const stableProps = useStableValue(props);
+  const indexWidget = useMemo(() => index(stableProps), [stableProps]);
 
   useEffect(() => {
     parentIndex.addWidgets([indexWidget]);
```
Haroenv, line 17: does this also happen twice, or is it only in InstantSearch?

packages/react-instantsearch-core-next/src/useConnector.ts
```diff
@@ -0,0 +1,54 @@
...
+} from 'instantsearch.js';
+
+type ConnectorState<TProps> = Omit<
+  WidgetRenderState<any, TProps>,
+  'widgetParams'
+>;
+
+export function useConnector<
+  TProps extends Record<string, unknown>,
+  TState extends ConnectorState<TProps>,
```
Haroenv, line 19: isn't render state already part of the widget description? 

```ts
TDescription['renderState']
```

packages/react-instantsearch-core-next/src/__tests__/useConnector.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    render({ instantSearchInstance }) {
+      renderFn({ query: 'query', instantSearchInstance, widgetParams }, false);
+    },
+    dispose() {
+      unmountFn();
+    },
+  });
+
+describe('useConnector', () => {
+  test('returns the connector render state', async () => {
```
Haroenv, line 40: we should also test that the widget has been unmounted and mounted to the parent (the right parent) or do you consider that testing React?

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
Haroenv: the other three tiems are:

```js
  [
      [ 'instantsearch.js (4.30.0)' ],
      [ 'react (16.14.0)' ],
      [ 'react-instantsearch (7.0.0-alpha.0)' ],
      [ 'react-instantsearch-hooks (7.0.0-alpha.0)' ],
      [ 'JS Helper (3.5.5)' ],
      [ 'JS Helper (3.5.5)' ]
    ]
```

not really sure how to enforce that as those other calls are also valid (and the client deduplicates anyway). We could assert those other calls too, or not consider it important there's 3 (and not more) calls

packages/react-instantsearch-core-next/src/useConnector.ts
```diff
@@ -0,0 +1,54 @@
...
+      // - We rely on a computed initial state before the InstantSearch Core lifecycle starts.
+      // - It prevents UI flashes when updating the widget props.
+      if (isFirstRender) {
+        return;
+      }
+
+      const { instantSearchInstance, widgetParams, ...renderState } =
+        connectorState;
+
+      setState(renderState as ConnectorState<TProps>);
```
sarahdayan, line 42: Why do you need to cast?

packages/react-instantsearch-core-next/src/__tests__/useConnector.test.tsx
francoischalifour: Yes, that's an important part indeed. Added a test in 87c3f0215ef0bc250098413de34a02aaa1c92301.

packages/react-instantsearch-core-next/src/useConnector.ts
francoischalifour: Indeed, thanks! See 32c4ed8fcf12a54045043ab233eb849243659503.

packages/react-instantsearch-core-next/src/useConnector.ts
francoischalifour: No need anymore since 32c4ed8fcf12a54045043ab233eb849243659503 🙂

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
Haroenv: maybe we could make the helper mocked for this test (in a separate file)?

packages/react-instantsearch-core-next/src/useStableValue.ts
francoischalifour: I'm preparing a test for it still.

packages/react-instantsearch-core-next/src/__tests__/InstantSearch.test.tsx
sarahdayan: Not sure it's necessary if we don't care how many times it's called. Just ensuring the calls with React, React InstantSearch and React InstantSearch Hooks should be enough to cover what we mean to assert here.

PR #3152, opened 2021-10-12
feat(hooks): introduce `useRefinementList`

francoischalifour:
This introduces the `useRefinementList` hook that is a bridge for `connectRefinementList`.

I added this official hook to validate the `useConnector` approach and as a convenience to use in our upcoming tests.


Haroenv:
@francoischalifour, as there's no sandbox yet, how should we try this out? I want to make sure initial results work when there's a refinement

Haroenv:
With initial ui state two network requests seem to be done, or is that because this doesn't have the InstantSearch update with deferred start yet? https://codesandbox.io/s/hooks-example-forked-7c3st?file=/App.tsx

francoischalifour:
@Haroenv Yes, this will be fixed with algolia/instantsearch.js#4925, I'm preparing a PR.

PR #3190, opened 2021-11-03
feat(hooks): introduce `useSortBy`

dhayab:
This adds `useSortBy` to our Hooks collection.

## API

This hook is a bridge to `connectSortBy`.

## Usage

```jsx
function SortBy(props) {
  const { currentRefinement, options, refine } = useSortBy(props);

  return <div>{/* ... */</div>;
}

function App(props) {
  return (
    <InstantSearch {...props}>
      {/* ... */}

      <SortBy items={[
        { label: 'Relevance', value: '...' },
        // ...
      ]} />
    </InstantSearch>
  );
}
```

### DIFF
--- a/examples/hooks/App.css
+++ b/examples/hooks/App.css
@@ -20,7 +20,7 @@ body {
   display: grid;
   align-items: flex-start;
   grid-template-columns: 200px 1fr;
-  gap: 0.5rem;
+  gap: .5rem;
 }
 
 .Search {
@@ -29,12 +29,10 @@ body {
 }
 
 .Search-header {
-  display: flex;
-}
-
-.SearchBox {
-  flex: 1;
-  margin-right: .5rem;
+  display: grid;
+  align-items: flex-start;
+  grid-template-columns: 1fr auto;
+  gap: .5rem;
 }
 
 .Hit-label {
--- a/examples/hooks/App.css
+++ b/examples/hooks/App.css
@@ -16,6 +16,20 @@ body {
   margin: 0 auto;
 }
 
+.SearchBox {
+  flex: 1;
+  margin-right: 1rem;
+}
+
+.Hit-label {
+  flex: 1;
+  margin-right: 1rem;
+}
+
+.Hit-price {
+  color: lightslategray;
+}
+
 .Pagination {
   padding: 1rem 0;
   margin: 0 auto;
--- a/examples/hooks/App.css
+++ b/examples/hooks/App.css
@@ -16,9 +16,25 @@ body {
   margin: 0 auto;
 }
 
-.SearchBox {
+.Container {
+  display: grid;
+  align-items: flex-start;
+  grid-template-columns: 200px 1fr;
+  gap: 0.5rem;
+}
+
+.Search {
+  display: grid;
+  gap: .5rem;
+}
+
+.Search-header {
+  display: flex;
+}
+
+.Search-box {
   flex: 1;
-  margin-right: 1rem;
+  margin-right: .5rem;
 }
 
 .Hit-label {
--- a/examples/hooks/App.css
+++ b/examples/hooks/App.css
@@ -32,7 +32,7 @@ body {
   display: flex;
 }
 
-.Search-box {
+.SearchBox {
   flex: 1;
   margin-right: .5rem;
 }
--- a/examples/hooks/App.tsx
+++ b/examples/hooks/App.tsx
@@ -9,6 +9,7 @@ import {
   Pagination,
   RefinementList,
   SearchBox,
+  SortBy,
 } from './components';
 
 import './App.css';
@@ -21,16 +22,21 @@ const searchClient = algoliasearch(
 type HitProps = {
   hit: AlgoliaHit<{
     name: string;
+    price: number;
   }>;
 };
 
 function Hit({ hit }: HitProps) {
   return (
-    <span
-      dangerouslySetInnerHTML={{
-        __html: (hit._highlightResult as any).name.value,
-      }}
-    />
+    <>
+      <span
+        className="Hit-label"
+        dangerouslySetInnerHTML={{
+          __html: (hit._highlightResult as any).name.value,
+        }}
+      />
+      <span className="Hit-price">${hit.price}</span>
+    </>
   );
 }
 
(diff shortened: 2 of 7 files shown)


packages/react-instantsearch-hooks/src/useSortBy.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useConnector } from './useConnector';
+
+import type {
+  SortByConnectorParams,
+  SortByWidgetDescription,
+} from 'instantsearch.js/es/connectors/sort-by/connectSortBy';
+
+export type UseSortByProps = SortByConnectorParams;
+
+export function useSortBy(props: UseSortByProps) {
```
Haroenv, line 12: this should be exported from the index

examples/hooks/App.tsx
```diff
@@ -56,7 +62,16 @@ export function App() {
           />
         </div>
         <div style={{ display: 'grid', gap: '.5rem' }}>
-          <SearchBox placeholder="Search" />
+          <div style={{ display: 'flex' }}>
```
Haroenv, line 65: this style probably should be in the css too for consistency? it's a little odd how it's mixed now

packages/react-instantsearch-hooks/src/useSortBy.ts
dhayab: Oh I haven't noticed, thanks!

examples/hooks/App.css
```diff
@@ -16,6 +16,36 @@ body {
...
+.Search {
+  display: grid;
+  gap: .5rem;
+}
+
+.Search-header {
+  display: flex;
+}
+
+.Search-box {
```
francoischalifour, line 35: To follow the component name:

```suggestion
.SearchBox {
```

examples/hooks/App.css
```diff
@@ -16,6 +16,36 @@ body {
...
+  flex: 1;
+  margin-right: .5rem;
```
francoischalifour, line 37: Since we use `gap` in `.Search`, we can also use it in `.Search-header`?

francoischalifour (approved):

Haroenv (approved):

examples/hooks/App.css
dhayab: Indeed! I adressed it in 6c516d5c. The grid layout removed the need to add styles for SearchBox, so that's also handled as well.

packages/react-instantsearch-hooks/src/__tests__/useSortBy.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { renderHook } from '@testing-library/react-hooks';
+
+import { createInstantSearchTestWrapper } from '../../../../test/utils';
+import { useSortBy } from '../useSortBy';
+
+const getSortByTestItems = () => [
+  { label: 'Featured', value: 'indexName' },
+  { label: 'Price (asc)', value: 'indexName_price_asc' },
+  { label: 'Price (desc)', value: 'indexName_price_desc' },
+];
```
sarahdayan, line 10: Why do you need to return the items from a function?

sarahdayan (commented):
When users interact with this widget (or any other for that matter), we'll perform a new search which means there will be a delay between the moment you interact and the moment you get the updated state. Hopefully this delay will be short, but on some slower connections, it might not (e.g., I have a _decent_ connection but I'm uploading a file right now and it made the demo slower to me).

It would be nice if we could also provide a way for users to know what's the current status of the search, similar to [what we provide on Autocomplete](https://www.algolia.com/doc/ui-libraries/autocomplete/core-concepts/state/#param-status). This way, they can update the UI to show something is going on. Maybe a `useSearchStatus` hook?

packages/react-instantsearch-hooks/src/__tests__/useSortBy.test.tsx
francoischalifour: Ah right I also wondered that, forgot to post the comment in the end.

francoischalifour:
@sarahdayan Yep, this is planned. I supported it in my original POC: https://github.com/francoischalifour/instantsearch-react-api/blob/872b7454c328b6f285e11b3bfde4f21bc2841529/lib/useSearchStatus.ts

packages/react-instantsearch-hooks/src/__tests__/useSortBy.test.tsx
dhayab: I (wrongly 😅 ) assumed the assertion would not check for deep equality, and I wanted to prevent sharing the same reference in the test. I checked jest and it indeed does deep comparison, I'll update the test accordingly.

packages/react-instantsearch-hooks/src/__tests__/useSortBy.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { renderHook } from '@testing-library/react-hooks';
+
+import { createInstantSearchTestWrapper } from '../../../../test/utils';
+import { useSortBy } from '../useSortBy';
+
+const getSortByTestItems = [
```
francoischalifour, line 6: We can rename this now. Suggestion:

```suggestion
const items = [
```

sarahdayan (approved):

sarahdayan:
@dhayab I just noticed you didn't add [documentation](https://github.com/algolia/react-instantsearch/tree/master/packages/react-instantsearch-hooks#api) for this new hook, can you add it in a separate PR?

dhayab:
@sarahdayan sorry, I didn't think about that, I'll get started on this now.

packages/react-instantsearch-hooks/src/__tests__/useSortBy.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    await waitForNextUpdate();
+
+    expect(result.current).toEqual({
+      currentRefinement: 'indexName',
+      options: items,
+      refine: expect.any(Function),
+      hasNoResults: expect.any(Boolean),
+    });
```
tkrugg, line 34: Sorry i'm a little late to the party 🥳

I'd love for us to add a test measuring the number of renders, preferably in way that's as close as possible from the real in-app use case.

I'm thinking for example how many re-render will this kind of code trigger, or how does it behave when the parent re-renders too often.

```ts
function MyComponent() {
   useSortBy({ 
         items: [ 
            { label: ... } /* <-- a new reference at each render */ 

         ] /* <-- a new reference at each render */ 
      }  /* <-- a new reference at each render */ 
   )
   ...
   return ...
}
```

@francoischalifour @dhayab

francoischalifour: `useConnector` [manipulates stable props](https://github.com/algolia/react-instantsearch/blob/6bd156c3977200946e254293ef3d170801bdf6ea/packages/react-instantsearch-hooks/src/useConnector.ts#L19) under the hood so users don't have to think about references. This avoids infinite loops and undesired rerenders.

There's a slight performance cost, but since our Connector props are basic types, it's negligible.

PR #3192, opened 2021-11-04
docs(hooks): add doc for useSortBy hook

dhayab:
**Summary**

Adds an API documentation for `useSortBy` in the README file for react-instantsearch-hooks.

### DIFF
--- a/packages/react-instantsearch-hooks/README.md
+++ b/packages/react-instantsearch-hooks/README.md
@@ -618,6 +618,92 @@ function RefinementList(props) {
 }
 ```
 
+### `useSortBy`
+
+> `(props: UseSortByProps) => SortByRenderState`
+
+Hook to [sort by](https://www.algolia.com/doc/api-reference/widgets/sort-by/js/) specified indices.
+
+**Types**
+
+<details>
+<summary><code>SortByItem</code></summary>
+
+```ts
+type SortByItem = {
+  /**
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-hooks/README.md
```diff
@@ -618,6 +618,92 @@ function RefinementList(props) {
 }
 ```
 
+### `useSortBy`
+
+> `(props: UseSortByProps) => SortByRenderState`
+
+Hook to [sort by](https://www.algolia.com/doc/api-reference/widgets/sort-by/js/) specified indices.
```
dhayab, line 625: Not sure about the wording I used here, so feel free to suggest something that would be a better match.

Haroenv: I think this wording is fine, otherwise the short description of the docs? 

> The sortBy widget displays a list of indices, allowing a user to change the way hits are sorted (with replica indices). Another common use case is to let the user switch between different indices.
> For this widget to work, you must define all indices that you pass down as options as replicas of the main index.

Haroenv (approved):

packages/react-instantsearch-hooks/README.md
francoischalifour: It's fine for now indeed!

francoischalifour (approved):

PR #3195, opened 2021-11-08
docs(hooks): update query on change only

francoischalifour:
The components from our Hooks example always triggered a query change on mount. This triggered additional renders, network requests, and created some routing issues by resetting the query.

### DIFF
--- a/examples/hooks/components/SearchBox.tsx
+++ b/examples/hooks/components/SearchBox.tsx
@@ -19,8 +19,14 @@ export function SearchBox(props?: SearchBoxProps) {
   }
 
   useEffect(() => {
-    refine(value);
-  }, [refine, value]);
+    if (query !== value) {
+      refine(value);
+    }
+    // We want to track when the value coming from the React state changes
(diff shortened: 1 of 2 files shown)


Haroenv (approved):

examples/hooks/components/SearchBox.tsx
```diff
@@ -19,8 +19,14 @@ export function SearchBox(props?: SearchBoxProps) {
   }
 
   useEffect(() => {
-    refine(value);
-  }, [refine, value]);
+    if (query !== value) {
```
Haroenv, line 22: we are relying on those both being `null` (or '') initially, right?

francoischalifour: Yes, `value` is initialized with InstantSearch's `query` ([which is an empty string initially](https://github.com/algolia/instantsearch.js/blob/29fafcaa60a0b004e808a5187bed47beca47e703/src/connectors/search-box/connectSearchBox.ts#L162)).

PR #3198, opened 2021-11-17
feat(hooks): add useRange

Haroenv:
**Summary**

implement useRange

**Result**

example implemented RangeInput with approximately the same code as existing React InstantSearch, with the value/start transformations done in InstantSearch as well: https://github.com/algolia/instantsearch.js/blob/dcbbd8804b4b6471d24820b42826b57388974c27/src/widgets/range-input/range-input.tsx#L144-L154

FX-443

### DIFF
--- a/examples/hooks/App.tsx
+++ b/examples/hooks/App.tsx
@@ -71,7 +71,7 @@ export function App() {
               showMore={true}
             />
           </Panel>
-          <Panel header="price">
+          <Panel header="Price">
             <RangeInput attribute="price" />
           </Panel>
         </div>
--- a/examples/hooks/App.tsx
+++ b/examples/hooks/App.tsx
@@ -48,7 +48,7 @@ function Hit({ hit }: HitProps) {
 
 export function App() {
   return (
-    <InstantSearch searchClient={searchClient} indexName="instant_search" routing>
+    <InstantSearch searchClient={searchClient} indexName="instant_search">
       <Configure hitsPerPage={15} />
 
       <div className="Container">
--- a/examples/hooks/App.tsx
+++ b/examples/hooks/App.tsx
@@ -9,6 +9,7 @@ import {
   Hits,
   Pagination,
   Panel,
+  RangeInput,
   RefinementList,
   SearchBox,
   SortBy,
@@ -47,7 +48,7 @@ function Hit({ hit }: HitProps) {
 
 export function App() {
   return (
-    <InstantSearch searchClient={searchClient} indexName="instant_search">
+    <InstantSearch searchClient={searchClient} indexName="instant_search" routing>
       <Configure hitsPerPage={15} />
 
       <div className="Container">
@@ -69,6 +70,9 @@ export function App() {
               ]}
               showMore={true}
             />
+          </Panel>
+          <Panel header="price">
+            <RangeInput attribute="price" />
           </Panel>
         </div>
         <div className="Search">
--- a/examples/hooks/components/RangeInput.tsx
+++ b/examples/hooks/components/RangeInput.tsx
@@ -55,7 +55,7 @@ export function RangeInput(props: RangeProps) {
           min={min}
           max={max}
           value={from}
-          placeholder={min?.toString()}
+          placeholder={min ? min.toString() : ''}
           disabled={!canRefine}
           onChange={(e) =>
             setState({ from: Number(e.currentTarget.value), to })
@@ -68,7 +68,7 @@ export function RangeInput(props: RangeProps) {
           min={min}
           max={max}
           value={to}
-          placeholder={max?.toString()}
+          placeholder={max ? max.toString() : ''}
           disabled={!canRefine}
           onChange={(e) =>
             setState({ from, to: Number(e.currentTarget.value) })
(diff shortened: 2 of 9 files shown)


examples/hooks/components/RangeInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+          refine([from, to]);
+        }}
+      >
+        <input
+          className={'ais-RangeInput-input ais-RangeInput-input--min'}
+          type="number"
+          min={min}
+          max={max}
+          value={from}
+          placeholder={min?.toString()}
```
dhayab, line 58: Apparently CSB doesn't play well with optional chaining (same for max input).

packages/react-instantsearch-hooks/README.md
```diff
@@ -468,6 +468,97 @@ function Hits(props) {
...
+};
+```
+
+</details>
+
+**Example**
+
+```jsx
+function Range(props) {
+  const { min, max } = useRange(props);
```
dhayab, line 554:
```suggestion
  const { range: { min, max } } = useRange(props);
```

packages/react-instantsearch-hooks/src/useRange.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useConnector } from './useConnector';
+
+import type {
+  RangeConnectorParams,
+  RangeWidgetDescription,
+} from 'instantsearch.js/es/connectors/range/connectRange';
+
+export type UseRangeProps = RangeConnectorParams;
+
+export function useRange(props?: UseRangeProps) {
```
dhayab, line 12: I think the `attribute` prop makes this a required argument.

dhayab (approved):

examples/hooks/App.tsx
```diff
@@ -70,6 +71,9 @@ export function App() {
               showMore={true}
             />
           </Panel>
+          <Panel header="price">
```
tkrugg, line 74:
```suggestion
          <Panel header="Price">
```

tkrugg (approved):

examples/hooks/components/RangeInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+    min:
+      minValue !== -Infinity && minValue !== min
+        ? minValue
+        : controlledUndefined,
+    max:
+      maxValue !== Infinity && maxValue !== max
+        ? maxValue
+        : controlledUndefined,
+  };
+  const [{ from, to }, setState] = useState({
```
francoischalifour, line 28: Naming suggestion, to narrow it down:

```suggestion
  const [{ from, to }, setRange] = useState({
```

examples/hooks/components/RangeInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+          max={max}
+          value={to}
+          placeholder={max ? max.toString() : ''}
+          disabled={!canRefine}
+          onChange={(e) =>
+            setState({ from, to: Number(e.currentTarget.value) })
+          }
+        />
+        <button className={'ais-RangeInput-submit'} type="submit">
+          Apply
```
francoischalifour, line 78: In IS.js [we use the word "Go"](https://instantsearchjs.netlify.app/stories/?path=/story/refinements-currentrefinements--with-rangeinput). Should we use the same here?

examples/hooks/components/RangeInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+      )}
+    >
+      <form
+        className={'ais-RangeInput-form'}
+        onSubmit={(e) => {
+          e.preventDefault();
+          refine([from, to]);
+        }}
+      >
+        <input
```
francoischalifour, line 52: We don't use the `step` property?

examples/hooks/components/RangeInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+      className={cx(
+        'ais-RangeInput',
+        !canRefine && 'ais-RangeInput-noRefinement',
+        props.className
+      )}
+    >
+      <form
+        className={'ais-RangeInput-form'}
+        onSubmit={(e) => {
+          e.preventDefault();
```
francoischalifour, line 48: For explicitness and consistency? (same below)

```suggestion
        onSubmit={(event) => {
          event.preventDefault();
```

examples/hooks/components/RangeInput.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { useEffect, useState } from 'react';
+
+import { useRange, UseRangeProps } from 'react-instantsearch-hooks';
+import { cx } from '../cx';
+
+export type RangeProps = React.ComponentProps<'div'> & UseRangeProps;
+
+// if the default value is undefined, React considers the component uncontrolled initially, which we don't want 0 or NaN as the default value
+const controlledUndefined = '' as unknown as number;
```
francoischalifour, line 9: Alternative name suggestion: `unsetRangeValue`

examples/hooks/components/RangeInput.tsx
Haroenv: can set it, I'm not really a fan on how precision gets turned into step in InstantSearch.js, seems like needless indirection

examples/hooks/components/RangeInput.tsx
Haroenv: copied from InstantSearch.css, same as InstantSearch.js is also good

PR #3203, opened 2021-11-22
docs(hooks): add Panel to example

Haroenv:
**Summary**

This isn't the full panel (doesn't communicate to hide or collapse), but does allow for more spacing between widgets. A more complete panel will require implementing a usePanel that we can do later, as the InstantSearch.js panel is only a widget, not a connector

**Result**

adds a minimal panel to the hooks example

FX-606

### DIFF
--- a/examples/hooks/components/Panel.tsx
+++ b/examples/hooks/components/Panel.tsx
@@ -11,7 +11,7 @@ export function Panel({
 }) {
   return (
     <div className="ais-Panel">
-      {header && <h2 className="ais-Panel-header">{header}</h2>}
+      {header && <div className="ais-Panel-header">{header}</div>}
       <div className="ais-Panel-body">{children}</div>
       {footer && <div className="ais-Panel-footer">{footer}</div>}
     </div>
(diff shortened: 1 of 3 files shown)


sarahdayan (approved):

examples/hooks/components/Panel.tsx
```diff
@@ -0,0 +1,19 @@
...
+  header,
+  footer,
+}: {
+  children: React.ReactNode;
+  header?: React.ReactNode;
+  footer?: React.ReactNode;
+}) {
+  return (
+    <div className="ais-Panel">
+      {header && <h2 className="ais-Panel-header">{header}</h2>}
```
francoischalifour, line 14:
```suggestion
      {header && <div className="ais-Panel-header">{header}</div>}
```

francoischalifour (commented):
For future reference, I have a similar implementation as in InstantSearch.js living in my original POC: https://github.com/francoischalifour/instantsearch-react-api/blob/872b7454c328b6f285e11b3bfde4f21bc2841529/lib/Panel.tsx

PR #3207, opened 2021-11-22
docs(hooks): make sure example runs without ts errors

Haroenv:
**Summary**

Fixes all TS errors in example repo

I don't understand why ci wasn't red before now, but in https://app.circleci.com/pipelines/github/algolia/react-instantsearch/1614/workflows/de241179-b866-4dc6-b576-0a4de456f2fb/jobs/47848 it is

**Result**

no more ts errors in hooks example and ci green


Haroenv:
can't tell whether this is the "false" passing tsc or the "right" passing one, and what makes CI pass even though tsc fails locally on master for me

examples/hooks/components/RefinementList.tsx
```diff
@@ -8,7 +8,10 @@ import { ControlledSearchBox } from './ControlledSearchBox';
 import { cx } from '../cx';
 
 export type RefinementListProps = React.ComponentProps<'div'> &
-  UseRefinementListProps;
+  UseRefinementListProps & {
+    searchable?: boolean;
+    searchablePlaceholder?: string;
```
francoischalifour, line 13: This should be in the IS.js types, right? Looks like we need to fix this at that level.

examples/hooks/components/SearchBox.tsx
```diff
@@ -39,12 +39,12 @@ export function SearchBox(props?: SearchBoxProps) {
 
   return (
     <ControlledSearchBox
-      className={props.className}
+      className={props && props.className}
```
francoischalifour, line 42: We should be able to avoid this if we replace the signature with:

```
SearchBox(props: SearchBoxProps)
```

examples/hooks/components/RefinementList.tsx
Haroenv: those are the InstantSearch widget types, not the ones from the connector

francoischalifour: So we can use the IS.js widget type for this component?

Haroenv: RefinementListWidgetParams contains unrelated types as well, are you suggesting `Pick<RefinementListWidgetParams, 'searchable' | 'searchablePlaceholder'>` ?

dhayab (approved):

PR #3212, opened 2021-11-26
feat(hooks): introduce `useQueryRules`

tkrugg:
This adds `useQueryRules` to our Hooks collection. Similarily to the current API, the example immplements
- `QueryRuleContext`: which applies ruleContexts based on filters to trigger context dependent Rules.
- `QueryRuleCustomData`:  which displays custom data from Rules.

This hook is a bridge to `connectQueryRules`.

Example of a simple banner

```js
export function Banner(props) {
  const { items } = useQueryRules({
    trackedFilters: {
      genre: () => ['Comedy', 'Thriller'],
    }
  });

  return <>
    { items.map(item => <a href={item.link}>{item.title}</a>) }
  </>
}

function App(props) {
  return (
    <InstantSearch {...props}>
      <Banner />

      {/* ... */}
    </InstantSearch>
  );
}
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,9 @@ jobs:
           name: Lint & Code styles
           command: yarn lint
       - run:
+          name: Build
+          command: yarn build
+      - run:
           name: Type checking
           command: yarn type-check
       - run:
--- a/examples/hooks/components/QueryRuleContext.tsx
+++ b/examples/hooks/components/QueryRuleContext.tsx
@@ -1,6 +1,9 @@
 import { useQueryRules, UseQueryRulesProps } from 'react-instantsearch-hooks';
 
-export type QueryRuleContextProps = Pick<UseQueryRulesProps, 'trackedFilters' | 'transformRuleContexts'>;
+export type QueryRuleContextProps = Pick<
+  UseQueryRulesProps,
+  'trackedFilters' | 'transformRuleContexts'
+>;
 
 export function QueryRuleContext(props: QueryRuleContextProps) {
   useQueryRules(props);
--- a/examples/hooks/components/QueryRuleContext.tsx
+++ b/examples/hooks/components/QueryRuleContext.tsx
@@ -1,0 +1,8 @@
+import { useQueryRules, UseQueryRulesProps } from 'react-instantsearch-hooks';
+
+export type QueryRuleContextProps = UseQueryRulesProps;
+
+export function QueryRuleContext(props: QueryRuleContextProps) {
+  useQueryRules(props);
+  return null;
+}
--- a/examples/hooks/components/QueryRuleContext.tsx
+++ b/examples/hooks/components/QueryRuleContext.tsx
@@ -1,6 +1,6 @@
 import { useQueryRules, UseQueryRulesProps } from 'react-instantsearch-hooks';
 
-export type QueryRuleContextProps = UseQueryRulesProps;
+export type QueryRuleContextProps = Pick<UseQueryRulesProps, 'trackedFilters' | 'transformRuleContexts'>;
 
 export function QueryRuleContext(props: QueryRuleContextProps) {
   useQueryRules(props);
--- a/examples/hooks/components/QueryRuleCustomData.tsx
+++ b/examples/hooks/components/QueryRuleCustomData.tsx
@@ -4,7 +4,7 @@ import { useQueryRules, UseQueryRulesProps } from 'react-instantsearch-hooks';
 import { cx } from '../cx';
 
 export type QueryRuleCustomDataProps = React.ComponentProps<'div'> &
-  UseQueryRulesProps & {
+  Pick<UseQueryRulesProps, 'transformItems'> & {
     children: (options: { items: any[] }) => React.ReactNode;
   };
 
--- a/examples/hooks/components/QueryRuleCustomData.tsx
+++ b/examples/hooks/components/QueryRuleCustomData.tsx
@@ -9,7 +9,7 @@ export type QueryRuleCustomDataProps = React.ComponentProps<'div'> &
   };
 
 export function QueryRuleCustomData(props: QueryRuleCustomDataProps) {
-  const { items } = useQueryRules();
+  const { items } = useQueryRules(props);
   return (
     <div className={cx('ais-QueryRuleCustomData', props.className)}>
       {props.children({ items })}
--- a/examples/hooks/components/QueryRuleCustomData.tsx
+++ b/examples/hooks/components/QueryRuleCustomData.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+
+import { useQueryRules, UseQueryRulesProps } from 'react-instantsearch-hooks';
+import { cx } from '../cx';
+
+export type QueryRuleCustomDataProps = React.ComponentProps<'div'> &
+  UseQueryRulesProps & {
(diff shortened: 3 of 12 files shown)


packages/react-instantsearch-hooks/README.md
```diff
@@ -1067,6 +1067,83 @@ function SortBy(props) {
...
+```ts
+type ParamTransformItems = TransformItems<any>;
+```
+</details>
+
+
+```jsx
+function Banner(props) {
+  const { items } = useQueryRules({
+    genre: () => ['Comedy', 'Thriller'],
```
Haroenv, line 1138:
```suggestion
    trackedFilters: {
      genre: () => ['Comedy', 'Thriller'],
    }
```

Haroenv: this probably should be two distinct examples

examples/hooks/components/QueryRuleContext.tsx
```diff
@@ -0,0 +1,8 @@
+import { useQueryRules, UseQueryRulesProps } from 'react-instantsearch-hooks';
+
+export type QueryRuleContextProps = UseQueryRulesProps;
```
Haroenv, line 3:
```suggestion
export type QueryRuleContextProps = Pick<UseQueryRulesProps, 'trackedFilters' | 'transformRuleContexts'>;
```

examples/hooks/components/QueryRuleCustomData.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+import { useQueryRules, UseQueryRulesProps } from 'react-instantsearch-hooks';
+import { cx } from '../cx';
+
+export type QueryRuleCustomDataProps = React.ComponentProps<'div'> &
+  UseQueryRulesProps & {
```
Haroenv, line 7:
```suggestion
  Pick<UseQueryRulesProps, 'transformItems'> & {
```

packages/react-instantsearch-hooks/src/__tests__/useQueryRules.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    };
+
+    const transformRuleContexts = (ruleContexts) =>
+      ruleContexts.map((ruleContext) => ruleContext.replace('ais-', 'custom-'));
+
+    const { result, waitForNextUpdate } = renderHook(
+      () =>
+        useQueryRules({
+          trackedFilters,
+          transformRuleContexts,
```
Haroenv, line 22: not using the `transformItems` prop

Haroenv (commented):
code lgtm, readme probably should classify the two components made with this

examples/hooks/components/QueryRuleCustomData.tsx
```diff
@@ -0,0 +1,18 @@
...
+    children: (options: { items: any[] }) => React.ReactNode;
+  };
+
+export function QueryRuleCustomData(props: QueryRuleCustomDataProps) {
```
francoischalifour, line 11: We can make this generic to type the items.

packages/react-instantsearch-hooks/README.md
```diff
@@ -1067,6 +1067,85 @@ function SortBy(props) {
...
+Hook to [query rules](https://www.algolia.com/doc/api-reference/widgets/query-rule-context/js/).
+
+**Types**
+
+<details>
+<summary><code>UseQueryRulesProps</code></summary>
+
+```ts
+type UseQueryRulesProps = {
+    trackedFilters?: ParamTrackedFilters;
```
francoischalifour, line 1083: This file doesn't look formatted.

packages/react-instantsearch-hooks/src/__tests__/useQueryRules.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    // Initial render state from manual `getWidgetRenderState`
+    expect(result.current).toEqual({
+      items: [],
+    });
+
+    await waitForNextUpdate();
+
+    // InstantSearch.js state from the `render` lifecycle step
+    expect(result.current).toEqual({
+      items: [],
```
francoischalifour, line 38: We can inject some rules in the API response, otherwise this test doesn't reveal much.

tkrugg: fix in b1e599bfde04a21be1178838bba0e0540c1d8acb

packages/react-instantsearch-hooks/README.md
tkrugg: fix in 7d9cb206

packages/react-instantsearch-hooks/src/__tests__/useQueryRules.test.tsx
tkrugg: fix in 4664c869

examples/hooks/components/QueryRuleCustomData.tsx
tkrugg: I think the type is actually changing because it entirely depends on what's inside your user Data. 
I don't think we commit to any structure at the moment: `{ title: string, link: string, banner: string }` only exists in our example, but you're allowed items of different types.

NB: same choice made here: https://github.com/algolia/instantsearch.js/blob/master/src/components/QueryRuleCustomData/QueryRuleCustomData.tsx#L20

francoischalifour: We don't expose the component types publicly in InstantSearch.js, which is why we didn't make it generic.

At least if we provide a way for users to type them, they can deal with their own types (with unions for instance). Right now they have no way to type their data.

tkrugg: It's also the connector: https://github.com/algolia/instantsearch.js/blob/master/src/connectors/query-rules/connectQueryRules.ts#L36

 To be clear, let's say:
```ts 
   userData: [{ banner: string }, { link: string }]
```

basically, we could do this

```ts
const { items } = useQueryRules<{ banner: string; } | { link: string }>(...)
// typeof items ({ banner: string; } | { link: string })[]
```

but only by introducing breaking changes on the connector types API in InstantSearch.js. Is that you meant?

francoischalifour: Well, we could only type the component but you're right, that would be a new concept (no BC required though), let's wait a little :)

.circleci/config.yml
```diff
@@ -50,6 +50,9 @@ jobs:
       - run:
           name: Lint & Code styles
           command: yarn lint
+      - run:
+          name: Build
+          command: yarn build
```
francoischalifour, line 55: Why this change? It'll slow down CI. (Good to merge besides that.)

francoischalifour (approved):

Haroenv (approved):

.circleci/config.yml
tkrugg: The type-check step, which comes right after this line, is failing without this change. (see [circleci build](https://app.circleci.com/pipelines/github/algolia/react-instantsearch/1686/workflows/c2edf0ca-d57f-4309-bb86-0043bd7fbf98/jobs/48288))
The type-check step will check types in the examples, based off types inside `react-instantsearch-hooks`.
So `react-instantsearch-hooks` need to be built.

PR #3216, opened 2021-11-29
feat(hooks): introduce `<DynamicWidgets>`

francoischalifour:
This adds the `<DynamicWidgets>` to the Hooks package. It's implemented on top of `useDynamicWidgets`.

`<DynamicWidgets>` is a widget that displays matching widgets based on by the corresponding settings of the index.

## Changes

I upgraded React from 16 to 17 because it better supports Strict Mode and memory leaks warnings, which `<DynamicWidgets>` was prone to. I explain in details how it happens in [`useConnector`](https://github.com/algolia/react-instantsearch/pull/3216/files#diff-acc4e0b4ef96a714bcc223f3ccbeb8be2f951d84564444b5a5a789d93ee0e275R34-R56).

## Usage

```js
function App(props) {
  return (
    <InstantSearch {...props}>
      <DynamicWidgets>
        <Panel header="Brands">
          <RefinementList attribute="brand" />
        </Panel>

        <Panel header="Categories">
          <Menu attribute="categories" />
        </Panel>

        <Panel header="Hierarchy">
          <HierarchicalMenu
            attributes={[
            ]}
          />
        </Panel>

        <Panel header="Price">
          <RangeInput attribute="price" />
        </Panel>
      </DynamicWidgets>
    </InstantSearch>
  );
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,7 +150,7 @@
     },
     {
       "path": "packages/react-instantsearch/dist/umd/Dom.min.js",
-      "maxSize": "41.50 kB"
+      "maxSize": "42 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
--- a/package.json
+++ b/package.json
@@ -158,7 +158,7 @@
     },
     {
       "path": "packages/react-instantsearch-hooks/dist/umd/ReactInstantSearchHooks.min.js",
-      "maxSize": "33.50 kB"
+      "maxSize": "34 kB"
     },
     {
       "path": "packages/react-instantsearch-dom/dist/umd/ReactInstantSearchDOM.min.js",
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "@wdio/selenium-standalone-service": "5.16.10",
     "@wdio/spec-reporter": "5.16.11",
     "@wdio/static-server-service": "5.16.10",
+    "@wojtekmaj/enzyme-adapter-react-17": "0.6.5",
     "algoliasearch": "4.10.5",
     "argos-cli": "0.1.3",
     "babel-eslint": "10.1.0",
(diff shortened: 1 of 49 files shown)


packages/react-instantsearch-hooks/src/DynamicWidgets.tsx
```diff
@@ -0,0 +1,86 @@
...
+\`\`\`
+<DynamicWidgets>
+  <MyComponent>
+    <RefinementList attribute="brand" />
+    <Menu attribute="categories" />
+  </MyComponent>
+</DynamicWidgets>
+\`\`\`
+`
+    );
```
francoischalifour, line 80: This is a new invariant that we don't yet have in the other flavors. It provides guidance on what to do in this scenario because `React.Children.only` throws a generic error when there are multiple children.

packages/react-instantsearch-hooks/src/useConnector.ts
```diff
@@ -17,23 +17,44 @@ export function useConnector<
...
-    });
+          // eslint-disable-next-line @typescript-eslint/no-use-before-define
+          setState(renderState);
+        }
+      },
+      () => {
+        // We'll ignore the next state update until we know for sure that
+        // InstantSearch.js re-inits the component.
+        shouldSetStateRef.current = false;
+      }
```
francoischalifour, line 56: This is the important part that avoids memory leaks warnings in React 16 and 17.

package.json
```diff
@@ -150,19 +150,19 @@
     },
     {
       "path": "packages/react-instantsearch/dist/umd/Dom.min.js",
-      "maxSize": "41.50 kB"
+      "maxSize": "42 kB"
```
Haroenv, line 153: react 17 increased this?

Haroenv (commented):
looks good!

package.json
francoischalifour: Yes

packages/react-instantsearch-hooks/src/__tests__/DynamicWidgets.test.tsx
```diff
@@ -0,0 +1,461 @@
...
+        <InstantSearchMock indexName="indexName" searchClient={searchClient}>
+          <DynamicWidgets transformItems={() => attributes}>
+            <RefinementList attribute="brand" />
+            <Menu attribute="categories" />
+          </DynamicWidgets>
+        </InstantSearchMock>
+      );
+    }
+
+    // @TODO: we swallow React act errors in this test because we're unable to
```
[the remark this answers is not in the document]
francoischalifour: I'm not sure. I don't understand why it that case it errors.

Haroenv (approved):
ok to merge as-is for now I think

PR #3217, opened 2021-11-30
fix(hooks): throw invariant violations in production

francoischalifour:
Our `invariant` util used to throw with a specified error message in development only, and keep the flow going without throwing in production. This is a dangerous behavior to have different flows in development and in production.

This diff now also throws in production, but with a tiny generic error message to keep the bundle size lite. In the future, we'll be able to implement our error strategy with error redirects (similar to React or Next.js) in this invariant.

For reference, [`tiny-invariant`](https://github.com/alexreardon/tiny-invariant) and other projects throw in both environments, only the error messages differ.

I added the production path in the test so that we're able to test that `invariant` acts as expected in production. `__DEV__` is `true` by default when running the test, but you can run the tests in production with the following command: `NODE_ENV=production yarn test invariant`

### DIFF
--- a/packages/react-instantsearch-hooks/src/utils/invariant.ts
+++ b/packages/react-instantsearch-hooks/src/utils/invariant.ts
@@ -7,11 +7,15 @@ export function invariant(
   condition: boolean,
   message: string | (() => string)
 ) {
-  if (!__DEV__) {
+  if (condition) {
     return;
   }
 
-  if (!condition) {
+  if (!__DEV__) {
+    throw new Error('Invariant failed');
+  }
+
+  if (__DEV__) {
     throw new Error(
       `[InstantSearch] ${typeof message === 'function' ? message() : message}`
     );
--- a/packages/react-instantsearch-hooks/src/utils/__tests__/invariant.test.ts
+++ b/packages/react-instantsearch-hooks/src/utils/__tests__/invariant.test.ts
@@ -1,6 +1,20 @@
 import { invariant } from '../invariant';
 
 describe('invariant', () => {
+  if (!__DEV__) {
+    test('throws a generic invariant violation in production', () => {
+      expect(() => {
+        invariant(false, 'invariant');
+      }).toThrow('Invariant failed');
+    });
+
+    test('does not throw when the condition is met in production', () => {
+      expect(() => {
+        invariant(true, 'invariant');
+      }).not.toThrow();
+    });
+  }
+
   if (__DEV__) {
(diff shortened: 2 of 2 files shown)


Haroenv (approved):

packages/react-instantsearch-hooks/src/utils/invariant.ts
```diff
@@ -7,11 +7,15 @@ export function invariant(
...
 ) {
-  if (!__DEV__) {
+  if (condition) {
     return;
   }
 
-  if (!condition) {
+  if (!__DEV__) {
+    throw new Error('Invariant failed');
+  }
```
sarahdayan, line 16: This is fine for now, but we should improve the DX at some point so users can easily debug production errors. Granted, they can reproduce in dev to get the full error message, but if they're using error reporting services like Sentry that stacks errors and labels them by error messages, it's really convenient to be able to get the main reason right away. Also useful in difficult to reproduce situations.

I like how [React ~~does~~ dit it](https://reactjs.org/docs/error-decoder.html/) (the decoder no longer seems to work). We could have good 'ol error numbers associated with invariants and have a comprehensive list documented with the full message, explanations of what went wrong, and possible leads to solve the problem.

sarahdayan (approved):

packages/react-instantsearch-hooks/src/utils/invariant.ts
Haroenv: it's mentioned in the description, but I indeed agree that the least we can do is have error codes, but a slightly more advanced code could collect all invariants and warnings, output them in a markdown or similar and have them accessible. Then the shortened message can include a link to that location.

In algolisearch v3 we actually had this system: 
- https://github.com/algolia/algoliasearch-client-javascript/wiki/Deprecated
- https://github.com/algolia/algoliasearch-client-javascript/blob/v3/src/deprecatedMessage.js#L6

sarahdayan: @Haroenv I agree, it would be nice to generate them. We could also leverage our `alg.li` to have shorter links and reduce size even more.

francoischalifour: Exactly, I touched this point in my PR description.

This change is just a more streamlined behavior for both development and production – we used to keep the flow going which is arguably worse because it could either end up generating cryptic errors or worse, end up in inconsistent behaviors that should have errored. This change is a preparation for better error management too.

PR #3221, opened 2021-12-03
feat(hooks): introduce SSR support

francoischalifour:
This introduces **server-side rendering** (SSR) support to React InstantSearch Hooks.

## Description

Server-side rendering (abbreviated "SSR") lets you generate HTML from InstantSearch components on the server.

Supporting SSR for InstantSearch:

- **Improves general performance**; the browser directly loads with HTML containing search results.
- **Improves perceived performance**; users won't see a UI flash when loading the page, but directly the search UI.
- **Improves SEO**; the content is crawlable by search engines.

Technically, SSR in InstantSearch consists of:

- On the server, retrieve the initial search results of the current search state (determined by the router, or more precisely the underlying UI state and its associated widgets)
- Then, on the server, render these initial search results to HTML and send the response
- Then, on the client, load the JavaScript code for InstantSearch
- Then, on the client, hydrate InstantSearch

_This PR doesn't implement hydrating the search client cache, which will come later. Note that this SSR solution still skips the initial network request on the browser, without relying on the search client cache, because it's based on algolia/instantsearch.js#4967._

## Usage

Here's an example with `react-instantsearch-hooks-server` on an [express](https://github.com/expressjs/express) server.

In `server.js`:

```js
import express from 'express';
import React from 'react';
import { renderToString } from 'react-dom/server';
import { getServerState } from 'react-instantsearch-hooks-server';
import App from './App';

const server = express();

server.get('/', async (_, res) => {
  const serverState = await getServerState(<App />);
  const html = renderToString(<App serverState={serverState} />);

  res.send(
    `
  <!DOCTYPE html>
  <html>
    <head>
      <script>window.__SERVER_STATE__ = ${JSON.stringify(serverState)}</script>
    </head>
    
    <body>
      <div id="root">${html}</div>
    </body>
    
    <script src="/app.js"></script>
  </html>
    `
  );
});

server.listen(8080);
```

In `browser.js`:

```js
import React from 'react';
import { hydrate } from 'react-dom';
import App from './App';

const SERVER_STATE = window.__SERVER_STATE__;

delete window.__SERVER_STATE__;

hydrate(
  <App serverState={SERVER_STATE} />,
  document.getElementById('root')
);
```

In `App.js`:

```js
import algoliasearch from 'algoliasearch/lite';
import React from 'react';
import {
  InstantSearch,
  InstantSearchSSRProvider,
} from 'react-instantsearch-hooks';

const searchClient = algoliasearch('YOUR_APP_ID', 'YOUR_API_KEY');

function Search() {
  return (
    <InstantSearch indexName="instant_search" searchClient={searchClient}>
      {/* ... */}
    </InstantSearch>
  );
}

export default function App({ serverState }) {
  return (
    <InstantSearchSSRProvider {...serverState}>
      <Search />
    </InstantSearchSsrProvider>
  );
}
```

### `getServerState(<App />)`

> `(children: React.ReactNode) => Promise<InstantSearchServerState>`

Function that returns the server state.

```js
import express from 'express';
import React from 'react';
import { renderToString } from 'react-dom/server';
import { getServerState } from 'react-instantsearch-hooks-server';
import App from './App';

const server = express();

server.get('/', async (_, res) => {
  const serverState = await getServerState(<App />);
  const html = renderToString(<App serverState={serverState} />);

  // ...
});
```

### `<InstantSearchSSRProvider {...serverState}>`

> `({ initialResults, children }: InstantSearchSSRProviderProps) => React.ReactNode`

React Context Provider that forwards the server state to `<InstantSearch>`.

```jsx
const searchClient = algoliasearch('YOUR_APP_ID', 'YOUR_API_KEY');

function App({ serverState }) {
  return (
    <InstantSearchSSRProvider {...serverState}>
      <InstantSearch searchClient={searchClient} indexName="indexName">
        {/* ... */}
      </InstantSearch>
    </InstantSearchSSRProvider>
  );
}
```

### Data flow

Here's the overall diagram that explains how SSR flows. To get into the specifics, you can refer to the [RFC](algolia/instantsearch-rfcs#58).

Diagram

### Exports breakdown

Here are the new APIs that are exported:

```js
import {
  // APIs
  InstantSearchServerContext, // used to flag that we're rendering on the server
  // Types (shared between `*-hooks` and `*-hooks-server`)
  InstantSearchServerContextApi,
  InstantSearchServerState,
} from 'react-instantsearch-hooks';
```

```js
import {
  // APIs
  getServerState,
} from 'react-instantsearch-hooks-server';
```

In the future React InstantSearch v7:

- `react-instantsearch-hooks` will be `react-instantsearch-core`
- the new `react-instantsearch-core` will be used in `react-instantsearch-dom`
- `react-instantsearch-hooks-server` will be `react-instantsearch-dom/server`

These future imports will therefore be the same as the current DOM v6, but using the new APIs.

## Next steps

- Add Hooks SSR example
- Add Hooks SSR Stream example (with the new React 18 SSR Stream APIs)
- Document these APIs in the new Algolia documentation Hooks flavor
- Write "Server-side rendering" guide

## Related

- algolia/instantsearch-rfcs#58
- algolia/instantsearch.js#4960
- algolia/instantsearch.js#4967


packages/react-instantsearch-hooks/package.json
```diff
@@ -42,10 +42,10 @@
     "@babel/runtime": "^7.1.2",
     "algoliasearch-helper": "^3.6.2",
     "dequal": "^2.0.0",
-    "instantsearch.js": "^4.32.0"
+    "instantsearch.js": "https://pkg.csb.dev/algolia/instantsearch.js/commit/3f954126/instantsearch.js"
```
francoischalifour, line 45: We'll be able to use the official InstantSearch.js dependency once algolia/instantsearch.js#4967 is released.

packages/react-instantsearch-hooks/package.json
```diff
@@ -42,10 +42,10 @@
...
   },
   "peerDependencies": {
     "algoliasearch": ">= 3.1 < 5",
-    "react": ">= 16.8.0 < 18"
+    "react": ">= 16.8.0 < 19"
```
francoischalifour, line 49: I tested the new APIs with React 18 and their new [SSR Stream APIs](https://github.com/reactwg/react-18/discussions/22) and they're compatible.

packages/react-instantsearch-hooks/src/useConnector.ts
```diff
@@ -104,9 +115,9 @@ export function useConnector<
     return {};
   });
 
-  // We use a layout effect to add the widget to the index before the index
-  // renders, otherwise it triggers 2 network requests.
-  useLayoutEffect(() => {
+  // Using a layout effect adds the widget at the same time as rendering, which
+  // triggers a single network request, instead of two with as regular effect.
+  useIsomorphicLayoutEffect(() => {
```
francoischalifour, line 120: `useLayoutEffect`s have been replaced by an internal `useIsomorphicLayoutEffect` to hide the warning triggered by React on the server

packages/react-instantsearch-hooks/src/useInstantSearch.ts
```diff
@@ -23,40 +37,89 @@ export function useInstantSearch({
...
-    forceUpdate();
+    // On SSR, the instance is already started so we don't start it again here.
+    if (!search.started) {
+      search.start();
+      forceUpdate();
+    }
 
     return () => {
       search.dispose();
-      forceUpdate();
```
francoischalifour, line 57: This lifecycle is now fully tested and this rerender is not required.

packages/react-instantsearch-hooks/src/useInstantSearch.ts
```diff
@@ -2,11 +2,25 @@ import instantsearch from 'instantsearch.js';
...
+import type { InstantSearchServerState } from './InstantSearchSSRProvider';
+import type {
+  InstantSearchOptions,
+  InstantSearch,
+  SearchClient,
+} from 'instantsearch.js';
+
+const defaultUserAgents = [
+  `react (${ReactVersion})`,
+  `react-instantsearch (${version})`,
```
Haroenv, line 21: React InstantSearch v6 added a specific server UA for the backend request. Do you think that's useful?

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -0,0 +1,120 @@
...
+import type { IndexWidget } from 'instantsearch.js/es/widgets/index/index';
+import type { ReactNode } from 'react';
+
+/**
+ * Returns the InstantSearch server state from a component.
+ */
+export async function getServerState(
+  children: ReactNode
+): Promise<InstantSearchServerState> {
+  const ssrClient: InstantSearchSSRClient = {
```
Haroenv, line 22: isn't this more like a random object than a search client? Maybe something like this makes more sense:

```
const instantSearchRef = { current: undefined }

// L26
instantSearchRef.current = instantSearchInstance
```

packages/react-instantsearch-hooks/src/useInstantSearch.ts
francoischalifour: The server UA is covered here: https://github.com/algolia/react-instantsearch/blob/f3b3533cd7b436fffa34f3e8748e5308d938a258/packages/react-instantsearch-hooks/src/useInstantSearch.ts#L102-L107

Did you mean there's another one? I can't find it in the codebase.

packages/react-instantsearch-hooks-server/src/getServerState.tsx
francoischalifour: This "client" naming doesn't refer to a search client, but rather to the API to communicate between server contexts. "client" may not be a great name though. At first I went for `ssrContext` but that's also confusing because of React Context.

Perhaps `ssrApi` is enough – it remains internal anyway.

Do you have other suggestions?

francoischalifour: Oh, I better understand your suggestion now. The thing is that it won't only be the InstantSearch ref soon, but also the search client cache.

packages/react-instantsearch-hooks/src/useInstantSearch.ts
Haroenv: I missed this part, thanks!

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: ssrApi might work as well, but as said it should be easy enough to access any properties we need from the InstantSearch instance, therefore a InstantSearchRef / Box, or even simply a non-boxed variable which is mutated would work for me

packages/react-instantsearch-hooks/src/__tests__/InstantSearchSSRContext.test.tsx
```diff
@@ -0,0 +1,7 @@
+import { InstantSearchSSRContext } from '../InstantSearchSSRContext';
+
+describe('InstantSearchSSRContext', () => {
+  test('exposes a displayName', () => {
```
sarahdayan, line 4: The `displayName` should be gone in production, you could test it by temporarily setting `__DEV__` to a falsy value.

packages/react-instantsearch-hooks/src/__tests__/InstantSearchSSRProvider.test.tsx
```diff
@@ -0,0 +1,275 @@
...
+</form>
+`);
+    });
+  });
+
+  test('with router provides initial UI state to InstantSearch', async () => {
+    const searchClient = createSearchClient();
+    const initialResults = {
+      indexName: {
+        state: { query: 'iphone' },
```
sarahdayan, line 153: IIUC this state should be computed by `getServerState` so it's unlikely that the `query` in `initialResults` be different from the location one (since the former would come from the latter), correct?

What happens if they differ?

packages/react-instantsearch-hooks/src/useIndex.ts
```diff
@@ -1,30 +1,60 @@
...
+  const serverContext = useInstantSearchServerContext();
+  const ssrContext = useInstantSearchSSRContext();
   const parentIndex = useIndexContext();
   const stableProps = useStableValue(props);
-  const indexWidget = useMemo(() => index(stableProps), [stableProps]);
+  // On SSR rendering, we add the Index early in a memo to render the initial
+  // results in the render pass. Since the Index is added already, we need to
+  // skip the usual browser effect that adds the widget when client-side
+  // rendering. We still need to run the effect lifecycle on all the other renders.
+  // This ref lets us keep track whether the initial effect to add the Index
```
sarahdayan, line 24:
```suggestion
  // This ref lets us track whether the initial effect to add the Index
```

packages/react-instantsearch-hooks/src/useConnector.ts
```diff
@@ -75,7 +86,7 @@ export function useConnector<
 
           return {
             ...scopedResult,
-            // We avoid all `results` being `null`.
+            // We avoid all `results` to be `null`.
```
sarahdayan, line 89:
```suggestion
            // We keep `results` from being `null`.
```

sarahdayan (commented):
Let's also document this feature on the main docs.

packages/react-instantsearch-hooks-server/README.md
```diff
@@ -0,0 +1,241 @@
...
+}
+```
+
+## API
+
+### `getServerState(<App />)`
+
+> `(children: React.ReactNode) => Promise<InstantSearchServerState>`
+
+Function that returns the server state.
```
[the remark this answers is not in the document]
francoischalifour: It doesn't necessarily take the full application, we actually want to advise to pass the minimal React tree that contains `<InstantSearchSSRProvider>` to optimize the double server rendering.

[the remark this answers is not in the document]
francoischalifour: Wdyt?

```suggestion
Function that takes the component that mounts `<InstantSearchSSRProvider>` and returns the server state.
```

packages/react-instantsearch-hooks-server/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "name": "react-instantsearch-hooks-server",
+  "version": "6.16.0",
+  "description": "⚡ Lightning-fast search for React, by Algolia",
+  "source": "src/index.ts",
+  "types": "dist/es/index.d.ts",
+  "main": "dist/cjs/index.js",
+  "module": "dist/es/index.js",
+  "sideEffects": false,
+  "license": "MIT",
+  "homepage": "https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/react/",
```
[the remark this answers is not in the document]
francoischalifour: I propose to do all these tasks in another PR for both the Hooks and Hooks Server package.

packages/react-instantsearch-hooks-server/rollup.config.js
```diff
@@ -0,0 +1,76 @@
...
+import replace from 'rollup-plugin-replace';
+import { uglify } from 'rollup-plugin-uglify';
+
+const clear = (x) => x.filter(Boolean);
+
+const version = process.env.VERSION || 'UNRELEASED';
+const algolia = '© Algolia, inc.';
+const link = 'https://github.com/algolia/react-instantsearch';
+const createBanner = (name) =>
+  `/*! React InstantSearch${name} ${version} | ${algolia} | ${link} */`;
```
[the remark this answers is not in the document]
francoischalifour: No sure really, I just used the same Rollup config as the other packages. I plan to refactor them in the future.

packages/react-instantsearch-hooks-server/rollup.config.js
```diff
@@ -0,0 +1,76 @@
...
+import filesize from 'rollup-plugin-filesize';
+import globals from 'rollup-plugin-node-globals';
+import resolve from 'rollup-plugin-node-resolve';
+import replace from 'rollup-plugin-replace';
+import { uglify } from 'rollup-plugin-uglify';
+
+const clear = (x) => x.filter(Boolean);
+
+const version = process.env.VERSION || 'UNRELEASED';
+const algolia = '© Algolia, inc.';
```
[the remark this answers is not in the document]
francoischalifour: I just used the same Rollup config as the other packages.

packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
```diff
@@ -0,0 +1,728 @@
...
+          facets: 'brand',
+          highlightPostTag: '__/ais-highlight__',
+          highlightPreTag: '__ais-highlight__',
+          hitsPerPage: 1,
+          maxValuesPerFacet: 10,
+          query: 'iphone',
+          page: 0,
+          tagFilters: '',
+        },
+      },
```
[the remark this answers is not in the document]
francoischalifour: It's because we mount a `<RefinementList>` in the root index, which requires a double query.

packages/react-instantsearch-hooks/src/InstantSearchServerContext.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import type { InstantSearch } from 'instantsearch.js';
+
+export type InstantSearchSSRClient = {
+  search?: InstantSearch;
+};
+
+export type InstantSearchServerApi = {
+  /**
+   * Notifies search internals to the server execution context to access them
```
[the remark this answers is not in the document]
francoischalifour: I meant that it passes them to the server execution context in `getServerState()`, so I believe the phrasing is correct?

packages/react-instantsearch-hooks/src/__tests__/InstantSearchSSRContext.test.tsx
francoischalifour: We cannot override Jest `globals` unfortunately.

packages/react-instantsearch-hooks/src/__tests__/InstantSearchSSRProvider.test.tsx
francoischalifour: Correct. In a real use case, `results[0].query` would have been "iphone" too. This test makes sure that the `<SearchBox>` render state is read by the provided helper state though, not from the result. So I think it makes sense to have it inconsistent here?

packages/react-instantsearch-hooks/src/__tests__/InstantSearchServerContext.test.tsx
```diff
@@ -0,0 +1,9 @@
+import { InstantSearchServerContext } from '../InstantSearchServerContext';
+
+describe('InstantSearchServerContext', () => {
+  test('exposes a displayName', () => {
+    expect(InstantSearchServerContext.displayName).toEqual(
+      'InstantSearchServer'
+    );
+  });
```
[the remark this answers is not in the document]
francoischalifour: We cannot override Jest `globals` unfortunately.

packages/react-instantsearch-hooks/src/useIndex.ts
francoischalifour: The original sentence is also correct, no? That's fewer words.

packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
```diff
@@ -0,0 +1,728 @@
...
+
+  test('returns initialResults', async () => {
+    const searchClient = createSearchClient();
+    const { App } = createTestEnvironment({ searchClient });
+
+    const serverState = await getServerState(<App />);
+
+    expect(Object.keys(serverState.initialResults)).toHaveLength(4);
+    expect(serverState.initialResults).toMatchInlineSnapshot(`
+Object {
```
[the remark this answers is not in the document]
francoischalifour: That's typically a good use of snapshot here. We want to assert any regressions on the initial results that are returned, and can't afford to just test some properties.

I put this test last so that the inline snapshot doesn't obstruct other tests. I'm OK with using an external snapshot too if you prefer.

packages/react-instantsearch-hooks-server/README.md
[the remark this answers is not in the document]
sarahdayan: That works 👍

packages/react-instantsearch-hooks-server/rollup.config.js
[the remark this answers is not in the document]
sarahdayan: I used AC for reference. We can probably tackle this later and have a single repos for our configs and utils.

packages/react-instantsearch-hooks/src/InstantSearchServerContext.ts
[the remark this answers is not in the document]
sarahdayan: Got it, but "notify to" isn't correct. In this case you can go with something like "forward" or "send".

packages/react-instantsearch-hooks/src/useIndex.ts
sarahdayan: My suggestion was missing a word, it should have been "keep track **of**".

I'm fine with your current sentence.

packages/react-instantsearch-hooks/src/__tests__/InstantSearchSSRProvider.test.tsx
sarahdayan: It does, yes.

By the helper state you mean l. 153 right? It's still unclear to me when I read the test from where "iphone" is picked up, here or `getLocation`.

packages/react-instantsearch-hooks/src/InstantSearchServerContext.ts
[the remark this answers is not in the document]
francoischalifour: Thanks. Updated in f574e24d56970e715d734399e02427704cb5737a.

packages/react-instantsearch-hooks/src/__tests__/InstantSearchSSRProvider.test.tsx
francoischalifour: The test name was confusing indeed. I renamed the two tests in fa7f06c327835ac67a1b2a808f96b2c0cccc7d55.

What these tests do is that they make sure that if you pass `initialUiState` or `routing`, the state coming from the initial results is the one used during render.

packages/react-instantsearch-hooks/package.json
francoischalifour: Updated in 5fdac8d6eda991bab5fa3e159efad62ac7a8ecd4.

packages/react-instantsearch-hooks-server/src/getServerState.tsx
francoischalifour: Addressed in ae76e8675eb98af797b9303649b158e130b80b57. I also removed the `InstantSearchSSRClient` type altogether.

Haroenv (approved):

sarahdayan (approved):

PR #3222, opened 2021-12-03
feat(hooks): introduce useCurrentRefinements

sarahdayan:
## Summary

This introduces the `useCurrentRefinements` hook.

For the example, I've used the [specs](https://github.com/algolia/instantsearch-specs) and the existing [`CurrentRefinements`](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-dom/src/components/CurrentRefinements.js) widget for inspiration. However I've used one list item per refinement, which looked clearer to me UI-wise.

Capture d’écran 2021-12-03 à 21 20 36

We can do a pass on responsiveness and refine the global UI at the end.

Documentation PR is open [here](https://github.com/algolia/doc/pull/6488).

### DIFF
--- a/examples/hooks/components/CurrentRefinements.tsx
+++ b/examples/hooks/components/CurrentRefinements.tsx
@@ -35,7 +35,7 @@ export function CurrentRefinements(props: CurrentRefinementsProps) {
         {refinements.map((refinement) => (
           <li key={refinement.label} className="ais-CurrentRefinements-item">
             <span className="ais-CurrentRefinements-label">
-              {getCategoryName(refinement)}:
+              {refinement.attribute}:
             </span>
             <span className="ais-CurrentRefinements-category">
               <span className="ais-CurrentRefinements-categoryLabel">
@@ -62,14 +62,4 @@ export function CurrentRefinements(props: CurrentRefinementsProps) {
       </ul>
     </div>
   );
-}
-
-function getCategoryName({
-  attribute,
-}: CurrentRefinementsConnectorParamsRefinement) {
-  if (attribute.startsWith('hierarchicalCategories')) {
-    return 'Hierarchy';
-  }
-
-  return attribute;
 }
--- a/examples/hooks/components/CurrentRefinements.tsx
+++ b/examples/hooks/components/CurrentRefinements.tsx
@@ -15,7 +15,7 @@ export function CurrentRefinements(props: CurrentRefinementsProps) {
   const { items, refine, canRefine } = useCurrentRefinements(props);
   const refinements = items.reduce(
     (acc, item) => [...acc, ...item.refinements],
-    []
+    [] as CurrentRefinementsConnectorParamsRefinement[]
   );
 
   return (
--- a/examples/hooks/components/CurrentRefinements.tsx
+++ b/examples/hooks/components/CurrentRefinements.tsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import {
+  useCurrentRefinements,
+  UseCurrentRefinementsProps,
+} from 'react-instantsearch-hooks';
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch-hooks/src/__tests__/useCurrentRefinements.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+      canRefine: false,
+      refine: expect.any(Function),
+      createURL: expect.any(Function),
+    });
+
+    await waitForNextUpdate();
+
+    // InstantSearch.js state from the `render` lifecycle step
+    expect(result.current).toEqual({
+      items: [],
```
Haroenv, line 28: it would be interesting to see a test *with* items (although we don't do anything special with them)

Haroenv (approved):
not sure about the divergence from spec, we however could make that change everywhere later

examples/hooks/components/CurrentRefinements.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const refinements = items.reduce(
+    (acc, item) => [...acc, ...item.refinements],
+    [] as CurrentRefinementsConnectorParamsRefinement[]
+  );
+
+  return (
+    <div
+      className={cx(
+        'ais-CurrentRefinements',
+        !canRefine && 'ais-CurrentRefinements-noRefinement',
```
francoischalifour, line 25:
```suggestion
        !canRefine && 'ais-CurrentRefinements--noRefinement',
```

francoischalifour (commented):
I think we should implement _exactly_ the UI spec otherwise it leaves room for disparities everywhere.

examples/hooks/components/CurrentRefinements.tsx
sarahdayan: Addressed in 4dc261d.

packages/react-instantsearch-hooks/src/__tests__/useCurrentRefinements.test.tsx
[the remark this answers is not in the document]
sarahdayan: Addressed in c8ebbc8.

sarahdayan: Adressed in c8ebbc8.

packages/react-instantsearch-hooks/src/__tests__/useCurrentRefinements.test.tsx
```diff
@@ -0,0 +1,96 @@
+import { renderHook } from '@testing-library/react-hooks';
+import React from 'react';
+
+import { RefinementList } from '../../../../examples/hooks/components';
```
francoischalifour, line 4: We shouldn't use a component from the examples in our lib tests. This is sufficient:

```tsx
function RefinementList(props: UseRefinementListProps) {
  useRefinementList(props);
  return null;
}
```

sarahdayan: Yep I saw that, I'm already on it ;)

sarahdayan:
@francoischalifour @Haroenv Makes sense to stick to the spec indeed for now. I've opened an umbrella ticket for all our accessibility issues in the backlog.

francoischalifour (approved):

packages/react-instantsearch-hooks/src/__tests__/useCurrentRefinements.test.tsx
```diff
@@ -0,0 +1,104 @@
+import { renderHook } from '@testing-library/react-hooks';
+import React from 'react';
+
+import { useRefinementList } from '..';
```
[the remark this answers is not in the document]
sarahdayan: Handled in 8c96e67a.

PR #3224, opened 2021-12-06
feat(hooks): introduce useInfiniteHits

dhayab:
**Summary**

This adds `useInfiniteHits` to our hooks collection.

**Result**

image

**Notes**

I took the liberty of adding a UI toggle to alternate between `Hits` and `InfiniteHits`, in order keep the example page from being too crowded. Let me know if that's ok or if there are any alternate solutions for that.

**Documentation**

https://github.com/algolia/doc/pull/6494


examples/hooks/components/layout/Tabs.css
```diff
@@ -0,0 +1,58 @@
...
+.ais-Tabs-header {
+  display: flex;
+}
+
+.ais-Tabs-title {
+  padding: 0 1rem;
+  flex: 1;
+  font-size: .875rem;
+  color: #23263b;
+  background: linear-gradient(-180deg,#fff,#fcfcfd);
```
francoischalifour, line 17:
```suggestion
  background: linear-gradient(-180deg, #fff, #fcfcfd);
```

examples/hooks/components/layout/Tabs.css
```diff
@@ -0,0 +1,58 @@
...
+.ais-Tabs-title:first-child {
+  border-radius: 3px 0 0 3px;
+}
+
+.ais-Tabs-title:last-child {
+  border-radius: 0 3px 3px 0;
+}
+
+.ais-Tabs-title:hover {
+  background: linear-gradient(-180deg,#fff,#f5f5fa);
```
francoischalifour, line 34:
```suggestion
  background: linear-gradient(-180deg, #fff, #f5f5fa);
```

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react';
+import { cx } from '../../cx';
+
+import './Tabs.css';
+
+export type TabProps = {
+  title: string;
+};
+
+export const Tabs = ({ children }) => {
```
francoischalifour, line 10: In the example we've been using functions.

```suggestion
export function Tabs({ children }) {
```

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+          >
+            {child}
+          </li>
+        ))}
+      </ul>
+    </div>
+  );
+};
+
+export const Tab: React.FC<TabProps> = ({ children }) => {
```
francoischalifour, line 49:
```suggestion
export function Tab({ children }: TabProps) {
```

examples/hooks/App.tsx
```diff
@@ -117,8 +119,15 @@ export function App() {
...
           </QueryRuleCustomData>
 
-          <Hits hitComponent={Hit} />
-          <Pagination className="Pagination" />
+          <Tabs>
+            <Tab title="Show Hits">
+              <Hits hitComponent={Hit} />
+              <Pagination className="Pagination" />
+            </Tab>
+            <Tab title="Show InfiniteHits">
```
francoischalifour, line 127: That's sufficient imo.

```suggestion
            <Tab title="Hits">
              <Hits hitComponent={Hit} />
              <Pagination className="Pagination" />
            </Tab>
            <Tab title="InfiniteHits">
```

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <div className="ais-Tabs">
+      <ul className="ais-Tabs-header">
+        {React.Children.map<React.ReactChild, React.ReactElement<TabProps>>(
+          children,
+          (child, index) => (
+            <li
+              className={cx(
+                'ais-Tabs-title',
+                currentTab === index && 'ais-Tabs-title--active'
```
Haroenv, line 22: not sure if it's worth it for this example, but this probably should be following the ARIA specs for tabs. I guess something like `aria-active` should be used if we want to show the good example.

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export type TabProps = {
+  title: string;
+};
+
+export const Tabs = ({ children }) => {
+  const [currentTab, setCurrentTab] = useState(0);
+
+  return (
+    <div className="ais-Tabs">
```
francoischalifour, line 14: The `<Tab>` component is not part of the InstantSearch.css spec so prefixing it with `ais` is confusing. The classname `.Tabs` should be fine. (Same for the inner components.)

Haroenv (approved):

examples/hooks/components/layout/Tabs.tsx
sarahdayan: I think it's worth it—some of our users might rely on it to use our demos, and it sets a good example.

[This markup](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html#sc1_label) should be fine.

examples/hooks/components/layout/Tabs.css
```diff
@@ -0,0 +1,58 @@
...
+
+.ais-Tabs-list {
+  margin-top: 1rem;
+}
+
+.ais-Tabs-item {
+  display: none;
+}
+
+.ais-Tabs-item.ais-Tabs-item--active {
```
sarahdayan, line 56:
```suggestion
.ais-Tabs-item--active {
```

examples/hooks/components/layout/Tabs.css
```diff
@@ -0,0 +1,58 @@
...
+.ais-Tabs-title:hover {
+  background: linear-gradient(-180deg,#fff,#f5f5fa);
+  z-index: 1;
+}
+
+.ais-Tabs-title + .ais-Tabs-title {
+  margin-left: -1px;
+}
+
+.ais-Tabs-title.ais-Tabs-title--active {
```
sarahdayan, line 42:
```suggestion
.ais-Tabs-title--active {
```

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+            className={cx(
+              'ais-Tabs-item',
+              currentTab === index && 'ais-Tabs-item--active'
+            )}
+            key={index}
+          >
+            {child}
+          </li>
+        ))}
+      </ul>
```
sarahdayan, line 44: See [here](https://github.com/algolia/react-instantsearch/pull/3224/files?authenticity_token=FQqTEgZ0AbCbLx1FVOcRzLs9DhcVP17M8uaC7vwEsR3EkIgq%2FW3hLvsHlBSrIvdN4dvHpmZB1%2BYsfSZ0QoPCxQ%3D%3D&file-filters%5B%5D=.css&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#r763389925) we should be fine with just `<div>`s and the right accessibility attributes.

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <div className="Tabs">
+      <div role="tablist" className="Tabs-header">
+        {React.Children.map<React.ReactChild, React.ReactElement<TabProps>>(
+          children,
+          (child, index) => (
+            <button
+              role="tab"
+              aria-selected={currentTab === index}
```
sarahdayan, line 43: You reuse this logic in multiple places (l. 46, l. 49), you could store it in a variable `isSelected`.

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+const getTabKey = (index: number, suffix?: string) =>
+  [`tab-${index}`, suffix].filter(Boolean).join('-');
+
+export function Tabs({ children }) {
+  const [currentTab, setCurrentTab] = useState(0);
+  const tabsRefs: HTMLElement[] = [];
+
+  const onKeyPress = (event: React.KeyboardEvent) => {
```
sarahdayan, line 18: Since this is called `onKeyDown` it'd make more sense.

```suggestion
  const onKeyDown = (event: React.KeyboardEvent) => {
```

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+              id={getTabKey(index, 'title')}
+              tabIndex={currentTab !== index ? -1 : 0}
+              className={cx(
+                'Tabs-title',
+                currentTab === index && 'Tabs-title--active'
+              )}
+              ref={(element) => (tabsRefs[index] = element!)}
+              key={getTabKey(index)}
+              onClick={() => setCurrentTab(index)}
+              onKeyDown={(event) => onKeyPress(event)}
```
sarahdayan, line 54:
```suggestion
              onKeyDown={onKeyDown}
```

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+        break;
+      case 'ArrowRight':
+        tabIndex = Math.min(currentTab + 1, React.Children.count(children) - 1);
+        break;
+      default:
+        return;
+    }
+
+    tabsRefs[tabIndex].focus();
+    setCurrentTab(tabIndex);
```
sarahdayan, line 32: I'd use an effect to set the focus when the tab changes instead.

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+  children: React.ReactNode;
+  title: string;
+};
+
+const getTabKey = (index: number, suffix?: string) =>
+  [`tab-${index}`, suffix].filter(Boolean).join('-');
+
+export function Tabs({ children }) {
+  const [currentTab, setCurrentTab] = useState(0);
+  const tabsRefs: HTMLElement[] = [];
```
sarahdayan, line 16: Since this is a plain array, doesn't it get reconstructed at each render?

Maybe you could have something like this instead:

```js
const tabsRef = useRef(React.Children.map((child) => null));
```

Then assign each ref per tab by index?

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+            {child}
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+}
+
+export function Tab({ children }: TabProps) {
+  return <div>{children}</div>;
```
sarahdayan, line 80: Is the wrapping `div` necessary?

dhayab: TypeScript raises an issue if I directly return the children (ReactNode) as its expecting a JSX.Element. I could probably wrap it in a Fragment to prevent adding another level of depth in the DOM though. What do you think?

examples/hooks/components/layout/Tabs.tsx
Haroenv: you can't focus outside of an event handler, no?

examples/hooks/components/layout/Tabs.tsx
Haroenv: Fragment seems better

examples/hooks/components/layout/Tabs.tsx
sarahdayan: Can't you? https://codesandbox.io/s/misty-frost-ekigj?file=/src/App.js

dhayab: It does work on my side as well.

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,82 @@
...
+        break;
+      case 'ArrowRight':
+        tabIndex = Math.min(currentTab + 1, React.Children.count(children) - 1);
+        break;
+      default:
+        return;
+    }
+
+    setCurrentTab(tabIndex);
+  };
```
francoischalifour, line 37: That's quite some code and it couples any "keydown" event with `setCurrentTab`.

How about inlining everything?

```suggestion
  const onKeyDown = (event: React.KeyboardEvent) => {
    if (event.key === 'ArrowLeft') {
      setCurrentTab(Math.max(0, currentTab - 1));
    } else if (event.key === 'ArrowRight') {
      setCurrentTab(
        Math.min(currentTab + 1, React.Children.count(children) - 1)
      );
    }
  };
```

examples/hooks/components/layout/Tabs.tsx
```diff
@@ -0,0 +1,85 @@
...
+            const isSelected = currentTab === index;
+            return (
+              <button
+                role="tab"
+                aria-selected={isSelected}
+                aria-controls={getTabKey(index, 'item')}
+                id={getTabKey(index, 'title')}
+                tabIndex={isSelected ? 0 : -1}
+                className={cx('Tabs-title', isSelected && 'Tabs-title--active')}
+                ref={(element) => (tabsRefs.current[index] = element!)}
```
sarahdayan, line 54: Not sure you need a callback ref. If `tabsRef` was already an array of the right amount of `null` values, you could do `ref={tabsRef[index]}`.

dhayab: I was not able to make it work with this syntax. I tried to simplify the use case but no luck: https://codesandbox.io/s/prod-glitter-bvenf?file=/src/Wrapper.js.

francoischalifour (approved):

examples/hooks/components/layout/Tabs.tsx
sarahdayan: My initial suggested code was wrong. Here's a working sandbox: https://codesandbox.io/s/laughing-snowflake-vm92d

dhayab: I tried it and apparently all the refs in the array are set to the last child.

image

I looked into it but could not figure out why 😕.

sarahdayan: No it's normal, my sandbox actually doesn't work since hooks can't be used in loops. I guess having a single ref array and assigning it with a callback ref is the only way 🙂

sarahdayan (approved):

examples/hooks/components/InfiniteHits.tsx
```diff
@@ -0,0 +1,56 @@
...
+          onClick={showPrevious}
+          disabled={isFirstPage}
+        >
+          Show previous results
+        </button>
+      )}
+      <ol className="ais-InfiniteHits-list">
+        {hits.map((hit) => (
+          <li key={hit.objectID} className="ais-InfiniteHits-item">
+            <Hit hit={hit as unknown as THit} />
```
tkrugg, line 40: we could avoid having this cast if useInfiniteHit<THit> passed down the generic to the hits render prop.
The connector is not doing that right now unfortunately. @Haroenv

examples/hooks/components/InfiniteHits.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { cx } from '../cx';
+
+export type InfiniteHitsProps<THit> = React.ComponentProps<'div'> &
+  UseInfiniteHitsProps & {
+    hitComponent: (props: { hit: THit }) => JSX.Element;
+  };
+
+export function InfiniteHits<THit extends AlgoliaHit<Record<string, unknown>>>({
+  hitComponent: Hit,
```
tkrugg, line 17: I would keep the name `hitComponent` or just rename it to `Hit` component, otherwise, at least for me, it's too confusing to remember if you're dealing with the type, or the prop.

examples/hooks/components/layout/Tabs.css
```diff
@@ -0,0 +1,43 @@
...
+
+.Tabs-title--active {
+  font-weight: bold;
+  box-shadow: inset 0 1px 4px 0 rgb(119 122 175 / 40%), inset 0 1px 1px 0 rgb(119 122 175 / 40%), 0 1px 0 0 rgb(35 38 59 / 5%);
+  z-index: 1;
+}
+
+.Tabs-list {
+  margin-top: 1rem;
+}
```
tkrugg, line 43: Great job on the tabs! 
I would make an extra effort and have the example follow the [satellite tabs UI](https://algolia-satellite.netlify.app/?path=/docs/navigation-tabs--basic).
I would either replicate the styles or use their tabs component. 

https://algolia-satellite.netlify.app/?path=/docs/navigation-tabs--basic

We usually don't put a lot of effort in the look of our demos, let's try to change that.

tkrugg (changes requested):
I left a comment on a possible readability improvement, not blocking. 

The tabs in the example were a great idea. Requesting changes on the styling of the tabs: let's try to better follow the satellite UI. I'll leave up to you on whether you'd like to replace the styles or reuse their components.

examples/hooks/components/layout/Tabs.css
dhayab: Didn't know we had that. I'll look into it and it might be better to use their tabs component indeed. Thanks!

Haroenv: I think we have gone back-and-forth too long about these tabs already, and what's used now fits the design as well, and is also accessible. We can swap it out for either the satellite or another third party later. Don't forget this is just an example!

examples/hooks/components/InfiniteHits.tsx
Haroenv: not sure how to do that properly, but we can introduce the generic in InstantSearch consistently afterwards, there's the same issue for hits and for autocomplete (more complicated as it can have multiple hits)

examples/hooks/components/InfiniteHits.tsx
sarahdayan: You meant `HitComponent` right?

examples/hooks/components/layout/Tabs.css
dhayab: Okay, I took a few minutes to try it out and their CSS overrides some of our own styles, so there might be some more work involved. I'll merge it for now if that's okay.

PR #3227, opened 2021-12-08
chore(release): remove examples update after release

francoischalifour:
This removes the examples update after release in an attempt at making the release script work. It's been broken for months and messes up our tags and releases.

I want to release the next version without this step to make sure that this is the culprit. Then, we should be able to rely on the monorepo feature instead of manually updating the examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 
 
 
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
makes sense for now

PR #3232, opened 2021-12-08
feat(hooks): bundle as es-module

Haroenv:
**Summary**

- [x] set type: module
- [x] set exports for transparent cjs usage (possible here easily as there's only a single entry file)
- [x] set up babel for adding .js extension
- [x] replace /es import by /cjs in the cjs build
- [x] make sure tests still run as expected
- [x] update instantsearch.js with ES-compatible version 

**Result**

FX-534

The package should now be consumable both by any legacy methods, but also modern setups like Next 12


packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -14,7 +14,7 @@ import type {
 /**
  * Returns the InstantSearch server state from a component.
  */
-export async function getServerState(
```
Haroenv, line 17: this should be a separate PR

scripts/jest/setupTests.ts
```diff
@@ -1,9 +1,13 @@
+// lint rule disabled, as due to disabling no-unresolved and having .js extensions
+// eslint can no longer know that warnCache actually is exported (and ts can't)
+// help here either.
+// eslint-disable-next-line import/named
```
Haroenv, line 4:
```suggestion
```

test/utils/createInstantSearchTestWrapper.tsx
```diff
@@ -1,5 +1,8 @@
 import React from 'react';
 import type { InstantSearchProps } from '../../packages/react-instantsearch-hooks/src';
+// this exists, but eslint doesn't know due to not using the typescript resolver
+// that's done to ensure .js extensions are allowed to import .ts files
+// eslint-disable-next-line import/named
```
Haroenv, line 5:
```suggestion
```

package.json
```diff
@@ -93,6 +93,7 @@
     "eslint": "6.8.0",
     "eslint-config-algolia": "16.0.0",
     "eslint-config-prettier": "6.15.0",
+    "eslint-import-resolver-typescript": "^2.5.0",
```
francoischalifour, line 96: We can pin this dev dep.

Haroenv:
While this PR stands on its own, to be fully an ES module, it requires https://github.com/algolia/instantsearch.js/pull/4971 to be used as well. That could be done separately, but should be in the same release as this PR

packages/react-instantsearch-hooks-server/package.json
```diff
@@ -30,7 +37,8 @@
     "clean": "rm -rf dist",
     "watch": "yarn build:cjs --watch",
     "build": "yarn build:cjs && yarn build:es && yarn build:umd && yarn build:types",
-    "build:cjs": "babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet",
+    "build:cjs": "babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet && yarn build:cjs-replace-imports",
+    "build:cjs-replace-imports": "find dist/cjs -type f -exec sed -i.bak 's|instantsearch.js/es|instantsearch.js/cjs|g' {} + && find dist/cjs -name '*.bak' -type f -exec rm {} +",
```
francoischalifour, line 41: We could move this to our `script/`. It would be more self-documenting and centralized somewhere when edits are needed.

scripts/babel/extension-resolver.js
```diff
@@ -0,0 +1,142 @@
...
+// and can't be resolved like is possible with CommonJS. To have large compatibility
+// without much hassle, we don't use extensions *inside* the source code, but add
+// them with this plugin.
+
+const fs = require('fs');
+const path = require('path');
+
+const PLUGIN_NAME = 'babel-plugin-extension-resolver';
+
+const importStartsWith = ['instantsearch.js/', 'react-dom/'];
```
francoischalifour, line 12: Are those exceptions, or could we generalize this behavior?

francoischalifour (commented):
Would it be possible to write a test that makes sure that the final build is ESM-spec compliant?

Haroenv:
> Would it be possible to write a test that makes sure that the final build is ESM-spec compliant?

Good idea, although I'm not really sure what that could be. Will experiment

scripts/babel/extension-resolver.js
Haroenv: I'm not actually 100% sure why react-dom/server also is required even though it's CommonJS, but I had an error in Next if that wasn't the case. There's other dependency (dequal/lite) that errored if I *did* add the extension. I'll also investigate what it is that makes the module required to be fully resolved for react-dom

Haroenv: These are the two dependencies that are required according to node. I think it's because dequal explicitly mentions it's cjs, but react doesn't

Haroenv:
test fails now as InstantSearch.js hasn't been updated yet

packages/react-instantsearch-hooks-server/test/is-a-module.mjs
```diff
@@ -0,0 +1,5 @@
+import assert from 'assert';
+
+import * as ReactInstantSearchHooksServer from 'react-instantsearch-hooks-server';
+
+assert.ok(ReactInstantSearchHooksServer);
```
francoischalifour, line 5: So that we know why it fails (if it does):

```suggestion
assert.ok(ReactInstantSearchHooksServer, 'The module `react-instantsearch-hooks-server` is not ESM compliant.');
```

packages/react-instantsearch-hooks/package.json
```diff
@@ -33,10 +40,12 @@
     "clean": "rm -rf dist",
     "watch": "yarn build:cjs --watch",
     "build": "yarn build:cjs && yarn build:es && yarn build:umd && yarn build:types",
-    "build:cjs": "babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet",
+    "build:cjs": "babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet && yarn build:cjs-replace-imports",
+    "build:cjs-replace-imports": "../../scripts/replace-import-for-cjs.sh",
```
francoischalifour, line 44: At this point we can inline everything since this script is just used for the CJS build?

```suggestion
    "build:cjs": "babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet && ../../scripts/replace-import-for-cjs.sh",
```

packages/react-instantsearch-hooks/src/useInstantSearch.ts
```diff
@@ -1,4 +1,4 @@
-import instantsearch from 'instantsearch.js';
+import instantsearch from 'instantsearch.js/es';
```
francoischalifour, line 1: Just making sure, removing `/es` would fail the `node test/is-a-module.mjs` test with the newer IS.js dependency?

scripts/babel/extension-resolver.js
francoischalifour: That's possible. Perhaps we can add a comment explaining why these 2 deps need to be treated as such?

Then as long as the module test fails without an item in this array, that's good to me.

scripts/replace-import-for-cjs.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e # exit on error
+
+# replace all imports of instantsearch.js with /cjs
+find dist/cjs -type f -exec sed -i.bak 's|instantsearch.js/es|instantsearch.js/cjs|g' {} +
```
francoischalifour, line 6: What's this for? I thought we changed imports from `from 'instantsearch.js'` to `from 'instantsearch.js/es'`.

( although `from 'instantsearch.js'` is still used for type imports)

francoischalifour (commented):
That looks good!

packages/react-instantsearch-hooks-server/test/is-a-module.mjs
Haroenv: yes, can change, although if it's not esm compliant it will throw an error before this assert, this assert is to make sure it's not empty

packages/react-instantsearch-hooks/src/useInstantSearch.ts
Haroenv: will check, but yes

scripts/replace-import-for-cjs.sh
Haroenv: cjs can not import esm, so inside the cjs build all /es needs to be changed to /cjs

packages/react-instantsearch-hooks-server/package.json
```diff
@@ -30,10 +37,12 @@
...
     "watch": "yarn build:cjs --watch",
     "build": "yarn build:cjs && yarn build:es && yarn build:umd && yarn build:types",
-    "build:cjs": "babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet",
+    "build:cjs": "babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet && yarn build:cjs-replace-imports",
+    "build:cjs-replace-imports": "../../scripts/replace-import-for-cjs.sh",
     "build:es": "BABEL_ENV=es babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/es --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet",
     "build:umd": "BABEL_ENV=rollup rollup -c rollup.config.js",
-    "build:types": "tsc -p ./tsconfig.declaration.json --outDir ./dist/es"
+    "build:types": "tsc -p ./tsconfig.declaration.json --outDir ./dist/es",
+    "test:build": "node test/is-a-module.mjs"
```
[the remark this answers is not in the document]
francoischalifour: We're testing that our tooling maps the imports correctly, not that the library code works as expected. For this reason I think it makes more sense to move all the tooling test top-level, and have focused library code in the packages.

We could have the two tests in `/test/packages-are-modules.mjs`.

packages/react-instantsearch-hooks-server/test/is-a-module.mjs
francoischalifour: Oh I see. Ignore my comment then.

scripts/replace-import-for-cjs.sh
francoischalifour: I understand now, I had misread the script.

dhayab (approved):
Looks good to me. I ran the test without issues after linking the relevant branch of instantsearch.js.

scripts/babel/extension-resolver.js
```diff
@@ -0,0 +1,145 @@
...
+
+const fs = require('fs');
+const path = require('path');
+
+const PLUGIN_NAME = 'babel-plugin-extension-resolver';
+
+// InstantSearch.js/es is an ES Module, so needs complete paths,
+// React-DOM also fails if the paths are incomplete
+// For verification, see test/module/packages-are-es-modules.mjs
+const importStartsWith = ['instantsearch.js/', 'react-dom/'];
```
francoischalifour, line 15: Actually, can we pass these imports to the `extensionResolver({ imports: ['instantsearch.js', 'react-dom'] })` call in the Babel config? (and add the `/` in this plugin)

It makes more sense to decouple this for the plugin itself to remain generic.

francoischalifour (approved):
I proposed a last update in the Babel plugin.

Thank you, that looks great!

scripts/babel/extension-resolver.js
Haroenv: yes, could do, although the slash is specifically because we only want non-global imports to have extension added. I guess that can be generalised indeed

tkrugg (approved):

PR #3235, opened 2021-12-08
Save on hits triggering rerenders

brunoAltinet:
**Summary**
I'm having an issue with hits being always new in connected component and triggering useEffect excessively. This is a moderate tweak which could be done better (with useMemo but i'm not sure how to use it in this instance), but even this should help with majority of usecases

**Result**
The connected component would get rendered a lot with always different array of hits, which throws useEffect out of whack. Hits should be memoized.

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectGeoSearch.js
+++ b/packages/react-instantsearch-core/src/connectors/connectGeoSearch.js
@@ -173,9 +173,10 @@ export default createConnector({
 
     const position =
       currentPositionFromSearchState || currentPositionFromSearchParameters;
+    const theHits = !results ? [] : results.hits;
 
     return {
-      hits: !results ? [] : results.hits.filter((_) => Boolean(_._geoloc)),
+      hits: theHits.every(x => Boolean(x._geoloc)) ? theHits : theHits.filter((_) => Boolean(_._geoloc)), //save on rerenders and effects further down the line
       isRefinedWithMap: Boolean(currentRefinement),
       currentRefinement,
       position,


Haroenv:
If I understand this PR correctly, you want to have the same reference of `hits` if they didn't change?

brunoAltinet:
I want to change the bounding box /viewport change if hits change. Right now, every rerender gets my viewport reset to the bounding box, eg. openeing pin popup
```
  useEffect(() => {
    trackViewed(embedPrefix, location, radius, props.hits);
    const vp = new WebMercatorViewport({ latitude: 0, longitude: 0, width: staticSize[0], height: staticSize[1] });
    const boundbox = bbox(props.hits.map((x) => x._geoloc));
    if (!boundbox) return;
    const { minLng, minLat, maxLat, maxLng } = boundbox;
    let { longitude, latitude, zoom } = vp.fitBounds(
      [
        [minLng, minLat],
        [maxLng, maxLat],
      ],
      { padding: 60 }
    );
    zoom = zoom > 14 ? 14 : zoom;
    setViewport({
      latitude: latitude,
      longitude: longitude,
      zoom: zoom,
      bearing: 0,
      pitch: 0,
    });
  }, [props.hits]); 
```

Haroenv:
I guess the problem is that you assume the viewport needs to be reset on every change of the hits. In our own component this is implemented as follows: https://github.com/algolia/react-instantsearch/blob/d459e62f5cae4c98427ab302531873f5ee23d149/packages/react-instantsearch-dom-maps/src/GoogleMaps.js#L55-L89

brunoAltinet:
Ok, I still feel like we should get same reference for same results, although i did check and it seems that invidiual hits are also newed-up so it doesn't really hold up even if you return the same array (though i'm not sure why that is).

grahammcculloch:
I've actually seen a similar problem to what @brunoAltinet is describing, but with the `items` prop passed from `connectRefinementList` and `connectHierarchicalMenu`. If I have something like:
```javascript
useEffect(() => {
  console.log('Items changed', items)
}, [items])
```
I will get (typically) three log statements with identical value items each time I make a new search request (e.g. by selecting a facet). I would expect the `items` object to be referentially the same unless the values change.

PR #3237, opened 2021-12-10
feat(hooks): introduce `useNumericMenu`

francoischalifour:
This adds `useNumericMenu` to our Hooks collection.

## API

This hook is a bridge to `connectNumericMenu`.

## Usage

```jsx
function NumericMenu(props) {
  const { items, refine } = useNumericMenu(props);

  return /* ... */;
}

function App(props) {
  return (
    <InstantSearch {...props}>
      <NumericMenu
        attribute="price"
        items={[
          { label: 'All' },
          { label: 'Less than 500$', end: 500 },
          { label: 'Between 500$ - 1000$', start: 500, end: 1000 },
          { label: 'More than 1000$', start: 1000 },
        ]}
      />

      {/* ... */}
    </InstantSearch>
  );
}
```

### DIFF
--- a/examples/hooks/App.tsx
+++ b/examples/hooks/App.tsx
@@ -5,21 +5,22 @@ import { InstantSearch, DynamicWidgets } from 'react-instantsearch-hooks';
 
 import {
   Configure,
+  CurrentRefinements,
   HierarchicalMenu,
   Highlight,
   Hits,
+  HitsPerPage,
   InfiniteHits,
+  Menu,
+  NumericMenu,
   Pagination,
   Panel,
-  RangeInput,
-  RefinementList,
-  Menu,
-  SearchBox,
-  SortBy,
-  HitsPerPage,
   QueryRuleContext,
   QueryRuleCustomData,
-  CurrentRefinements,
+  RangeInput,
+  RefinementList,
+  SearchBox,
+  SortBy,
 } from './components';
 import { Tab, Tabs } from './components/layout';
 
@@ -53,7 +54,7 @@ export function App() {
       indexName="instant_search"
       routing={true}
     >
-      <Configure hitsPerPage={15} />
+      <Configure ruleContexts={[]} />
 
       <div className="Container">
         <div>
@@ -83,6 +84,17 @@ export function App() {
               <RangeInput attribute="price" />
             </Panel>
           </DynamicWidgets>
+          <Panel header="NumericMenu">
+            <NumericMenu
+              attribute="price"
+              items={[
+                { label: 'All' },
+                { label: 'Less than 500$', end: 500 },
+                { label: 'Between 500$ - 1000$', start: 500, end: 1000 },
+                { label: 'More than 1000$', start: 1000 },
+              ]}
+            />
+          </Panel>
         </div>
         <div className="Search">
           <div className="Search-header">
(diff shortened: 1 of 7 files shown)


examples/hooks/App.tsx
```diff
@@ -81,6 +82,17 @@ export function App() {
               <RangeInput attribute="price" />
             </Panel>
           </DynamicWidgets>
+          <Panel header="NumericMenu">
```
dhayab, line 85: Could this panel also be handled by `<DynamicWidgets>` for consistency?

Also, I see some panel names refering to the type of widgets, and other refering to the faceting attribute, maybe we could choose one or the other but not mix it, wdyt?

examples/hooks/App.tsx
```diff
@@ -81,6 +82,17 @@ export function App() {
...
+            <NumericMenu
+              attribute="price"
+              items={[
+                { label: 'All' },
+                { label: 'Less than 500$', end: 500 },
```
dhayab, line 90: This is nitpicking, but I'd place the $ currency before the value.

dhayab (approved):
Looks good to me, I only have comments unrelated to the feature on my side.

examples/hooks/App.tsx
francoischalifour: You're right. I copied the example from the docs but your suggestion is more correct.

examples/hooks/App.tsx
francoischalifour: The `<NumericMenu>` doesn't display when within `<DynamicWidgets>`. I believe that's because of the Algolia index configuration.

Can you confirm @Haroenv?

francoischalifour: The inconsistency started with the "Hierarchy" panel because the `categories` and `hierarchicalCategories` are essentially the same. We should find another name though.

I'll rename this panel to "Price range" if that's good for you.

examples/hooks/components/NumericMenu.tsx
```diff
@@ -0,0 +1,43 @@
...
+  UseNumericMenuProps;
+
+export function NumericMenu(props: NumericMenuProps) {
+  const { hasNoResults, items, refine } = useNumericMenu(props);
+
+  return (
+    <div
+      className={cx(
+        'ais-NumericMenu',
+        hasNoResults && 'ais-NumericMenu--noRefinement',
```
Haroenv, line 16: should we use canRefine here?

Haroenv (commented):
code lgtm

examples/hooks/components/NumericMenu.tsx
francoischalifour: It's not returned by the connector unfortunately. Let's maybe add that in IS.js in a minor, and remove `hasNoResults` in the next major?

examples/hooks/App.tsx
francoischalifour: See db6e4e1a8e93781e8d33b9e2a32c0b6326417344.

examples/hooks/App.tsx
francoischalifour: See db6e4e1a8e93781e8d33b9e2a32c0b6326417344.

examples/hooks/components/NumericMenu.tsx
Haroenv: sounds good, I didn't know this widget was missed

Haroenv (approved):

PR #3239, opened 2021-12-10
docs(examples): add a Next.js hooks example

Haroenv:
**Summary**

adds a Next.js example with React InstantSearch hook as it's fully compatible with ES Modules

**Result**

Simpler example as the "hooks" one copied (maybe we should ship the components and just say they're experimental? Different discussion anyway)


examples/hooks-next/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "dependencies": {
+    "algoliasearch": "^4.11.0",
+    "instantsearch.css": "^7.4.5",
+    "next": "12.0.7",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
+    "react-instantsearch-hooks": "^6.17.0",
+    "react-instantsearch-hooks-server": "^6.17.0"
+  },
+  "resolutions": {
```
dhayab, line 19: I haven't been able to run the example locally or in CSB with the resolutions of those packages, but the code looks good to me. I'll retry when those packages will be released.

examples/hooks-next/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "rules": {
```
dhayab, line 2: Building and linting fail if ESLint doesn't load the Next.js plugin.

```suggestion
  "extends": ["next"],
  "rules": {
```

dhayab (changes requested):

examples/hooks-next/.eslintrc
Haroenv: it's just a warning right? otherwise the example test doesn't work

dhayab: On top of the warning, it fails when it's run outside of the react-instantsearch repository structure:

```sh
curl https://codeload.github.com/algolia/react-instantsearch/tar.gz/feat/next-hooks-example | tar -xz --strip=2 react-instantsearch-feat-next-hooks-example/examples/hooks-next
```

image

I'm not sure why, maybe ESLint tries to find missing configuration in parent directories and since we have one at the root of **react-instantsearch** its okay.

Haroenv: I think it's fixed now

dhayab (approved):
I retested in a clean docker environment and still can't build the project when isolated from the main repository, although the dev server works correctly and now ESLint accepts the `react-hooks/exhaustive-deps` as well.

Also, the sandbox works (https://codesandbox.io/s/github/algolia/react-instantsearch/tree/feat/next-hooks-example/examples/hooks-next).

Since it's an example, maybe this remaining issue should not block the PR.

francoischalifour (approved):

PR #3242, opened 2021-12-15
feat(dynamicWidgets): send facets * and maxValuesPerFacet by default

Haroenv:
FX-622

**Summary**

The default additional search parameters of `{ facets: ['*'], maxValuesPerFacet: 20 }` prevents an extra network request when widgets mount/unmount, as both queries will be identical, and thus cached.

If a refinement is set, this still results in two network requests, as the refinements can't be applied before it's known if the widget should be mounted.

**Result**

When dynamic widgets is used without an explicit configure widget with `facets: ['*']` and `maxValuesPerFacet: highestLimit` an extra network request happens on load, which has a negative impact on performance. With the new additional parameters default value, widgets mounting or unmounting no longer require a separate network request

Two new parameters are added to `connectDynamicWidgets` and `dynamicWidgets`:

```js
<DynamicWidgets
  // ... other parameters
  // overridden with the default value
  facets={['*']}
  maxValuesPerFacet={20}
/>
```

The default behaviour can also be avoided, and the previous default (non-matching parameters, which does two network requests) can be done like this:

```js
<DynamicWidgets
  // ... other parameters
  // overridden default to ensure two requests are sent,
  // and that not all facets are requested
  facets={[]}
/>
```

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
+++ b/packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
@@ -1,16 +1,22 @@
+import type { SearchParameters } from 'algoliasearch-helper';
 import PropTypes from 'prop-types';
 import createConnector from '../core/createConnector';
 import { getResults } from '../core/indexUtils';
+
+const MAX_WILDCARD_FACETS = 20;
 
 export default createConnector({
   displayName: 'AlgoliaDynamicWidgets',
 
   defaultProps: {
     transformItems: (items) => items,
+    maxValuesPerFacet: 20,
   },
 
   propTypes: {
     transformItems: PropTypes.func,
+    facets: PropTypes.arrayOf(PropTypes.string),
+    maxValuesPerFacet: PropTypes.number,
   },
 
   getProvidedProps(props, _searchState, searchResults) {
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch-core/src/connectors/connectDynamicWidgets.ts
```diff
@@ -30,8 +51,36 @@ export default createConnector({
...
+      // eslint-disable-next-line no-console
+      console.warn(
+        `More than ${MAX_WILDCARD_FACETS} facets are requested to be displayed without explicitly setting which facets to retrieve. This could have a performance impact. Set "facets" to [] to do two smaller network requests, or explicitly to ['*'] to avoid this warning.`
+      );
+    }
+
+    if (props.maxValuesPerFacet < results._state.maxValuesPerFacet) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `The maxValuesPerFacet set by dynamic widgets (${props.maxValuesPerFacet}) is smaller than one of the limits set by a widget (${results._state.maxValuesPerFacet}). This causes a mismatch in query parameters and thus an extra network request when that widget is mounted.`
```
tkrugg, line 66: [same as for IS.js,](https://github.com/algolia/instantsearch.js/pull/4968/files) I think this message would be more helpful it mentioned the name of the widget or the attribute, and clearly said: your should set your maxValuesPerFacet to max(limit/showMoreLimit)

Haroenv: I don't think we can know that, as the maxValuesPerFacet isn't related to a single widget, but to all of them

PR #3245, opened 2022-01-04
feat(hooks): introduce `useBreadcrumb()`

FabienMotte:
## Summary

This adds `useBreadcrumb()` to our Hooks collection.

[FX-452](https://algolia.atlassian.net/browse/FX-452)

## Deps

Waiting for this PR to be merged and released in InstantSearch.js: https://github.com/algolia/instantsearch.js/pull/4980
Unit tests are failing because of that for now.

## API

This hook is a bridge to `connectBreadcrumb`.

## Usage

```jsx
function Breacrumb(props) {
  const { items, refine, createURL } = useBreadcrumb(props);

  return /* ... */;
}

function App(props) {
  return (
    <InstantSearch {...props}>
      <Breadcrumb
        attributes={[
          'hierarchicalCategories.lvl2',
        ]}
      />

      {/* ... */}
    </InstantSearch>
  );
}
```


examples/hooks/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,73 @@
...
+export type BreadcrumbProps = React.ComponentProps<'div'> & UseBreadcrumbProps;
+
+export function Breadcrumb(props: BreadcrumbProps) {
+  const { items, refine, createURL } = useBreadcrumb(props);
+
+  return (
+    <div
+      className={cx(
+        'ais-Breadcrumb',
+        items.length === 0 && 'ais-Breadcrumb--noRefinement',
```
francoischalifour, line 15: We can rely on [`canRefine`](https://github.com/algolia/instantsearch.js/blob/8b0af3e44f3cbb891623cab5cfb32e4cf9046f93/src/connectors/breadcrumb/connectBreadcrumb.ts#L218) for this check. (Same applies in other places.)

examples/hooks/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,73 @@
...
+    >
+      <ul className="ais-Breadcrumb-list">
+        <li
+          className={cx(
+            'ais-Breadcrumb-item',
+            items.length === 0 && 'ais-Breadcrumb-item--selected'
+          )}
+        >
+          <a
+            href={createURL(null)}
```
francoischalifour, line 27: We've been using `undefined` to reflect that the refined value must be unset.

```suggestion
            href={createURL(undefined)}
```

examples/hooks/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,73 @@
...
+            onClick={(event) => {
+              event.preventDefault();
+              refine(null);
```
francoischalifour, line 30: We've been using `undefined` to reflect that the refined value must be unset.

```suggestion
              refine(undefined);
```

test/utils/createInstantSearchTestWrapper.tsx
```diff
@@ -5,11 +5,17 @@ import { InstantSearch } from '../../packages/react-instantsearch-hooks/src';
...
+  }
 ) {
-  const searchClient = createSearchClient();
   const wrapper = ({ children }) => (
-    <InstantSearch searchClient={searchClient} indexName="indexName" {...props}>
+    <InstantSearch
+      searchClient={options.searchClient}
+      indexName="indexName"
+      {...props}
+    >
```
Haroenv, line 18: I think this change isn't needed, if it's back to how the options are previously, provided searchClient overrides the `const searchClient` too, right?

examples/hooks/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,73 @@
...
+            Home
+          </a>
+        </li>
+
+        {items.map((item, idx) => {
+          const isLast = idx === items.length - 1;
+
+          return (
+            <li
+              key={item.label + idx}
```
Haroenv, line 43: the index is probably enough to distinguish between the levels right?

Haroenv (changes requested):
implementation looks great! couple small nits that can be simplified still

test/utils/createInstantSearchTestWrapper.tsx
FabienMotte: Good catch, thank you!

examples/hooks/components/Breadcrumb.tsx
FabienMotte: Yes indeed, thanks!

examples/hooks/components/Breadcrumb.tsx
Haroenv: why did you resolve @FabienMotte? it seems like you still use `null`?

FabienMotte: @Haroenv Yes indeed, I discussed about it with @francoischalifour and we decided to keep it as it is, to use `undefined`, we should first update the typing here: https://github.com/algolia/instantsearch.js/blob/master/src/connectors/breadcrumb/connectBreadcrumb.ts#L30
WDYT?

francoischalifour: I think we can use `undefined` with a `ts-expect-error` and fix the typing in IS.js.

Haroenv: hmm, or we use `null` in the InstantSearch.js component too? I guess it doesn't matter too much for now as it just checks if the value is falsy https://github.com/algolia/instantsearch.js/blob/master/src/connectors/breadcrumb/connectBreadcrumb.ts#L131

FabienMotte: I'll go with @francoischalifour proposal, thanks!

examples/hooks/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,73 @@
...
+          className={cx(
+            'ais-Breadcrumb-item',
+            !canRefine && 'ais-Breadcrumb-item--selected'
+          )}
+        >
+          <a
+            href={createURL(null)}
+            onClick={(event) => {
+              event.preventDefault();
+              refine(null);
```
francoischalifour, line 30: We should wrap this in `if (!isModifierClick(event))` to allow <kbd>Cmd</kbd> + Click (accessibility). (Same applies below.)

francoischalifour (commented):
Last comment and we're good.

examples/hooks/components/Breadcrumb.tsx
FabienMotte: Good point, thanks!

Haroenv (approved):

PR #3247, opened 2022-01-07
chore(examples): move examples to workspace

francoischalifour:
## Description

This updates how we handle examples in the repository to always keep them synced with our libraries version.

This fixes our failing release process.

## Changes

- We had _some_ examples in the monorepo → Move all examples to the monorepo
- We used multiple outdated versions → Bump React, Jest and Algolia dependencies
- We ran only some examples' tests → Run all the tests from the examples that have a `test` npm script


examples/hooks-next/package.json
```diff
@@ -22,6 +21,6 @@
     "@types/react": "^17.0.37",
     "eslint": "8.4.0",
     "eslint-config-next": "12.0.7",
-    "typescript": "^4.5.2"
+    "typescript": "4.5.4"
```
Haroenv, line 24: will we have some way to keep those in sync? or just paying attention when adding/updating an example to deduplicate?

scripts/test-examples.js
```diff
@@ -1,21 +1,29 @@
...
-const examples = glob.sync(
-  path.join(
-    __dirname,
-    '..',
-    'examples',
-    '!(default-theme|e-commerce|media|tourism|hooks|hooks-ssr)'
-  )
-);
+const examples = glob
+  .sync(path.join(__dirname, '..', 'examples', '!(react-native*)'))
```
Haroenv, line 7: before this PR the react native examples *were* tested, is this change expected?

examples/hooks-next/package.json
francoischalifour: It's not a big deal if we have different TypesScript versions here. I took this opportunity to bump it to the latest and to pin it.

scripts/test-examples.js
```diff
@@ -1,21 +1,29 @@
...
-  );
-});
+    return {
+      path: examplePath,
+      shouldRunTest: Boolean(pkg.scripts && pkg.scripts.test),
+    };
+  });
+
+examples
+  .filter((example) => example.shouldRunTest)
```
Haroenv, line 20: do we want to print a warning for an example without tests?

ship.config.js
```diff
@@ -1,6 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-// const getLatestVersion = require('latest-version');
```
Haroenv, line 3: when updating with learn, the version numbers of all examples will be updated, but are the dependencies also updated automatically? where's the code for that, or is it just because it's in the workspace they get updated too?

Haroenv (commented):
looks good! just have a couple questions to be sure we solve all current issues with the tooling

scripts/test-examples.js
francoischalifour: The tooling didn't make it easy. I'll try on another branch and I'll let you know.

scripts/test-examples.js
francoischalifour: I don't think so, as we didn't prior to this PR. We want to track that the examples that contain test are valid.

ship.config.js
francoischalifour: This is using the same strategy as [Autocomplete](https://github.com/algolia/autocomplete) which does update all the libraries deps in the examples with Lerna.

Haroenv (approved):
let's go

PR #3251, opened 2022-01-07
feat(hooks): introduce `usePoweredBy`

sarahdayan:
## Summary

This introduces the `usePoweredBy` Hook.

Unlike [`poweredBy`](https://www.algolia.com/doc/api-reference/widgets/powered-by/js/), this Hook doesn't take any props. Given the renderless nature of Hooks, there's no point in passing a `url` parameter—if users want to create a component with a custom link, the Hook isn't useful.

The Hook exists for consistency purposes, it doesn't retain any state. Therefore I didn't use `useConnector` but returned the render state directly, so we avoid any unnecessary overhead.

**Documentation PR:** https://github.com/algolia/doc/pull/6555

Capture d’écran 2022-01-07 à 22 48 59


examples/hooks/App.css
```diff
@@ -54,3 +54,23 @@ body {
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+}
```
sarahdayan, line 72: This visually hides the element while keeping it visible for screen readers.

packages/react-instantsearch-hooks/src/usePoweredBy.ts
```diff
@@ -0,0 +1,12 @@
+import { safelyRunOnBrowser } from 'instantsearch.js/es/lib/utils';
```
Haroenv, line 1: not 100% sure we can rely on internals like this, changing the name or export isn't a breaking change

packages/react-instantsearch-hooks/src/__tests__/usePoweredBy.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { usePoweredBy } from '../usePoweredBy';
+
+describe('usePoweredBy', () => {
+  test('returns the connector render state', () => {
+    const { url } = usePoweredBy();
+
+    expect(url).toBe(
+      `https://www.algolia.com/?utm_source=react-instantsearch&utm_medium=website&utm_content=${
+        window.location?.hostname || ''
+      }&utm_campaign=poweredby`
```
francoischalifour, line 10: Can we inline the value in the test instead of reusing the implementation code? We can't guess if the expectation is correct.

packages/react-instantsearch-hooks/src/usePoweredBy.ts
francoischalifour: Since we're in control of the IS.js version used, I think it's fine?

sarahdayan: AFAIK we don't have a common source of helpers, so we can probably copy `safelyRunOnBrowser` in the codebase for now and refactor later on. What do you think?

packages/react-instantsearch-hooks/src/usePoweredBy.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import type { PoweredByRenderState } from 'instantsearch.js/es/connectors/powered-by/connectPoweredBy';
+
+export function usePoweredBy(): PoweredByRenderState {
+  return {
+    url: `https://www.algolia.com/?utm_source=react-instantsearch&utm_medium=website&utm_content=${safelyRunOnBrowser(
```
francoischalifour, line 7: Can we extract the `safelyRunOnBrowser()` call to a variable to ease reading?

francoischalifour, line 7: Should the `utm_source` hint RIS Hooks?

packages/react-instantsearch-hooks/src/__tests__/usePoweredBy.test.tsx
sarahdayan: I did this because I wasn't sure it would always be the same in CI. But I guess we can try.

sarahdayan: Addressed in 3f06720. Also added a test for when `window.location` isn't available in 1c93b38.

packages/react-instantsearch-hooks/src/usePoweredBy.ts
sarahdayan: Addressed in c1ef574.

examples/hooks/App.css
Haroenv: shouldn't we have the test be inside the svg as a title or something?

packages/react-instantsearch-hooks/src/usePoweredBy.ts
sarahdayan: Good catch, addressed in 31afa25.

examples/hooks/App.css
sarahdayan: I used the same HTML as the [current component](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-dom/src/components/PoweredBy.js). In a previous PR we agreed not to drift for now.

Haroenv: ah yes, but React InstantSearch is the only one which diverts from the spec, InstantSearch.js has the right dom: https://github.com/algolia/instantsearch.js/blob/master/src/components/PoweredBy/PoweredBy.tsx

see also #3044

sarahdayan: :ah: Gotcha, addressed in f325e9c8.

I didn't add the `theme` option since it's a DOM option, it doesn't make sense in a Hook.

We can add a full dark theme to the demo and toggle the `ais-PoweredBy--dark`/`ais-PoweredBy--light` class, but let's maybe finish the demo with all the Hooks first, and tackle this with other UI topics such as responsiveness, wdyt?

packages/react-instantsearch-hooks/src/__tests__/usePoweredBy.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const originalWindow = { ...window };
+    const windowSpy = jest.spyOn(global, 'window', 'get');
+    const { location, ...windowWithoutLocation } = originalWindow;
+    // @ts-expect-error
+    windowSpy.mockImplementation(() => windowWithoutLocation);
+
+    const { url } = usePoweredBy();
+
+    expect(url).toBe(
+      `https://www.algolia.com/?utm_source=react-instantsearch&utm_medium=website&utm_content=&utm_campaign=poweredby`
```
Haroenv, line 22:
```suggestion
      `https://www.algolia.com/?utm_source=react-instantsearch&utm_medium=website&utm_campaign=poweredby`
```

shouldn't it be this?

francoischalifour: They're identical so I think it's fine to leave it this way, to not increase the code complexity.

francoischalifour (approved):

packages/react-instantsearch-hooks/src/__tests__/usePoweredBy.test.tsx
sarahdayan: Apparently no: https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/connectors/__tests__/connectPoweredBy.node.js#L16

The use case [isn't tested in instantsearch.js](https://github.com/algolia/instantsearch.js/blob/master/src/connectors/powered-by/__tests__/connectPoweredBy-test.ts) but [that's the behavior](https://github.com/algolia/instantsearch.js/blob/master/src/connectors/powered-by/connectPoweredBy.ts#L55-L58).

I have no idea whether it makes a difference in Google Analytics, but we've never received complaints so far I believe so it's probably no biggie. If we can avoid extra code, it's always nice.

Haroenv (approved):

PR #3258, opened 2022-01-11
test: send junit tests report to circleci

dhayab:
**Summary**

This PR enables detailed e2e tests reports in Circle CI, giving us more information regarding failing or flaky tests.


Haroenv:
what do you think about making this part of the instantsearch-e2e-tests package by default? it won't break InstantSearch (and if it does, we can make it a major), so that this config doesn't need to be duplicated

dhayab:
> what do you think about making this part of the instantsearch-e2e-tests package by default? it won't break InstantSearch (and if it does, we can make it a major), so that this config doesn't need to be duplicated

We can add the `junit` reporter to the default setting, but we would probably need to make it a breaking change, as it will require us to add `@wdio/junit-reporter` in target projects.

Haroenv:
yes, that's fine for me, a major version in a devDep for ourselves is easy to deal with and a clear marker that the junit reporter can be used

dhayab:
@Haroenv I looked into it and I'll probably take the opportunity to also bump wdio to v7 at the same time ([v5 is deprecated](https://webdriver.io/versions/) since Jan. 2021).

This would take more time, so I will create a dedicated task. Until then we could close this PR. Wdyt?

Haroenv:
I guess that makes sense, but otoh you can also only add junit there and later update? to save on time

dhayab:
That works for me: https://github.com/algolia/instantsearch-e2e-tests/pull/29.

Haroenv (approved):
thanks!

PR #3259, opened 2022-01-12
feat(hooks-server): load data twice in the case of dynamic widget usage

Haroenv:
**Summary**

dynamic widgets requires two requests to function properly, as the widgets mounted depend on what widgets are in the search result.

**Result**

if a dynamic widget is detected, getServerState calls itself a second time with the previous results provided

### DIFF
--- a/examples/hooks-next/pages/index.tsx
+++ b/examples/hooks-next/pages/index.tsx
@@ -75,7 +75,9 @@ function FallbackComponent({ attribute }: { attribute: string }) {
       <ul>
         {items.map((item) => (
           <li key={item.label}>
-            <button onClick={() => refine(item.value)}>{item.label}</button>
+            <button onClick={() => refine(item.value)}>
+              {item.isRefined ? <strong>{item.label}</strong> : item.label}
+            </button>
           </li>
         ))}
       </ul>
@@ -87,11 +89,7 @@ export async function getServerSideProps({ req }) {
   const url = new URL(
     req.headers.referer || `https://${req.headers.host}${req.url}`
   ).toString();
-  const initialServerState = await getServerState(<HomePage url={url} />);
-  // const initialServerState = {};
-  const serverState = await getServerState(
-    <HomePage url={url} serverState={initialServerState} />
-  );
+  const serverState = await getServerState(<HomePage url={url} />);
 
   return {
     props: {
--- a/examples/hooks-next/pages/index.tsx
+++ b/examples/hooks-next/pages/index.tsx
@@ -13,6 +13,8 @@ import { Highlight } from '../components/Highlight';
 import { Hits } from '../components/Hits';
 import { SearchBox } from '../components/SearchBox';
 import { history } from 'instantsearch.js/es/lib/routers/index.js';
+import { RefinementList } from '../components/RefinementList';
+import { Panel } from '../components/Panel';
 
 const client = algoliasearch('latency', '6be0576ff61c053d5f9a3225e2a90f76');
 
@@ -59,29 +61,25 @@ export default function HomePage({ serverState, url }: HomePageProps) {
           }),
         }}
       >
-        <SearchBox />
-        <DynamicWidgets fallbackComponent={FallbackComponent} />
-        <Hits hitComponent={Hit} />
+        <div className="Container">
+          <div>
+            <DynamicWidgets fallbackComponent={FallbackComponent} />
+          </div>
+          <div>
+            <SearchBox />
+            <Hits hitComponent={Hit} />
+          </div>
+        </div>
       </InstantSearch>
     </InstantSearchSSRProvider>
   );
 }
 
 function FallbackComponent({ attribute }: { attribute: string }) {
-  const { items, refine } = useRefinementList({ attribute });
   return (
-    <div>
-      <p>{attribute}</p>
-      <ul>
-        {items.map((item) => (
-          <li key={item.label}>
-            <button onClick={() => refine(item.value)}>
-              {item.isRefined ? <strong>{item.label}</strong> : item.label}
-            </button>
-          </li>
-        ))}
-      </ul>
-    </div>
+    <Panel header={attribute}>
+      <RefinementList attribute={attribute} />
+    </Panel>
   );
 }
 
--- a/examples/hooks-next/pages/index.tsx
+++ b/examples/hooks-next/pages/index.tsx
@@ -2,9 +2,11 @@ import Head from 'next/head';
 import algoliasearch from 'algoliasearch/lite';
 import { Hit as AlgoliaHit } from '@algolia/client-search';
 import {
+  DynamicWidgets,
   InstantSearch,
   InstantSearchServerState,
   InstantSearchSSRProvider,
+  useRefinementList,
 } from 'react-instantsearch-hooks';
 import { getServerState } from 'react-instantsearch-hooks-server';
 import { Highlight } from '../components/Highlight';
@@ -58,9 +60,26 @@ export default function HomePage({ serverState, url }: HomePageProps) {
         }}
       >
         <SearchBox />
+        <DynamicWidgets fallbackComponent={FallbackComponent} />
         <Hits hitComponent={Hit} />
       </InstantSearch>
     </InstantSearchSSRProvider>
+  );
+}
+
+function FallbackComponent({ attribute }: { attribute: string }) {
+  const { items, refine } = useRefinementList({ attribute });
+  return (
+    <div>
+      <p>{attribute}</p>
+      <ul>
+        {items.map((item) => (
+          <li key={item.label}>
+            <button onClick={() => refine(item.value)}>{item.label}</button>
+          </li>
+        ))}
+      </ul>
+    </div>
   );
 }
 
@@ -68,7 +87,11 @@ export async function getServerSideProps({ req }) {
   const url = new URL(
     req.headers.referer || `https://${req.headers.host}${req.url}`
   ).toString();
-  const serverState = await getServerState(<HomePage url={url} />);
+  const initialServerState = await getServerState(<HomePage url={url} />);
+  // const initialServerState = {};
+  const serverState = await getServerState(
+    <HomePage url={url} serverState={initialServerState} />
+  );
 
   return {
     props: {
--- a/examples/hooks-next/pages/index.tsx
+++ b/examples/hooks-next/pages/index.tsx
@@ -6,7 +6,6 @@ import {
   InstantSearch,
   InstantSearchServerState,
   InstantSearchSSRProvider,
-  useRefinementList,
 } from 'react-instantsearch-hooks';
 import { getServerState } from 'react-instantsearch-hooks-server';
 import { Highlight } from '../components/Highlight';
--- a/packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
+++ b/packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
@@ -7,6 +7,7 @@ import {
   InstantSearch,
   InstantSearchSSRProvider,
   Index,
+  DynamicWidgets,
   useHits,
   useRefinementList,
   useSearchBox,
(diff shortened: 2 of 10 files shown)


examples/hooks-next/pages/index.tsx
```diff
@@ -58,12 +60,31 @@ export default function HomePage({ serverState, url }: HomePageProps) {
         }}
       >
         <SearchBox />
+        <DynamicWidgets fallbackComponent={FallbackComponent} />
```
Haroenv, line 63: should I remove the dynamic widgets from the example or polish it (using a real refinement list)

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -57,6 +72,29 @@ export function getServerState(
...
+      let hasDynamicWidgets = false;
+      if (!refetching) {
+        walkIndex(searchRef.current!.mainIndex, (index) => {
+          hasDynamicWidgets =
+            hasDynamicWidgets ||
+            index
+              .getWidgets()
+              .some((widget) => widget.$$type === 'ais.dynamicWidgets');
+        });
+      }
```
Haroenv, line 86: not sure if this is the best approach? Maybe a while is better?

packages/react-instantsearch-hooks/src/DynamicWidgets.tsx
```diff
@@ -10,13 +10,19 @@ function FallbackComponent() {
   return null;
 }
 
+type AtLeastOne<
```
Haroenv, line 13: you can have no children if there's a fallback

packages/react-instantsearch-hooks/src/InstantSearchSSRProvider.tsx
```diff
@@ -22,6 +22,12 @@ export function InstantSearchSSRProvider({
   children,
   ...props
 }: InstantSearchSSRProviderProps) {
+  // If there's no server state passed, we don't want to override something
```
francoischalifour, line 25: What does "something" mean? Can we be as explicit as we can with our comments?

If I understand the code it's because we want to override the `<InstantSearchSSRProvider>` context value in the refetching of `getServerState()`: https://github.com/algolia/react-instantsearch/blob/785954e1eeab1eb1d21339ce5a6d9c2a3beab622/packages/react-instantsearch-hooks-server/src/getServerState.tsx#L88-L95

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -86,26 +124,33 @@ function waitForResults(search: InstantSearch) {
   });
 }
 
+/**
+ * Recurse over all child indices
+ *
+ * @param indexWidget initial index widget
+ * @param callback gets called for each index widget
+ * @returns void
```
francoischalifour, line 132: The function description is useful but we don't add annotations in the other functions, because they're explicit enough to be used internally, and reduces the maintenance cost. I think it's fair to remove them. Wdyt?

```suggestion
```

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -16,6 +19,18 @@ import type {
...
+}
+
+/**
+ * Implementation of getServerState
+ * @param children the application
+ * @param refetching pass true if this is the second time being called
+ */
+function renderAndFetch(
+  children: ReactNode,
+  refetching: boolean
```
francoischalifour, line 33: I find that flipping the naming makes the code and conditions easier to read.

```suggestion
  shouldRefetch: boolean
```

Then:

```ts
  let hasDynamicWidgets = false;
  if (shouldRefetch) {
    // ...
  }

  if (!shouldRefetch || !hasDynamicWidgets) {
    return serverState;
  }

  return renderAndFetch(
    <InstantSearchSSRProvider {...serverState}>
      {children}
    </InstantSearchSSRProvider>,
    false
  );
```

packages/react-instantsearch-hooks/src/InstantSearchSSRProvider.tsx
```diff
@@ -22,6 +22,12 @@ export function InstantSearchSSRProvider({
...
+  // passed from higher in the tree
+  if (Object.keys(props).length === 0) {
+    return <>{children}</>;
```
francoischalifour, line 28: I'm wondering about the implications of having `<InstantSearchSSRContext.Provider>` inconsistently in the React tree. What is the difference with rendering `<InstantSearchSSRContext.Provider value={null}>`?

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -57,6 +72,29 @@ export function getServerState(
       return {
         initialResults,
       };
+    })
+    .then((serverState) => {
```
francoischalifour, line 76:
```suggestion
    .then((serverState) => {
      // We handle <DynamicWidgets> differently than other components because it
      // requires another query to retrieve the dynamic widgets to render.
```

packages/react-instantsearch-hooks-server/src/getServerState.tsx
francoischalifour: This is not great because we keep walking the whole render tree after we have the answer we need. A recursive `for of` should do it?

packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
```diff
@@ -267,6 +277,72 @@ describe('getServerState', () => {
     expect(Object.keys(serverState.initialResults)).toHaveLength(4);
     expect(serverState.initialResults).toMatchSnapshot();
   });
+
+  test('searches twice (cached) with dynamic widgets', async () => {
```
francoischalifour, line 281: Can we add a test with a nested `<DynamicWidgets>`?

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: what do you think about `canRefetch` with your logic?

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: for of gets polyfilled with generators (even though this code is technically only for node), but I can rewrite this in a while that doesn't read after detecting a dynamic widget (although these trees never would be *that* big that the number of calls to getWidgets would be superfluous)

packages/react-instantsearch-hooks-server/src/getServerState.tsx
francoischalifour: "can" doesn't reflect what the argument is doing, is it?

examples/hooks-next/pages/index.tsx
sarahdayan: Why not use the same as in the non-SSR example?

packages/react-instantsearch-hooks-server/src/getServerState.tsx
sarahdayan: +1 on "should"; it's a flag, not a test.

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -16,6 +19,18 @@ import type {
...
 export function getServerState(
   children: ReactNode
+): Promise<InstantSearchServerState> {
+  return renderAndFetch(children, false);
+}
+
+/**
+ * Implementation of getServerState
+ * @param children the application
+ * @param refetching pass true if this is the second time being called
```
sarahdayan, line 29: This description is confusing, we should clarify it.

Maybe:

>Whether to fetch a second time after first rendering.

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: I'm not sure if "shouldRefetch" being true for the initial call, as it should only refetch if that flag is set + it is actually using dynamic widgets; it won't refetch in other cases

francoischalifour: I think that the code is not correctly coupled. We don't need this flag.

Here's a proposal, without the concept of "is refetching". I renamed `hasDynamicWidgets` to a generic `shouldRefetch` though, but there's only one variable to track this value.

```tsx
export function getServerState(
  children: ReactNode
): Promise<InstantSearchServerState> {
  const searchRef: { current: InstantSearch | undefined } = {
    current: undefined,
  };

  const notifyServer: InstantSearchServerContextApi['notifyServer'] = ({
    search,
  }) => {
    searchRef.current = search;
  };

  // eslint-disable-next-line no-shadow
  function execute(children: ReactNode) {
    ReactDOM.renderToString(
      <InstantSearchServerContext.Provider value={{ notifyServer }}>
        {children}
      </InstantSearchServerContext.Provider>
    );

    // We wait for the component to mount so that `notifyServer()` is called.
    return new Promise((resolve) => setTimeout(resolve, 0))
      .then(() => {
        // If `notifyServer()` is not called by then, it means that <InstantSearch>
        // wasn't within the `children`.
        // We decide to go with a strict behavior in that case; throwing. If users have
        // some routes that don't mount the <InstantSearch> component, they would need
        // to try/catch the `getServerState()` call.
        // If this behavior turns out to be too strict for many users, we can decide
        // to warn instead of throwing.
        if (!searchRef.current) {
          throw new Error(
            "Unable to retrieve InstantSearch's server state in `getServerState()`. Did you mount the <InstantSearch> component?"
          );
        }

        return waitForResults(searchRef.current);
      })
      .then(() => {
        const initialResults = getInitialResults(searchRef.current!.mainIndex);

        return {
          initialResults,
        };
      });
  }

  return execute(children).then((serverState) => {
    let shouldRefetch = false;

    // <DynamicWidgets> requires another query to retrieve the dynamic widgets
    // to render.
    walkIndex(searchRef.current!.mainIndex, (index) => {
      shouldRefetch =
        shouldRefetch ||
        index
          .getWidgets()
          .some((widget) => widget.$$type === 'ais.dynamicWidgets');
    });

    if (shouldRefetch) {
      return execute(
        <InstantSearchSSRProvider {...serverState}>
          {children}
        </InstantSearchSSRProvider>
      );
    }

    return serverState;
  });
}
```

The code is more linear, it follows the data flow. It's also less error prone because there's less back and forth and we don't keep track of a state variable.

packages/react-instantsearch-hooks/src/components/InstantSearchSSRProvider.tsx
```diff
@@ -22,6 +22,13 @@ export function InstantSearchSSRProvider({
   children,
   ...props
 }: InstantSearchSSRProviderProps) {
+  // When there is dynamic widgets on the page, a second Provider is used
+  // above the user-land InstantSearchSSRProvider. To avoid the user's provider
+  // overriding the value with an empty object, this provider is skipped
```
francoischalifour, line 27:
```suggestion
  // When <DynamicWidgets> is mounted, a second provider is used above the user-land
  // <InstantSearchSSRProvider> in `getServerState()`.
  // To avoid the user's provider overriding the context value with an empty object,
  // we skip this provider.
```

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -27,37 +30,64 @@ export function getServerState(
...
+
+    // <DynamicWidgets> requires another query to retrieve the dynamic widgets
+    // to render.
+    walkIndex(searchRef.current!.mainIndex, (index) => {
+      shouldRefetch =
+        shouldRefetch ||
+        index
+          .getWidgets()
+          .some((widget) => widget.$$type === 'ais.dynamicWidgets');
     });
```
francoischalifour, line 79: Do you want to refactor this logic to break the loop early in this PR?

Haroenv: I thought the numbers of index widgets likely isn't big enough to make a difference

francoischalifour: To me the main argument is that it's hard to understand this code, based on what it's doing. But I'm fine addressing this later.

Let's keep this conversation unresolved to keep it visible then.

packages/react-instantsearch-hooks/src/components/InstantSearchSSRProvider.tsx
```diff
@@ -22,6 +22,14 @@ export function InstantSearchSSRProvider({
   children,
   ...props
 }: InstantSearchSSRProviderProps) {
+  // When <DynamicWidgets> is mounted, a second provider is used above the user-land
+  // <InstantSearchSSRProvider> in `getServerState()`.
+  // To avoid the user's provider overriding the context value with an empty object,
+  // we skip this provider.
+  if (Object.keys(props).length === 0) {
+    return <>{children}</>;
```
francoischalifour, line 30: This should be sufficient?

```suggestion
    return children;
```

francoischalifour (approved):
Great change!

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: alternative:

```js
    let allWidgets: string[] = [];
    walkIndex(searchRef.current!.mainIndex, (index) => {
      allWidgets = allWidgets.concat(
        index.getWidgets().map((widget) => widget.$$type)
      );
    });
    // <DynamicWidgets> requires another query to retrieve the dynamic widgets
    // to render.
    const shouldRefetch = allWidgets.indexOf('ais.dynamicWidgets') > -1;
```

packages/react-instantsearch-hooks/src/components/InstantSearchSSRProvider.tsx
Haroenv: TS failed without the fragment, as the type became `{} | JSX.Element`, similar issue as to https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33006 I think

sarahdayan: Yep, I often run into this issue and wrapping with a fragment is the way to go for now.

sarahdayan (approved):

PR #3268, opened 2022-01-20
feat(server): load data twice in the case of dynamic widget usage

Haroenv:
**Summary**

Same as #3259, but for React InstantSearch core, not hooks.

We render twice in the case that dynamic widgets are discovered

**Result**

- no search request when there's a refinement applied using server side rendering
- no flash of non-filtered content
- https://codesandbox.io/s/keen-moon-wdxou fixed by this PR (select a refinement and reload)

### DIFF
--- a/packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
+++ b/packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
@@ -550,7 +550,7 @@ describe('findResultsState', () => {
       );
     });
 
-    it.only('searches twice with dynamic widgets and a refinement', async () => {
+    it('searches twice with dynamic widgets and a refinement', async () => {
       const RefinementList = connectRefinementList(() => null);
       const App = (props) => (
         <InstantSearch {...props}>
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch-dom/src/core/__tests__/createInstantSearchServer.js
```diff
@@ -503,6 +524,78 @@ describe('findResultsState', () => {
...
+          maxValuesPerFacet: 20,
+          tagFilters: '',
+        },
+      });
+
+      // second query does have the fallback widget mounted, and thus also the refinement
+      expect(props.searchClient.search.mock.calls[1][0][0]).toEqual({
+        indexName: 'instant_search',
+        params: {
+          facetFilters: [['categories:refined!']],
```
Haroenv, line 590: according to the test, facetFilters isn't present in this case, but it is in multi-index context. When I'm trying it out manually in the Next app, it seems that the right parameters are used.

Haroenv: so in the test the refinement list doesn't mount, but it does mount when it's nested in an index. as far as I can tell the results are passed correctly though, so I don't yet see the difference between the two cases

Haroenv: solved in 77ce0f1d

francoischalifour (approved):

sarahdayan (approved):

PR #3274, opened 2022-01-24
chore(release): bump all lib deps on release

francoischalifour:
This fixes our release process by updating all our lib dependencies (including the ones in our examples) on release. The fix is to rely on Lerna to update the deps, and not Ship.js.

You can see #3273 that shows the output of this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-replace": "2.2.0",
     "rollup-plugin-uglify": "6.0.3",
-    "shipjs": "0.24.0",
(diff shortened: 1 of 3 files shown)


Haroenv (approved):

francoischalifour:
In e96d33537484c6c1d0f6b4dcfeec1bb556a10f44 I also fix an issue where Ship.js's [`getAppName()`](https://github.com/algolia/shipjs/blob/2a8e7c569754dd57c71ff684a43ae8cfc8d1f384/packages/shipjs-lib/src/lib/util/getAppName.js) is unable to retrieve the lib name because we're in a monorepo. So I updated the automatic Slack message.

image

PR #3276, opened 2022-01-24
wip: apply initial search parameters in hooks

Haroenv:
**Summary**

the state of "results" when not doing server side rendering isn't aware of the widgets yet. This is incorrect, and thus we need to call getWidgetSearchParameters before rendering.

This is still wrong when the state has been set by another widget, in the case of breadcrumb/hierarchical menu. I think that can be fixed by having both widgets apply the same state (maybe the already do it, I didn't test)

**Result**

slower initial mount, but actually correct 📦

### DIFF
--- a/packages/react-instantsearch-hooks/src/connectors/__tests__/useSearchBox.test.tsx
+++ b/packages/react-instantsearch-hooks/src/connectors/__tests__/useSearchBox.test.tsx
@@ -28,4 +28,35 @@ describe('useSearchBox', () => {
       refine: expect.any(Function),
     });
   });
+
+  test('returns the connector render state with ui state', async () => {
+    const wrapper = createInstantSearchTestWrapper({
+      initialUiState: {
+        indexName: {
+          query: 'testio',
+        },
+      },
+    });
+    const { result, waitForNextUpdate } = renderHook(() => useSearchBox(), {
+      wrapper,
+    });
+
+    // Initial render state from manual `getWidgetRenderState`
+    expect(result.current).toEqual({
+      query: 'testio',
+      isSearchStalled: false,
+      clear: expect.any(Function),
+      refine: expect.any(Function),
+    });
+
+    await waitForNextUpdate();
+
+    // InstantSearch.js state from the `render` lifecycle step
+    expect(result.current).toEqual({
+      query: 'testio',
+      isSearchStalled: false,
+      clear: expect.any(Function),
+      refine: expect.any(Function),
+    });
+  });
 });
--- a/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
+++ b/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
@@ -100,7 +100,13 @@ export function useConnector<
         instantSearchInstance: search,
         results,
         scopedResults,
-        state: results._state,
+        state: widget.getWidgetSearchParameters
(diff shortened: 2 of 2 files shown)


packages/react-instantsearch-hooks/src/hooks/useConnector.ts
```diff
@@ -100,7 +100,13 @@ export function useConnector<
         instantSearchInstance: search,
         results,
         scopedResults,
-        state: results._state,
+        state: widget.getWidgetSearchParameters
```
Haroenv, line 103: this probably only needs to be done in the case of a synthetic results?

packages/react-instantsearch-hooks/src/connectors/__tests__/useSearchBox.test.tsx
```diff
@@ -28,4 +28,35 @@ describe('useSearchBox', () => {
...
+        },
+      },
+    });
+    const { result, waitForNextUpdate } = renderHook(() => useSearchBox(), {
+      wrapper,
+    });
+
+    // Initial render state from manual `getWidgetRenderState`
+    expect(result.current).toEqual({
+      query: 'testio',
```
Haroenv, line 46: before the changes in useConnector, this is `query: ''`

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
```diff
@@ -75,7 +75,16 @@ export function useConnector<
...
+        parentIndex.getResults() ||
+        createSearchResults(
+          widget.getWidgetSearchParameters
+            ? widget.getWidgetSearchParameters(helper.state, {
+                uiState: parentIndex.getWidgetUiState({})[
+                  parentIndex.getIndexId()
+                ],
+              })
+            : helper.state
+        );
```
francoischalifour, line 87:
```suggestion
        parentIndex.getResults() ||
        createSearchResults(
          widget.getWidgetSearchParameters?.(helper.state, {
            uiState: parentIndex.getWidgetUiState({})[parentIndex.getIndexId()],
          }) || helper.state
        );
```

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
```diff
@@ -75,7 +75,12 @@ export function useConnector<
       // The helper exists because we've started InstantSearch.
       const helper = parentIndex.getHelper()!;
       const results =
-        parentIndex.getResults() || createSearchResults(helper.state);
+        parentIndex.getResults() ||
+        createSearchResults(
+          widget.getWidgetSearchParameters?.(helper.state, {
+            uiState: parentIndex.getWidgetUiState({})[parentIndex.getIndexId()],
+          }) || helper.state
+        );
```
francoischalifour, line 83:
```suggestion
        // On SSR, we get the results injected on the Index.
        parentIndex.getResults() ||
        // On the browser, we create fallback results based on the widget's
        // `getWidgetSearchParameters()` method to inject the initial UI state,
        // or fall back to the helper state.
        createSearchResults(
          widget.getWidgetSearchParameters?.(helper.state, {
            uiState: parentIndex.getWidgetUiState({})[parentIndex.getIndexId()],
          }) || helper.state
        );
```

francoischalifour (approved):

dhayab (approved):

PR #3283, opened 2022-01-28
fix(hooks): enable pause on exceptions on warning

francoischalifour:
## Description

This slightly improves the DX by updating our `warning()` util to throw an error and catch it immediately as a convenience so that you can find the source of the warning that appears in the console by enabling "Pause on exceptions" in the JavaScript debugger.

A similar strategy is used in:

- [React](https://github.com/facebook/fbjs/blob/865675200037a867727c7361ff06be772effe169/packages/fbjs/src/__forks__/warning.js#L28-L33)
- [React Router](https://github.com/remix-run/react-router/blob/f16c5490dfa75f15dcfb86d2a981a7c58a9d1a33/packages/react-router/index.tsx#L27-L35)

## Preview

Screen Shot 2022-01-28 at 11 38 16

## Resources

- [Automatically pause on any exception](https://developers.google.com/web/updates/2015/05/automatically-pause-on-any-exception)

### DIFF
--- a/packages/react-instantsearch-hooks/src/lib/warn.ts
+++ b/packages/react-instantsearch-hooks/src/lib/warn.ts
@@ -1,3 +1,5 @@
+/* eslint-disable no-console, no-empty */
+
 export const warnCache = {
   current: {},
 };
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-hooks/src/lib/warn.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-console, no-empty */
```
francoischalifour, line 1: I moved these ESLint ignore rules at the top of the file to decluster our "Welcome to debugging" message.

dhayab (approved):
Nice!

sarahdayan (approved):
Lovely! Let's have that in our other libraries.

Haroenv (approved):

PR #3351, opened 2022-02-10
docs(examples): introduce React InstantSearch Hooks with React Native example

sarahdayan:
## Summary

This introduces a minimal example of React Native usage with React InstantSearch Hooks. It's based on the [`doc-code-samples` example](https://github.com/algolia/doc-code-samples/tree/master/React%20InstantSearch%20Hooks%20Native/getting-started), without the filters to keep it simple.

This is mainly intended to be used as a way to check React Native usage still works in new PRs.


Haroenv (changes requested):
the folder should be in the workspace so that the new yarn lock isn't needed, right? otherwise the test won't test the local React InstantSearch version.

Did you do it to avoid ballooning the main lockfile?

sarahdayan:
@Haroenv No I just forgot to remove it.

By the way, I noticed that other examples have their own lockfile—is there a good reason?

Haroenv:
> By the way, I noticed that other examples have their own lockfile—is there a good reason?

No, I guess they also should be removed (in a separate PR)

Haroenv (approved):

PR #3358, opened 2022-02-16
feat(hooks-dom): introduce Hooks DOM package

Haroenv:
**Summary**

Introduces the `react-instantsearch-hooks-dom` package

**Result**

A new package, which only re-exports the API of React InstantSearch hooks

### DIFF
--- a/packages/react-instantsearch-hooks-dom/src/index.ts
+++ b/packages/react-instantsearch-hooks-dom/src/index.ts
@@ -1,0 +1,1 @@
+export * from 'react-instantsearch-hooks';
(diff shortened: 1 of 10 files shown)


packages/react-instantsearch-hooks-dom/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from 'react-instantsearch-hooks';
```
dhayab, line 1: Do we re-export all the hooks from `react-instantsearch-hooks`? I might have misread it, but my understanding of the RFC was to only re-export renderless components (like `<Configure />`).

Haroenv: We export everything, so a user doesn't need to add both -hooks and -hooks-dom. This is the same as -core and -dom today :)

dhayab (approved):

PR #3359, opened 2022-02-16
feat(hooks): allow additional widget properties to be passed from hooks

Haroenv:
**Summary**

widgetType (one of the additional properties) gets used to distinguish between builtin components and user-defined components, which either have no widgetType or a user-defined one.

This also allows a user to override a specific life cycle of a widget, if that would be needed

**Result**

- useConnector accepts "additional properties" that get set on widget instance
- all hooks accept "additional properties" optionally
- builtin components (configure, dynamic widgets) set the default widgetType, but don't allow further "additional properties" (can be added later without breaking change)

### DIFF
--- a/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
+++ b/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
@@ -15,7 +15,7 @@ export function useConnector<
 >(
   connector: Connector<TDescription, TProps>,
   props: TProps = {} as TProps,
-  widgetType?: string
+  widgetType: string | undefined
 ): TDescription['renderState'] {
   const serverContext = useInstantSearchServerContext();
   const search = useInstantSearchContext();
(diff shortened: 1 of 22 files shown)


dhayab (approved):

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
```diff
@@ -14,7 +14,8 @@ export function useConnector<
   TDescription extends WidgetDescription
 >(
   connector: Connector<TDescription, TProps>,
-  props: TProps = {} as TProps
+  props: TProps = {} as TProps,
+  widgetType: string | undefined
```
francoischalifour, line 18: We could rather pass the object signature as an object (`{ $$widgetType }`) and spread it on the widget instance instead of just the string widget type. Here, when we need to introduce something new, it will force us to introduce yet another argument, with this one being able to be undefined.

francoischalifour (changes requested):

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
Haroenv: Yes, you're right, I copied from the WIP branch instead of the RFC; good catch!

francoischalifour (approved):

PR #3360, opened 2022-02-17
feat(hooks): implement SearchBox

Haroenv:
**Summary**

Implements the SearchBox component

FX-1144

**Result**

- [x] SearchBox UI component (based on ControlledSearchBox from the example)
- [x] SearchBox widget
- [x] update examples to use builtin widget
- [ ] write new tests (something is wrong with Jest?)


dhayab (approved):
👏

packages/react-instantsearch-hooks-dom/src/widgets/SearchBox.tsx
```diff
@@ -1,12 +1,22 @@
...
-export type SearchBoxProps = React.ComponentProps<'div'> & UseSearchBoxProps;
+import type { SearchBoxProps as SearchBoxUiComponentProps } from '../ui/SearchBox';
+import type { ChangeEvent } from 'react';
+import type { UseSearchBoxProps } from 'react-instantsearch-hooks';
+
+export type SearchBoxProps = Omit<
+  SearchBoxUiComponentProps,
+  'inputRef' | 'isSearchStalled' | 'onChange' | 'onReset' | 'value'
+> &
+  UseSearchBoxProps;
```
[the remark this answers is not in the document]
Haroenv: We are spreading the props, and overriding a couple of them, so Omit fits best. (note that this includes all html div props, as expected)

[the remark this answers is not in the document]
dhayab: Ah, indeed. That's clear now. Thank you!

FabienMotte (approved):

packages/react-instantsearch-hooks-dom/package.json
```diff
@@ -52,5 +52,8 @@
   "peerDependencies": {
     "react": ">= 16.8.0 < 19",
     "react-dom": ">= 16.8.0 < 19"
+  },
+  "devDependencies": {
+    "@testing-library/user-event": "^13.5.0"
   }
```
francoischalifour, line 58: This should be declared at the root package.

examples/hooks-next/components/ControlledSearchBox.tsx
```diff
@@ -1,6 +1,8 @@
-import type { ChangeEvent, FormEvent, RefObject, ComponentProps } from 'react';
+import React from 'react';
 
-export type ControlledSearchBoxProps = ComponentProps<'div'> & {
+import type { ChangeEvent, FormEvent, RefObject } from 'react';
+
+export type ControlledSearchBoxProps = React.ComponentProps<'div'> & {
```
Haroenv, line 5:
```suggestion
export type ControlledSearchBoxProps = React.HTMLAttributes<HTMLDivElement> & {
```

Haroenv: not needed

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
```diff
@@ -0,0 +1,162 @@
...
+  inputRef: RefObject<HTMLInputElement>;
+  isSearchStalled: boolean;
+  onChange(event: ChangeEvent): void;
+  onReset(event: FormEvent): void;
+  onSubmit?(event: FormEvent): void;
+  placeholder?: string;
+  value: string;
+  resetIconComponent?: JSXElementConstructor<Record<string, any>>;
+  submitIconComponent?: JSXElementConstructor<Record<string, any>>;
+  loadingIconComponent?: JSXElementConstructor<Record<string, any>>;
```
francoischalifour, line 22: It would make more sense to accept elements instead of components here, don't you think?

Haroenv: I went with components, as that way the classNames can be passed (in a later PR), but if we don't think that's important, can be changed to element as well I guess

PR #3363, opened 2022-02-18
feat(hooks): implement Hits component

Haroenv:
**Summary**

Implements the Hits component

FX-1167

**Result**

- [x] New Hits component (generic)
- [x] write correct tests of Hits with actual hits asserted in the dom
- [x] remove Hits from all examples


packages/react-instantsearch-hooks-dom/src/widgets/Hits.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Hits as HitsUiComponent } from '../ui/Hits';
+
+import type { HitsProps as HitsUiComponentProps } from '../ui/Hits';
+import type { Hit } from 'instantsearch.js';
+import type { UseHitsProps } from 'react-instantsearch-hooks';
+
+export type HitsProps<THit> = Omit<HitsUiComponentProps<Hit & THit>, 'hits'> &
+  UseHitsProps;
+
+export function Hits<THit extends Record<string, any> = Record<string, any>>(
```
sarahdayan, line 13:
```suggestion
export function Hits<THit extends Record<string, unknown> = Record<string, unknown>>(
```

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+import { cx } from './lib/cx';
+
+import type { Hit } from 'instantsearch.js';
+import type { JSXElementConstructor } from 'react';
+
+export type HitsProps<THit> = React.ComponentProps<'div'> & {
```
sarahdayan, line 8: There's the following note in the React types for `ComponentProps`:

```txt
NOTE: prefer ComponentPropsWithRef, if the ref is forwarded,
or ComponentPropsWithoutRef when refs are not supported.
```

We don't forward `ref`s in this component, but the user could technically pass one (although it wouldn't work and generate a warning in the console). Could we hint them by using `ComponentPropsWithoutRef`? Also, what's the pro of using `React.ComponentProps` vs. `React.HTMLAttributes<HTMLDivElement>`? I usually go for the latter, interested in the choice.

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
```diff
@@ -0,0 +1,44 @@
...
+  hits: THit[];
+  hitComponent?: JSXElementConstructor<{ hit: THit }>;
```
sarahdayan, line 10: Why import this type instead of referencing it from `React` like you did with `ComponentProps`?

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+function DefaultHitComponent({ hit }: { hit: Hit }) {
+  return (
+    <div
+      style={{
+        borderBottom: '1px solid #bbb',
+        paddingBottom: '5px',
+        marginBottom: '5px',
```
sarahdayan, line 19: Why are those needed?

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+import { cx } from './lib/cx';
+
+import type { Hit } from 'instantsearch.js';
```
sarahdayan, line 5: Don't we order imports?

packages/react-instantsearch-hooks-dom/src/widgets/Hits.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { useHits } from 'react-instantsearch-hooks';
+
+import { Hits as HitsUiComponent } from '../ui/Hits';
+
+import type { HitsProps as HitsUiComponentProps } from '../ui/Hits';
+import type { Hit } from 'instantsearch.js';
```
sarahdayan, line 7: Don't we order imports?

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+import React from 'react';
+
+import { Hits } from '../Hits';
+
+import type { HitsProps } from '../Hits';
+import type { Hit } from 'instantsearch.js';
+
+describe('Hits', () => {
+  function createProps<THit extends Hit = Hit>(
+    props: Partial<HitsProps<THit>>
```
sarahdayan, line 11:
```suggestion
    props: Partial<HitsProps<THit>> = {}
```

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+      hits: [
+        { objectID: 'abc', __position: 1 },
+        { objectID: 'def', __position: 2 },
+      ],
+      ...props,
+    };
+  }
+
+  test('renders with default props', () => {
+    const props = createProps({});
```
sarahdayan, line 23:
```suggestion
    const props = createProps();
```

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  test('renders with custom hitComponent', () => {
+    const props = createProps({
+      hitComponent: ({ hit }) => <strong>{hit.objectID}</strong>,
+    });
+
+    const { container } = render(<Hits {...props} />);
+
+    expect(container.querySelectorAll('strong')).toHaveLength(2);
```
sarahdayan, line 68: Is this needed? The following assertion covers it.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+              <strong>
+                def
+              </strong>
+            </li>
+          </ol>
+        </div>
+      </div>
+    `);
+  });
+});
```
sarahdayan, line 97: Can we also add a test to ensure a custom `className` is concatenated on the root element? And maybe forwarding some `<div>` attributes?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+  children: React.ReactNode;
+} & Partial<InstantSearchProps>;
+
+function SearchProvider({ children, ...props }: SearchProviderProps) {
+  return (
+    <InstantSearch searchClient={searchClient} indexName="indexName" {...props}>
+      {children}
+    </InstantSearch>
+  );
+}
```
sarahdayan, line 26: We'll likely need this util for all other components, might it make sense to move it now?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+describe('Hits', () => {
+  test('renders with default props', async () => {
+    const { container } = render(
+      <SearchProvider>
+        <Hits />
+      </SearchProvider>
+    );
+
+    await waitFor(() => {
+      expect(container.querySelector('.ais-Hits')).toMatchInlineSnapshot(`
```
sarahdayan, line 37: Where does the wrapping `<div>` around `.ais-Hits` come from?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+    });
+
+    const { container } = render(
+      <SearchProvider searchClient={client}>
+        <Hits hitComponent={({ hit }) => <strong>{hit.__position}</strong>} />
+      </SearchProvider>
+    );
+
+    await waitFor(() => {
+      expect(container.querySelectorAll('strong')).toHaveLength(3);
```
sarahdayan, line 71: Same comment as earlier.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+    });
+  });
+
+  test('renders with custom hitComponent', async () => {
+    const client = createSearchClient({
+      search: (requests) =>
+        Promise.resolve(
+          createMultiSearchResponse(
+            ...requests.map((request) =>
+              createSingleSearchResponse<Record<string, any>>({
```
sarahdayan, line 55:
```suggestion
              createSingleSearchResponse<Record<string, unknown>>({
```

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Hits.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+                3
+              </strong>
+            </li>
+          </ol>
+        </div>
+      `);
+    });
+  });
+
+  test('renders with a non-default hit shape', async () => {
```
sarahdayan, line 106: Does this need its own test? Could we merge it with the previous one?

sarahdayan (commented):
I left mostly nits and questions but it looks clean!

packages/react-instantsearch-hooks-dom/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from 'react-instantsearch-hooks';
+export * from './widgets/Hits';
```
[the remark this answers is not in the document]
Haroenv: No reason really, I thought it could be easy to miss `index`, as it's in the middle of the files, but we already use the pattern in some other places

packages/react-instantsearch-hooks-dom/src/widgets/Hits.tsx
Haroenv: Not fully the same, but indeed better

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
Haroenv: I just copied this from somewhere else, `React.HTMLAttributes<HTMLDivElement>` seems easiest indeed

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
Haroenv: I guess the border is otherwise really close to the text? This is the same default component as in the current React InstantSearch, I didn't question it

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
Haroenv: they are imported, this is a type import which comes last (eslint deals with it)

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
sarahdayan: Definitely makes sense if there's a border to have padding, but I was questioning the border as well.

Ultimately we don't care much since this won't be used in production, but changing it is _technically_ a breaking change and we'll soon have new UI guidelines to apply. Maybe we can stay conservative on this one and just make sure it doesn't generate a horizontal scrollbar, as the `word-break` does, wdyt?

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Hits.test.tsx
Haroenv: A snapshot is easy to accidentally update wrong, so I like to have a quick assertion of the "important part" as well

sarahdayan: Gotcha. The snapshot seems quite readable to me, but this second assertion has virtually zero performance impact so I'm fine if you want to leave it.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Hits.test.tsx
sarahdayan: Seems to be added by default by Testing Library. We could assert the full things though. This way we test we don't accidentally add any additional markup around the UI component in the widget.

packages/react-instantsearch-hooks-dom/src/ui/Hits.tsx
Haroenv:
```suggestion
export type HitsProps<THit> = React.HTMLAttributes<HTMLDivElement> & {
```

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Hits.test.tsx
Haroenv: I'd leave that as-is, no need for variadic call signatures in tests

sarahdayan: As you prefer.

packages/react-instantsearch-hooks-dom/src/widgets/Hits.tsx
```diff
@@ -0,0 +1,22 @@
...
+export type HitsProps<THit> = Omit<HitsUiComponentProps<Hit & THit>, 'hits'> &
+  UseHitsProps;
+
+export function Hits<THit extends Record<string, any> = Record<string, any>>(
+  props: HitsProps<THit>
+) {
+  // @TODO: make useHit generic so this cast is no longer needed.
+  // for that, connectHits needs to be generic too,
+  // and therefore Connector as well
+  const hits = useHits(props).hits as Array<THit & Hit>;
```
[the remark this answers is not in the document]
Haroenv: was easier to add here after all https://github.com/algolia/react-instantsearch/pull/3363/commits/53e4929ae5c4e6afb2992edd3865253fbd78d542

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Hits.test.tsx
Haroenv: done in 469540b2

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      name,
+      renderedWidgets,
+      widget: renderedWidgets[0],
+    };
+  });
+}
+
+describe('widgets', () => {
+  const widgets = initializeWidgets();
+  test('renders one widget', () => {
```
sarahdayan, line 53:
```suggestion

  test('renders one widget', () => {
```

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+        name,
+        renderedWidgets: expect.objectContaining({
+          length: 1,
+        }),
+      });
+    });
+  });
+
+  describe('$$type', () => {
+    test('present in every widget', () => {
```
sarahdayan, line 65:
```suggestion
    test('is present in every widget', () => {
```

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      widgets.forEach(({ name, renderedWidgets }) => {
+        renderedWidgets.forEach((widget) =>
+          expect([name, widget.$$type]).toEqual([name, expect.any(String)])
+        );
+      });
+    });
+
+    test('starts with ais.', () => {
```
sarahdayan, line 73:
```suggestion
    test('starts with "ais."', () => {
```

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      widgets.forEach(({ name, widget }) =>
+        expect([name, widget.$$type.substr(0, 4)]).toEqual([name, 'ais.'])
+      );
+    });
+  });
+
+  describe('$$widgetType', () => {
+    test('present in every widget', () => {
```
sarahdayan, line 81:
```suggestion
    test('is present in every widget', () => {
```

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      widgets.forEach(({ name, widget }) =>
+        expect([name, widget.$$widgetType]).toEqual([name, expect.any(String)])
+      );
+    });
+
+    test('starts with ais.', () => {
```
sarahdayan, line 87:
```suggestion
    test('starts with "ais."', () => {
```

sarahdayan, line 87: Maybe instead of asserting the prefix, we can map all values into an array of string.

This way we test:
1. That every widget has it (no wholes in the array)
2. That the names are stable (necessary for metadata not to have them change)

Since imports are alphabetically sorted with ESLint, this will only need to be updated when we add a new widget.

wdyt?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
sarahdayan, line 65: See other suggestion below.

test/utils/InstantSearchHooksTestWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { createSearchClient } from '../mock';
+
+import type { InstantSearchProps } from 'react-instantsearch-hooks';
+
+const searchClient = createSearchClient();
+type InstantSearchHooksTestWrapperProps = {
+  children: React.ReactNode;
+} & Partial<InstantSearchProps>;
+export function InstantSearchHooksTestWrapper({
```
sarahdayan, line 12:
```suggestion
const searchClient = createSearchClient();

type InstantSearchHooksTestWrapperProps = {
  children: React.ReactNode;
} & Partial<InstantSearchProps>;

export function InstantSearchHooksTestWrapper({
```

sarahdayan (commented):
Some final thoughts.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
Haroenv: Maybe the existing tests + an extra one that asserts the list of types? I'm not sure when we add a new item we will remember they need to have these requirements

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/index.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+      widgets.forEach(({ name, widget }) =>
+        expect([name, widget.$$widgetType!.substr(0, 4)]).toEqual([
+          name,
+          'ais.',
+        ])
+      );
+    });
+  });
+
+  test('name, type and widgetType are equivalent', () => {
```
sarahdayan, line 108:
```suggestion
  test('name, $$type and $$widgetType are equivalent', () => {
```

sarahdayan (approved):

PR #3368, opened 2022-02-24
feat(hooks): implement CurrentRefinements component

sarahdayan:
## Summary

This introduces the `CurrentRefinements` DOM widget.

Ticket: [FX-1151](https://algolia.atlassian.net/browse/FX-1151)

## Results

- New `CurrentRefinements` widget
- New `isModifierClick` utility + tests (reused from examples)
- Replaced `CurrentRefinements` with the widget in example


Haroenv:
Seems like the wrong list of commits as it was based on an older version of the hits PR, could you rebase?

sarahdayan:
@Haroenv Yes I plan do do it later, this doesn't prevent reviewing though because the diff is the same.

dhayab (approved):
Looks great!

We could probably remove **CurrentRefinements.tsx** from the hooks example project now it's been replaced with the proper component.

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { cx } from './lib/cx';
+
+type CurrentRefinementCategory = Pick<
+  React.ButtonHTMLAttributes<HTMLButtonElement>,
+  'onClick'
+> & { label: string };
+
+type CurrentRefinement = {
+  label: string;
+  category: CurrentRefinementCategory[];
```
dhayab, line 12: _Feel free to disregard this nitpicking comment_: we could pluralize this attribute name as `categories` as it references an array.

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { CurrentRefinements as CurrentRefinementsUiComponent } from '../ui/CurrentRefinements';
+
+import { isModifierClick } from './lib/isModifierClick';
+
+import type { CurrentRefinementsProps as CurrentRefinementsUiComponentProps } from '../ui/CurrentRefinements';
+import type { UseCurrentRefinementsProps } from 'react-instantsearch-hooks';
+
+export type CurrentRefinementsProps = Omit<
+  CurrentRefinementsUiComponentProps,
+  'disabled' | 'onClick'
```
dhayab, line 13: I don't think we should expose `items` and `hasRefinements` from the UI props, if I'm not mistaken.

```suggestion
  'disabled' | 'onClick' | 'items' | 'hasRefinements'
```

sarahdayan:
@dhayab We've decided to leave the example components for now because we link them from the documentation. They're removed from the barrel (so they're no longer exported) and we use the new widget in the example, so this is dead code anyway. We'll remove them all at once later, when we update the docs.

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
sarahdayan: I used this naming to be fully 1:1 with the matching class name, but we could indeed slightly deviate.

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
sarahdayan: Good catch!

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
```diff
@@ -0,0 +1,55 @@
...
+        label: item.label,
+        category: item.refinements.map((refinement) => ({
+          label: refinement.label,
+          onClick: (event) => {
+            if (isModifierClick(event)) {
+              return;
+            }
+            event.preventDefault();
+
+            refine(refinement);
```
Haroenv, line 49: this onClick should be part of the ui component IMO and category passing refine as expected. It's similarly coupled in both cases, but this mapping doesn't seem needed to me

Haroenv (commented):
looks great! some changes to what is the ui component and what is the widget maybe?

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
sarahdayan: Addressed in 87e5c9fc.

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
```diff
@@ -0,0 +1,61 @@
...
+            {item.category.map(({ label, onClick }) => (
+              <span key={label} className="ais-CurrentRefinements-category">
+                <span className="ais-CurrentRefinements-categoryLabel">
+                  {label}
+                </span>
+                <button
+                  onClick={onClick}
+                  className="ais-CurrentRefinements-delete"
+                >
+                  ✕
```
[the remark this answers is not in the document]
sarahdayan: The spec doesn't describe an `aria-label`, and we don't have one in InstantSearch.js. Similarly, the `currentRefinements` doesn't provide a template for the button content (same in all flavors).

Didn't we agree to stay spec compliant for now and improve it later on in an accessibility-focused task?

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
sarahdayan: I find the coupling looser this way, and less leaky of the InstantSearch API. We do the same in [`SearchBox`](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-hooks-dom/src/widgets/SearchBox.tsx#L58-L59)—we're not passing `setValue` directly, but callbacks instead.

Open to discuss this (along with @dhayab and @francoischalifour as well), as this will impact how we design the next UI components.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/CurrentRefinements.test.tsx
```diff
@@ -0,0 +1,648 @@
...
+          <ul
+            class="ais-CurrentRefinements-list ais-CurrentRefinements-list--noRefinement"
+          />
+        </div>
+      </div>
+    `);
+  });
+});
+
+function RefinementList(props: UseRefinementListProps) {
```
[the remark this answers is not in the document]
sarahdayan: Addressed in cb5e7197.

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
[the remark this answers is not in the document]
Haroenv: good point!

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
Haroenv: I see why it could make sense, but to me an onClick is part of the ui component, not something that "framework wrappers" should care about. The contract of `item.refinements` is already defined by the connector, and I don't really see why we should stray

Haroenv (changes requested):
code all works as expected, just not sure whether the api should be `items[i].onClick` or the same as the connector. IMHO there's no need for the discrepancy between the two, but we should decide as a team

sarahdayan:
@Haroenv I've scheduled a 15mn team sync so we can make a decision on this 👍

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
francoischalifour: My take is that this should be in the UI component because it should handle accessibility for us (we don't want to write this code in all the flavors, we want consistency):

```js
if (isModifierClick(event)) {
  return;
}
event.preventDefault();
```

And this should be in a `onToggle()` callback, because _this_ is specific to InstantSearch:

```js
refine(refinement);
```

That's how I've designed the other components.

sarahdayan: I agree with moving the click modifying part in the UI component—I initially hesitated to do so. We indeed don't want to replicate this, ideally there should be a single `isModifierClick` function, not one per flavor.

And indeed we probably don't need to map `onClick` callbacks to have one per category. We can have a top-level `onClick` or `onRemove` callback within which we call `refine`. The only important thing for me is that we shouldn't have a `refine` prop on the UI component because that's an InstantSearch API, not a UI concern.

Haroenv: I don't think we can create that api easily, or will need to map back to the item, `refine` isn't global for current refinements I thought

sarahdayan: @Haroenv I think it is. There's a [single `refine` function](https://www.algolia.com/doc/api-reference/widgets/current-refinements/react-hooks/#widget-param-refine) returned by the connector, and it takes a refinement. So there can be a single callback.

I'm good with both local `onClick` functions on each items (current situation) or a single `onRemove` callback at the top level (@dhayab's proposal), there shouldn't be any differences.

sarahdayan:
Post-sync summary: with @Haroenv we suggest doing the following:

- A single `onRemove` function at the top-level (the `refine` function returned by the Hook and the one on each `CurrentRefinementsItem` are exactly the same, no need to have local ones for now).
- We use the same shape for the passed `items` in the UI component so we don't have to map them.

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
Haroenv: I'm ok with a global onRemove and items={items} :)

sarahdayan: @Haroenv @dhayab @francoischalifour Updated in 2e6607bc. There's slightly more coupling now since `onRemove` exposes a `CurrentRefinementsConnectorParamsRefinement`, but we no longer map `items`.

If we want to decouple later on, we can map and have local `onClick` functions on items (my personal preference). I think this is okay, even after we move components to `ui-components`. We can have breaking changes in `ui-components` without having breaking changes in InstantSearch.

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+
+import { cx } from './lib/cx';
+import { isModifierClick } from './lib/isModifierClick';
+
+import type {
+  CurrentRefinementsConnectorParamsItem,
+  CurrentRefinementsConnectorParamsRefinement,
+} from 'instantsearch.js/es/connectors/current-refinements/connectCurrentRefinements';
```
Haroenv, line 9: it's ok to have imported types from InstantSearch here, but once we implement InstantSearch.js ui components, we'll likely retype these, to avoid circular imports

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
```diff
@@ -0,0 +1,41 @@
...
+> &
+  UseCurrentRefinementsProps;
+
+export function CurrentRefinements(props: CurrentRefinementsProps) {
+  const {
+    includedAttributes,
+    excludedAttributes,
+    transformItems,
+    ...otherProps
+  } = props;
```
Haroenv, line 21:
```suggestion
export function CurrentRefinements({
    includedAttributes,
    excludedAttributes,
    transformItems,
    ...props
  }: CurrentRefinementsProps) {
    // and using props instead of otherProps lower
```

what do you think about avoiding naming props, as it doesn't get used?

Haroenv (approved):
a small suggestion left

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
sarahdayan: At some point we could have an `instantsearch-core` package with connectors and common types. InstantSearch.js should likely be a flavor like the others, and consume the primitives from a core package. WDYT?

packages/react-instantsearch-hooks-dom/src/widgets/CurrentRefinements.tsx
sarahdayan: Yup! I moved the destructuring in the signature in 052ec448.

packages/react-instantsearch-hooks-dom/src/ui/CurrentRefinements.tsx
Haroenv: yes, that's indeed another option we can consider when ui gets split out 👍

Haroenv (approved):

PR #3370, opened 2022-02-28
feat(hooks): implement Highlight and Snippet

Haroenv:
**Summary**

Implements the "non-widget" Highlight and Snippet widgets

**Result**

- Highlight component
- Snippet component
- Highlight underlying UI component

FX-1153


packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -0,0 +1,98 @@
...
+  value: string;
+};
+
+export type HighlightProps = React.HTMLAttributes<HTMLSpanElement> & {
+  baseClassName: string;
+  highlightedTagName?: React.ReactType;
+  nonHighlightedTagName?: React.ReactType;
+  separator?: React.ReactNode;
+  parts: HighlightedPart[];
+  partsMultiple: HighlightedPart[][];
```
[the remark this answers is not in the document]
Haroenv: actually, it could always be multiple, we just need to wrap the existing "parts" in an array, the rest of the code can be the same again, will amend tomorrow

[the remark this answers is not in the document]
Haroenv: 51f24950

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -0,0 +1,83 @@
...
+    </TagName>
+  );
+}
+
+type HighlightedPart = {
+  isHighlighted: boolean;
+  value: string;
+};
+
+export type HighlightProps = React.HTMLAttributes<HTMLSpanElement> & {
```
sarahdayan, line 40: Will `React.HTMLAttributes<HTMLSpanElement>` work with any tag name the user would provide?

packages/react-instantsearch-hooks-dom/src/widgets/Highlight.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Highlight as HighlightUiComponent } from '../ui';
+
+import type { HighlightProps as HighlightUiComponentProps } from '../ui';
+import type { BaseHit, Hit } from 'instantsearch.js';
+
+/**
+ * Make certain keys of an object optional
+ */
+type PartialKeys<TObj, TKeys extends keyof TObj> = Omit<TObj, TKeys> &
+  Partial<Pick<TObj, TKeys>>;
```
sarahdayan, line 16: We can probably already move it into a common `types/` directory to attract attention on it and avoid reimplementing it inadvertently in new PRs.

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -0,0 +1,98 @@
...
+  baseClassName: string;
+  highlightedTagName?: React.ReactType;
+  nonHighlightedTagName?: React.ReactType;
+  separator?: React.ReactNode;
+  parts: HighlightedPart[];
+  partsMultiple: HighlightedPart[][];
+};
+
+export function Highlight({
+  baseClassName,
```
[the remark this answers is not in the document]
sarahdayan: A suit-like utility will probably be more useful because we can reuse it in other components, but it's not as greppable (but no solution erases the problem so that's not a dealbreaker IMO).

Are we currently happy with our suit utility in other flavors? Do we find it useful after using it for a while, or do we find it more of a needless indirection than a help?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+            data: {
+              subdata: {
+                matchedWords: ['test'],
+                matchLevel: 'partial',
+                value: '<mark>te</mark>st',
+              },
+            },
+          },
+        }}
+        // @ts-expect-error TypeScript type does not accept string, as a way to discourage it
```
sarahdayan, line 115: 👍 Did we already deprecate this notation?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+  });
+
+  test("renders nothing when there's no match", () => {
+    const { container } = render(
+      <Highlight
+        hit={{
+          objectID: '1',
+          __position: 1,
+        }}
+        // @ts-expect-error TS doesn't allow an attribute which doesn't exist
```
sarahdayan, line 188:
```suggestion
        // @ts-expect-error TypeScript doesn't allow an attribute which doesn't exist
```

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+    expect(container).toMatchInlineSnapshot(`
+      <div>
+        <span
+          class="ais-Highlight"
+        />
+      </div>
+    `);
+  });
+
+  test('forwards className and root props', () => {
```
sarahdayan, line 202:
```suggestion
  test('forwards `className` and root props', () => {
```

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+    const { container } = render(
+      <Highlight
+        className="custom-rootclass"
+        hidden={true}
+        hit={{
+          objectID: '1',
+          __position: 1,
+        }}
+        // @ts-expect-error TS doesn't allow an attribute which doesn't exist
+        attribute="something-that-doesnt-exist"
```
sarahdayan, line 212: We don't need this in that test.

```suggestion
```

sarahdayan (commented):
👍 Neat!

I left mostly nitpicks and a few questions. When applicable, feedback applies to both `Highlight` and `Snippet`.

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { Fragment } from 'react';
+
+import { cx } from './lib/cx';
+
+type HighlightPartProps = {
+  children: React.ReactNode;
+  baseClassName: string;
+  highlightedTagName: React.ReactType;
+  nonHighlightedTagName: React.ReactType;
```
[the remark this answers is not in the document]
Haroenv: Yes, JSXElementConstructor is class or function, ReactType is also html string, which isn't allowed for hitComponent

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
Haroenv: We don't officially deprecate it. Should we, or do I change the TS type to allow any string, not just top-level keys?

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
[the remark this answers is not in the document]
Haroenv: I've always found `suit` useful, but @dhayab mentioned he preferred as much strings as possible

packages/react-instantsearch-hooks-dom/src/widgets/Snippet.tsx
```diff
@@ -0,0 +1,52 @@
...
+export function Snippet<THit extends Hit<BaseHit>>({
+  hit,
+  attribute,
+  highlightedTagName,
+  nonHighlightedTagName,
+  separator,
+  ...props
+}: SnippetProps<THit>) {
+  const property =
+    getPropertyByPath(hit._highlightResult, attribute as string) || [];
```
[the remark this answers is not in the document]
Haroenv: oops yes, last minute copy paste mistake :)

packages/react-instantsearch-hooks-dom/src/ui/index.ts
```diff
@@ -1 +1,4 @@
 export * from './ClearRefinements';
+export * from './Highlight';
+export * from './Hits';
+export * from './SearchBox';
```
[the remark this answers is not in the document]
Haroenv: snippet doesn't exist in ui

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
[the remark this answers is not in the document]
francoischalifour: I also prefer regular strings, and the [grep test](http://jamie-wong.com/2013/07/12/grep-test/).

We could pass the full class names as props, instead of just the base.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
sarahdayan: We can pass a valid attribute then, so we don't need the `@ts-expect-error`. This isn't what we're testing here, and it distracts from the actual test case.

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
sarahdayan: True, I got mixed up between types.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
Haroenv: I tried a lot to make it work in TS, but it's simply super hard to type it correct, the execution keeps timing out. See https://github.com/algolia/react-instantsearch/pull/3377 for my trial. I don't think we need to deprecate, but we can still leave TypeScript with the "simpler" type

sarahdayan: Simpler type being that it errors with strings, right? This is a fine tradeoff to me, as long as we don't document the dotted notation but the alternative instead.

FabienMotte (approved):

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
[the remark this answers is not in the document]
Haroenv: https://github.com/algolia/react-instantsearch/pull/3379

dhayab (approved):

PR #3371, opened 2022-02-28
feat(hooks): implement Pagination component

sarahdayan:
## Summary

This introduces the `Pagination` DOM widget.

The implementation slightly differs from the spec, which doesn't add a `ais-Pagination-item--page` class on numeric pagination items if they're also selected. This behavior is inconsistently implemented in every flavor, but InstantSearch.js does allow the class on every numeric pagination item, which makes sense. The spec is updated [here](https://github.com/algolia/instantsearch-specs/pull/119).

Ticket: [FX-1158](https://algolia.atlassian.net/browse/FX-1158)

## Results

- New `Pagination` widget
- New `isModifierClick` utility (tested in #3368)
- Replaced `Pagination` with the widget in example

### DIFF
--- a/packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
+++ b/packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
@@ -1,100 +1,129 @@
 import React from 'react';
 
 import { cx } from './lib/cx';
+import { isModifierClick } from './lib/isModifierClick';
 
 import type { CreateURL } from 'instantsearch.js';
 
-export type PaginationItem = {
-  label: string;
-  value: number;
-};
-
 export type PaginationProps = React.HTMLAttributes<HTMLDivElement> & {
-  items: PaginationItem[];
+  pages: number[];
   currentPage: number;
   isFirstPage: boolean;
   isLastPage: boolean;
   nbPages: number;
+  showFirst?: boolean;
+  showPrevious?: boolean;
+  showNext?: boolean;
+  showLast?: boolean;
   createURL: CreateURL<number>;
-  onClick: React.AnchorHTMLAttributes<HTMLAnchorElement>['onClick'];
+  onNavigate: (page: number) => void;
 };
 
 export function Pagination({
-  items,
+  pages,
   currentPage,
   nbPages,
   isFirstPage,
   isLastPage,
+  showFirst = true,
+  showPrevious = true,
+  showNext = true,
+  showLast = true,
   createURL,
-  onClick,
+  onNavigate,
   ...props
 }: PaginationProps) {
+  const firstPageIndex = 0;
+  const previousPageIndex = currentPage - 1;
+  const nextPageIndex = currentPage + 1;
+  const lastPageIndex = nbPages - 1;
+
   return (
     <div
       {...props}
       className={cx(
         'ais-Pagination',
-        items.length === 0 && 'ais-Pagination--noRefinement',
+        nbPages <= 1 && 'ais-Pagination--noRefinement',
         props.className
       )}
     >
       <ul className="ais-Pagination-list">
-        <PaginationItem
-          isDisabled={isFirstPage}
-          className="ais-Pagination-item--firstPage"
-          aria-label="First"
-          href={createURL(0)}
-          onClick={onClick}
-        >
-          ‹‹
-        </PaginationItem>
-        <PaginationItem
-          isDisabled={isFirstPage}
-          className="ais-Pagination-item--previousPage"
-          aria-label="Previous"
-          href={createURL(currentPage - 1)}
-          onClick={onClick}
-        >
-          ‹
-        </PaginationItem>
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
```diff
@@ -0,0 +1,168 @@
...
+    <li className={cx('ais-Pagination-item', className)}>
+      <a
+        className="ais-Pagination-link"
+        href={href}
+        onClick={(event) => {
+          if (isModifierClick(event)) {
+            return;
+          }
+
+          event.preventDefault();
```
sarahdayan, line 160: This prevents from reloading the page. Not sure how to properly test this though.

Haroenv: you can check event.defaultPrevented or similar, but not sure if that is worth testing. If the click is not prevented, the js stops right (or does the testing utility not implement that?)

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,1110 @@
...
+    userEvent.click(pageOneLink as HTMLAnchorElement, { button: 1 });
+    userEvent.click(pageOneLink as HTMLAnchorElement, { altKey: true });
+    userEvent.click(pageOneLink as HTMLAnchorElement, { ctrlKey: true });
+    userEvent.click(pageOneLink as HTMLAnchorElement, { metaKey: true });
+    userEvent.click(pageOneLink as HTMLAnchorElement, { shiftKey: true });
+
+    expect(props.onNavigate).not.toHaveBeenCalled();
+  });
+
+  test('hides the "First" item', () => {
```
Haroenv, line 627:
```suggestion
  test('hides the "First" item when showFirst is false', () => {
```

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                ››
+              </a>
+            </li>
+          </ul>
+        </div>
+      </div>
+    `);
+  });
+
+  test('hides the "Previous" item', () => {
```
Haroenv, line 705:
```suggestion
  test('hides the "Previous" item when showPrevious is false', () => {
```

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                ››
+              </a>
+            </li>
+          </ul>
+        </div>
+      </div>
+    `);
+  });
+
+  test('hides the "Next" item', () => {
```
Haroenv, line 783:
```suggestion
  test('hides the "Next" item when showNext is false', () => {
```

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                ››
+              </a>
+            </li>
+          </ul>
+        </div>
+      </div>
+    `);
+  });
+
+  test('hides the "Last" item', () => {
```
Haroenv, line 858:
```suggestion
  test('hides the "Last" item when showNext is false', () => {
```

Haroenv (commented):
👍 looks great! small questions

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Pagination.test.tsx
sarahdayan: Addressed in 22700e76.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Pagination.test.tsx
sarahdayan: Addressed in 22700e76.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Pagination.test.tsx
sarahdayan: Addressed in 22700e76.

packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
sarahdayan: If the click isn't prevented, the JS is never executed because the `<a>` behavior takes over (it navigates) so indeed this is implicitly tested by the `expect(props.onNavigate).toHaveBeenCalledTimes(n)` in other tests 👍

Haroenv (approved):

PR #3373, opened 2022-03-02
feat(hooks): implement SortBy component

sarahdayan:
## Summary

This introduces the `SortBy` DOM widget.

Ticket: [FX-1164](https://algolia.atlassian.net/browse/FX-1164)

## Results

- New `SortBy` widget
- Replaced `SortBy` with the widget in example

### DIFF
--- a/packages/react-instantsearch-hooks-dom/src/ui/SortBy.tsx
+++ b/packages/react-instantsearch-hooks-dom/src/ui/SortBy.tsx
@@ -2,14 +2,13 @@ import React from 'react';
 
 import { cx } from './lib/cx';
 
-import type { PartialKeys } from '../types';
 import type { UseSortByProps } from 'react-instantsearch-hooks';
 
 export type SortByProps = Omit<
   React.HTMLAttributes<HTMLDivElement>,
   'onChange'
 > &
-  PartialKeys<UseSortByProps, 'items'> &
+  Pick<UseSortByProps, 'items'> &
   Pick<React.SelectHTMLAttributes<HTMLSelectElement>, 'value'> & {
     onChange?(value: string): void;
   };
--- a/packages/react-instantsearch-hooks-dom/src/ui/SortBy.tsx
+++ b/packages/react-instantsearch-hooks-dom/src/ui/SortBy.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+
+import { cx } from './lib/cx';
+
+import type { PartialKeys } from '../types';
+import type { UseSortByProps } from 'react-instantsearch-hooks';
+
+export type SortByProps = Omit<
+  React.HTMLAttributes<HTMLDivElement>,
+  'onChange'
+> &
+  PartialKeys<UseSortByProps, 'items'> &
+  Pick<React.SelectHTMLAttributes<HTMLSelectElement>, 'value'> & {
+    onChange?(value: string): void;
+  };
+
+export function SortBy({
+  items = [],
+  value,
+  onChange = () => {},
+  ...props
+}: SortByProps) {
+  return (
+    <div {...props} className={cx('ais-SortBy', props.className)}>
+      <select
+        className="ais-SortBy-select"
+        onChange={(event) => onChange(event.target.value)}
+        value={value}
+      >
+        {items.map((item) => (
+          <option
+            className="ais-SortBy-option"
+            key={item.value}
+            value={item.value}
+          >
+            {item.label}
+          </option>
+        ))}
+      </select>
+    </div>
+  );
+}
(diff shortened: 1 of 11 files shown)


packages/react-instantsearch-hooks-dom/src/ui/SortBy.tsx
```diff
@@ -0,0 +1,42 @@
...
+  React.HTMLAttributes<HTMLDivElement>,
+  'onChange'
+> &
+  PartialKeys<UseSortByProps, 'items'> &
+  Pick<React.SelectHTMLAttributes<HTMLSelectElement>, 'value'> & {
+    onChange?(value: string): void;
+  };
+
+export function SortBy({
+  items = [],
```
Haroenv, line 18: should items really be optional?

packages/react-instantsearch-hooks-dom/src/ui/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './ClearRefinements';
```
Haroenv, line 1: I don't really have an opinion on this, but we've been flip-flopping in the PRs, what are the pros/cons?

Pro:
- closer to what a real package will look like

Cons:
- higher chance of circular imports
- need to be careful to mark as tree-shakeable

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/SortBy.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+    );
+
+    await wait(0);
+
+    userEvent.selectOptions(
+      document.querySelector('.ais-SortBy-select') as HTMLSelectElement,
+      getByRole('option', { name: 'Price (asc)' })
+    );
+
+    await wait(0);
```
Haroenv, line 132: we want to see that the search results also come from that other index, otherwise you could have a select without refine still passing this test

Haroenv (commented):
👍

Haroenv:
looks like something with the types loses specificity in the index.test.tsx, maybe try putting the component instantiation inside the switch instead of just the props?

packages/react-instantsearch-hooks-dom/src/ui/index.ts
dhayab: Is there a possibility for something other than a 1:1 direct import mapping from widget → UI (except Highlight/Snippet → Highlight)? If not, I don't think of an issue with having an index right now.

sarahdayan:
@Haroenv I'm afraid that's a little too dynamic for TypeScript, and we'll need to either maintain code to make it understand the mapping better (defeating the purpose of importing all widgets from the barrel) or just disable type checking there.

Test will fail if we pass wrong parameters anyway so I'd say, let's `@ts-ignore` here. wdyt?

Haroenv:
I fixed it in 1e1ef033 @sarahdayan, with the same cast still needed (as Object.entries isn't type-safe for some dumb reason https://github.com/microsoft/TypeScript/issues/35101)

packages/react-instantsearch-hooks-dom/src/ui/SortBy.tsx
sarahdayan: Addressed in f0084987

packages/react-instantsearch-hooks-dom/src/ui/index.ts
sarahdayan: Not much of an opinion either, but there's no _need_ for a barrel at this level so I'm fine removing it and add it only if/when we feel the pain.

Addressed in afeb265e.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/SortBy.test.tsx
sarahdayan: Addressed in 810473f3.

packages/react-instantsearch-hooks-dom/src/ui/SortBy.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { cx } from './lib/cx';
+
+import type { PartialKeys } from '../types';
+import type { UseSortByProps } from 'react-instantsearch-hooks';
+
+export type SortByProps = Omit<
+  React.HTMLAttributes<HTMLDivElement>,
+  'onChange'
+> &
+  PartialKeys<UseSortByProps, 'items'> &
```
Haroenv, line 12:
```suggestion
 UseSortByProps &
```

(then prettier)

sarahdayan: This would be `Pick` instead (we don't pass `transformItems` to the UI component) but good catch 👍

Addressed in b3840aa8.

Haroenv (approved):
todo bem

PR #3366, opened 2022-02-22
feat(hooks): implement InfiniteHits component

Haroenv:
**Summary**

FX-1155

Implements the InfiniteHits component

**Result**

- Implements the component
- use in the example
- to do: find out why tests don't pass

PR #3378, opened 2022-03-03
fix(infiniteHits): mark results as artificial

Haroenv:
based on https://github.com/algolia/instantsearch.js/pull/5023

fixes the tests for #3366


francoischalifour (commented):
It would make sense to ensure that the code paths in `useConnector()` that rely on `createSearchResults()` do have this `__isArtificial` flag imo. Once this is tested, we can consider this as an internal API we can rely on. Wdyt?

Haroenv:
Is the test in f8e27473 enough to rely on, or were you thinking of something else as well @francoischalifour ?

PR #3379, opened 2022-03-04
feat(ui): introduce "classNames" prop for highlight

Haroenv:
**Summary**

This allows us to reuse the same ui component for multiple widgets

**Result**

- ui component accepts an array of classes for every "key", and is required
- widget accepts a string for every "key", and isn't required, these are only additional

### DIFF
--- a/packages/react-instantsearch-hooks-dom/src/types/index.ts
+++ b/packages/react-instantsearch-hooks-dom/src/types/index.ts
@@ -7,7 +7,7 @@ export type PartialKeys<TObj, TKeys extends keyof TObj> = Omit<TObj, TKeys> &
   Partial<Pick<TObj, TKeys>>;
 
 /**
- * Map from CSS Classes used internally by `ui` to the classes passed by users
+ * Map from CSS classes used in UI components to the `classNames` API exposed to users
  */
 export type ClassNames<
   TProps extends { classNames: Record<string, CSSClass> }
--- a/packages/react-instantsearch-hooks-dom/src/types/index.ts
+++ b/packages/react-instantsearch-hooks-dom/src/types/index.ts
@@ -9,6 +9,6 @@ export type PartialKeys<TObj, TKeys extends keyof TObj> = Omit<TObj, TKeys> &
 /**
  * Map from CSS Classes used internally by `ui` to the classes passed by users
  */
-export type CSSClasses<
-  TProps extends { cssClasses: Record<string, CSSClass> }
-> = { cssClasses?: { [className in keyof TProps['cssClasses']]?: string } };
+export type ClassNames<
+  TProps extends { classNames: Record<string, CSSClass> }
+> = { classNames?: { [className in keyof TProps['classNames']]?: string } };
--- a/packages/react-instantsearch-hooks-dom/src/types/index.ts
+++ b/packages/react-instantsearch-hooks-dom/src/types/index.ts
@@ -1,5 +1,14 @@
+import { CSSClass } from '../ui/lib/cx';
+
 /**
  * Make certain keys of an object optional
  */
 export type PartialKeys<TObj, TKeys extends keyof TObj> = Omit<TObj, TKeys> &
   Partial<Pick<TObj, TKeys>>;
+
+/**
+ * Map from CSS Classes used internally by `ui` to the classes passed by users
+ */
+export type CSSClasses<
+  TProps extends { cssClasses: Record<string, CSSClass> }
+> = { cssClasses?: { [className in keyof TProps['cssClasses']]?: string } };
--- a/packages/react-instantsearch-hooks-dom/src/types/index.ts
+++ b/packages/react-instantsearch-hooks-dom/src/types/index.ts
@@ -1,4 +1,4 @@
-import { CSSClass } from '../ui/lib/cx';
+import type { CSSClass } from '../ui/lib/cx';
 
 /**
  * Make certain keys of an object optional
--- a/packages/react-instantsearch-hooks-dom/src/types/index.ts
+++ b/packages/react-instantsearch-hooks-dom/src/types/index.ts
@@ -1,5 +1,3 @@
-import type { CSSClass } from '../ui/lib/cx';
-
 /**
  * Make certain keys of an object optional
  */
@@ -9,6 +7,5 @@ export type PartialKeys<TObj, TKeys extends keyof TObj> = Omit<TObj, TKeys> &
 /**
  * Map from CSS classes used in UI components to the `classNames` API exposed to users
  */
-export type ClassNames<
-  TProps extends { classNames: Record<string, CSSClass> }
-> = { classNames?: { [className in keyof TProps['classNames']]?: string } };
+export type ClassNames<TProps extends { classNames: Record<string, string> }> =
+  Omit<TProps, 'classNames'> & { classNames?: Partial<TProps['classNames']> };
--- a/packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
+++ b/packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
@@ -2,8 +2,6 @@ import React, { Fragment } from 'react';
 
 import { cx } from './lib/cx';
 
-import type { CSSClass } from './lib/cx';
-
 type HighlightPartProps = {
   children: React.ReactNode;
   classNames: HighlightClassNames;
@@ -24,9 +22,7 @@ function HighlightPart({
   return (
     <TagName
       className={
-        isHighlighted
-          ? cx(classNames.highlighted)
-          : cx(classNames.nonHighlighted)
+        isHighlighted ? classNames.highlighted : classNames.nonHighlighted
       }
     >
       {children}
@@ -43,19 +39,19 @@ export type HighlightClassNames = {
   /**
    * Class names to apply to the root element
    */
-  root: CSSClass;
+  root: string;
   /**
    * Class names to apply to the highlighted parts
    */
-  highlighted: CSSClass;
+  highlighted: string;
   /**
    * Class names to apply to the non-highlighted parts
    */
-  nonHighlighted: CSSClass;
+  nonHighlighted: string;
   /**
    * Class names to apply to the separator between highlighted parts
    */
-  separator: CSSClass;
+  separator: string;
 };
 
 export type HighlightProps = React.HTMLAttributes<HTMLSpanElement> & {
@@ -95,7 +91,7 @@ export function Highlight({
             ))}
 
             {!isLastPart && (
-              <span className={cx(classNames.separator)}>{separator}</span>
+              <span className={classNames.separator}>{separator}</span>
             )}
           </Fragment>
         );
--- a/packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
+++ b/packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
@@ -6,14 +6,14 @@ import type { CSSClass } from './lib/cx';
 
 type HighlightPartProps = {
   children: React.ReactNode;
-  cssClasses: HighlightCSSClasses;
+  classNames: HighlightClassNames;
   highlightedTagName: React.ReactType;
   nonHighlightedTagName: React.ReactType;
   isHighlighted: boolean;
 };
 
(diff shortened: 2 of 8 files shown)


Haroenv:
once this PR is merged, we can introduce it for the other ui components & widgets

francoischalifour:
I think the term `classNames` is more used than `cssClasses`?

dhayab (commented):
That looks great! Just a small suggestion below. Also not sure about the naming, as I think `classNames` works better in React, but I understand `cssClasses` is already in place in IS.js, so I'm okay with both.

francoischalifour:
For the record we use [`classNames`](https://www.algolia.com/doc/ui-libraries/autocomplete/api-reference/autocomplete-js/autocomplete/#param-classnames) in Autocomplete.

sarahdayan (commented):
The naming doesn't matter as much as in InstantSearch.js or Autocomplete here since this is an "internal" API—it will be exposed once we move the components to the separate repo, but this isn't a first-class InstantSearch API.

I'd still go with `classNames` instead of `cssClasses` because we're in React, and this feels more idiomatic. If this was a top-level API at the widget level, we'd probably go for `classNames`.

packages/react-instantsearch-hooks-dom/src/widgets/Highlight.tsx
```diff
@@ -37,7 +39,15 @@ export function Highlight<THit extends Hit<BaseHit>>({
...
-      baseClassName="ais-Highlight"
+      cssClasses={{
+        root: ['ais-Highlight', cssClasses.root],
+        highlighted: ['ais-Highlight-highlighted', cssClasses.highlighted],
+        nonHighlighted: [
+          'ais-Highlight-nonHighlighted',
+          cssClasses.nonHighlighted,
+        ],
+        separator: ['ais-Highlight-separator', cssClasses.separator],
+      }}
```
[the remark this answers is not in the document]
Haroenv: I experimented with that, but concluded it was easier to rely on cx, which already accepted an array of items

packages/react-instantsearch-hooks-dom/src/ui/lib/cx.ts
```diff
@@ -1,5 +1,9 @@
...
-  ...classNames: Array<string | number | boolean | undefined | null>
-) {
-  return classNames.filter(Boolean).join(' ');
+type Root = string | number | boolean | undefined | null;
+export type CSSClass = Root | Root[];
+
+export function cx(...classNames: CSSClass[]) {
+  return classNames
+    .reduce<Root[]>((acc, curr) => acc.concat(curr), [])
+    .filter(Boolean)
```
sarahdayan, line 7: If we keep this (see previous comment) could we do the filtering in `reduce` to avoid the two passes? And test it?

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Highlight.test.tsx
```diff
@@ -115,7 +133,37 @@ describe('Highlight', () => {
...
+        />
+      </div>
+    `);
+  });
+
+  test('accepts array of classes for classNames', () => {
+    const { container } = render(
+      <Highlight
+        classNames={{
+          root: ['ROOT', 'EXTRA', false],
```
sarahdayan, line 162: I can see why this is convenient, but couldn't this be done by calling `cx` on the passed classes instead?

```diff
-<Highlight classNames={{ root: ['ROOT', 'EXTRA', false] }} />
+<Highlight classNames={{ root: cx('ROOT', 'EXTRA', false) }} />
```

packages/react-instantsearch-hooks-dom/src/types/index.ts
```diff
@@ -1,5 +1,14 @@
+import type { CSSClass } from '../ui/lib/cx';
+
 /**
  * Make certain keys of an object optional
  */
 export type PartialKeys<TObj, TKeys extends keyof TObj> = Omit<TObj, TKeys> &
   Partial<Pick<TObj, TKeys>>;
+
+/**
+ * Map from CSS Classes used internally by `ui` to the classes passed by users
```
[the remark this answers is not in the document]
Haroenv: it's not a map of, it's a type which maps from, that's not the same right?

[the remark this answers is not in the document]
sarahdayan: Yes you're right, I was a little fast here—ultimately used to _generate class maps_ but what the type does is map from.

Is this better?

```suggestion
 * Map from CSS classes used in UI components to the `classNames` API exposed to users
```

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -37,24 +39,43 @@ type HighlightedPart = {
   value: string;
 };
 
+export type HighlightClassNames = {
+  /**
+   * Class names to apply to the root element
+   */
+  root: CSSClass;
```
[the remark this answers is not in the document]
dhayab: I don't mind having both. Here it serves the purpose of listing exhaustively all the different type of elements that can be styled.

packages/react-instantsearch-hooks-dom/src/ui/lib/cx.ts
Haroenv: If you'd pass in reduce immediately, you'd need something like `acc.concat(Array.isArray(curr) ? curr.filter(Boolean) : [curr].filter(Boolean))` or similar, that doesn't seem better to me

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Highlight.test.tsx
Haroenv: I guess calling cx on two levels is fairly equivalent to flattening inside cx. Not sure what we prefer as the team?

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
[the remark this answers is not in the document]
sarahdayan: That's a good point @dhayab, it enhances the DX and centralizes where classes are passed—either you override default classes and you define everything in `classNames`, or you just want to add your own root class and pass `className`.

packages/react-instantsearch-hooks-dom/src/ui/lib/cx.ts
sarahdayan: Fair, and the performance implications are likely nonexistent.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Highlight.test.tsx
sarahdayan: I'm good either way. But if we accept arrays, we need to add a test in the `cx` test suite.

sarahdayan (approved):

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -73,7 +94,7 @@ export function Highlight({
             ))}
 
             {!isLastPart && (
-              <span className={`${baseClassName}-separator`}>{separator}</span>
+              <span className={cx(classNames.separator)}>{separator}</span>
```
francoischalifour, line 97: Why do we call `cx()` here?
```suggestion
              <span className={classNames.separator}>{separator}</span>
```

packages/react-instantsearch-hooks-dom/src/types/index.ts
```diff
@@ -1,5 +1,14 @@
...
  * Make certain keys of an object optional
  */
 export type PartialKeys<TObj, TKeys extends keyof TObj> = Omit<TObj, TKeys> &
   Partial<Pick<TObj, TKeys>>;
+
+/**
+ * Map from CSS classes used in UI components to the `classNames` API exposed to users
+ */
+export type ClassNames<
+  TProps extends { classNames: Record<string, CSSClass> }
```
francoischalifour, line 13: We don't want to support all variants of `CSSClass` in the public API. We should just type it as `string`.

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -37,24 +39,43 @@ type HighlightedPart = {
...
   parts: HighlightedPart[][];
 };
 
 export function Highlight({
-  baseClassName,
   parts,
   highlightedTagName = 'mark',
   nonHighlightedTagName = 'span',
   separator = ', ',
+  classNames,
```
francoischalifour, line 74: This component doesn't support the original `className` prop, as opposed to all the other components. I think we should remove `classNames.root` and replace it with the `className` prop then. Wdyt?

packages/react-instantsearch-hooks-dom/src/types/index.ts
Haroenv: yes, the output of this type is in the line below and only accepts string :)

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
Haroenv: This is because each key in classNames was a string, number etc, but on reflection, that doesn't actually make sense, will change it to string

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
Haroenv: see also https://github.com/algolia/react-instantsearch/pull/3379#discussion_r825977066

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Highlight.test.tsx
```diff
@@ -115,7 +133,37 @@ describe('Highlight', () => {
...
+      <div>
+        <span
+          aria-hidden="true"
+          class="ROOT custom-root"
+        />
+      </div>
+    `);
+  });
+
+  test('accepts array of classes for classNames', () => {
```
francoischalifour, line 158: This test looks outdated since we don't support array of classes anymore.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Highlight.test.tsx
```diff
@@ -222,6 +222,29 @@ describe('Highlight', () => {
     `);
   });
 
+  test('forwards `classNames`', () => {
```
francoischalifour, line 225: Can we add a test with both `className` and `classNames.root`?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Snippet.test.tsx
```diff
@@ -219,6 +218,29 @@ describe('Snippet', () => {
     `);
   });
 
+  test('forwards `classNames`', () => {
```
francoischalifour, line 221: Same here?

packages/react-instantsearch-hooks-dom/src/ui/Highlight.tsx
```diff
@@ -37,24 +35,44 @@ type HighlightedPart = {
...
   highlightedTagName = 'mark',
   nonHighlightedTagName = 'span',
   separator = ', ',
+  className,
+  classNames,
   ...props
 }: HighlightProps) {
   return (
-    <span {...props} className={cx(baseClassName, props.className)}>
+    <span {...props} className={cx(classNames.root, className)}>
```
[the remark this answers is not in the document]
Haroenv: In the others we add our own class name first too 🤔 I'm not sure what's more logical though, the static part being first also makes sense. See eg https://github.com/algolia/react-instantsearch/blob/70e7844310b507cbcda68fb5137903dfb14523b8/packages/react-instantsearch-hooks-dom/src/ui/ClearRefinements.tsx#L20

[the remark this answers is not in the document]
sarahdayan: I'd keep the user's class names at the end, as it's additive to our own. Ultimately it really doesn't matter.

dhayab (approved):

PR #3381, opened 2022-03-13
feat(hooks): implement RefinementList component

sarahdayan:
## Summary

This introduces the `RefinementList` DOM widget.

Ticket: [FX-1163](https://algolia.atlassian.net/browse/FX-1163)
Depends on: https://github.com/algolia/instantsearch.js/pull/5024

## Implementation details

Regarding the `RefinementList` UI component, I decided to stay minimal for now and avoid early abstractions. Some widgets share a **Show more** button, but they don't share much else, so it doesn't seem necessary to "force" a common abstraction (at least for now).

The picked design exposes "slots" to provide the different parts of the component. This avoid having to pass a ton of props, and keeps the UI component generic. The downside is that we have some of the UI logic in the widget, which must be reproduced in every flavor. **Still undecided which one is best, let's discuss it.**

## Results

- New `RefinementList` widget
- Replaced `RefinementList` with the widget in example


packages/react-instantsearch-hooks/package.json
```diff
@@ -49,7 +49,7 @@
     "@babel/runtime": "^7.1.2",
     "algoliasearch-helper": "^3.7.0",
     "dequal": "^2.0.0",
-    "instantsearch.js": "^4.39.0"
+    "instantsearch.js": "https://pkg.csb.dev/algolia/instantsearch.js/commit/70fb1c6d/instantsearch.js"
```
sarahdayan, line 52: This will be updated once https://github.com/algolia/instantsearch.js/pull/5024 is released.

packages/react-instantsearch-hooks-dom/src/widgets/RefinementList.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import type { RefinementListProps as RefinementListUiComponentProps } from '../ui/RefinementList';
+import type { RefinementListItem } from 'instantsearch.js/es/connectors/refinement-list/connectRefinementList';
+import type { RefinementListWidgetParams } from 'instantsearch.js/es/widgets/refinement-list/refinement-list';
+import type { UseRefinementListProps } from 'react-instantsearch-hooks';
+
+export type RefinementListProps = Omit<
+  RefinementListUiComponentProps,
+  'items' | 'onRefine' | 'searchBox' | 'showMoreButton'
+> &
```
dhayab, line 17: I don't see a use for the sole leftover prop `noResults` here. We could replace all this with `React.HTMLAttributes<HTMLDivElement>` instead?

packages/react-instantsearch-hooks-dom/src/ui/ShowMoreButton.tsx
```diff
@@ -0,0 +1,14 @@
...
+type ShowMoreButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+  isShowingMore: boolean;
+};
+
+export function ShowMoreButton({
+  isShowingMore,
+  ...props
+}: ShowMoreButtonProps) {
+  return (
+    <button {...props}>{isShowingMore ? 'Show less' : 'Show more'}</button>
```
dhayab, line 12: How could we add localization for this button? In RIS, for example, there's a [`translations`](https://www.algolia.com/doc/api-reference/widgets/refinement-list/react/#widget-param-translations) prop that can help with this.

packages/react-instantsearch-hooks-dom/src/widgets/RefinementList.tsx
sarahdayan: Good catch, I forgot to omit it. Addressed in 17f5cdbc.

packages/react-instantsearch-hooks-dom/src/ui/ShowMoreButton.tsx
sarahdayan: We discussed it in another PR with @Haroenv and there's [this RFC](https://github.com/algolia/instantsearch-rfcs/blob/5414147664e12457136f50bb9511c9081705929f/accepted/translations.md) which we could adopt.

dhayab: I was thinking of something "simpler" in the meantime, like the `resetLabel` prop we have for `<ClearRefinements>` for example. It's fine for me if that's in another PR.

dhayab (approved):
Looks good to me, after bumping `instantsearch.js`.

packages/react-instantsearch-hooks-dom/src/ui/ShowMoreButton.tsx
sarahdayan: It's not only this prop, we have five of them.

What we can do is leave this PR as is for now (we did merge some other widgets without translations) and add a new task to introduce the Translations API, then implement it in all widgets. WDYT?

dhayab: Works for me!

packages/react-instantsearch-hooks-dom/src/ui/RefinementList.tsx
```diff
@@ -0,0 +1,63 @@
...
+                  className="ais-RefinementList-checkbox"
+                  type="checkbox"
+                  value={item.value}
+                  onChange={() => {
+                    onRefine(item);
+                  }}
+                />
+                <span
+                  className="ais-RefinementList-labelText"
+                  dangerouslySetInnerHTML={{ __html: item.highlighted! }}
```
francoischalifour, line 52: Could we use the `<Highlight>` component here?

sarahdayan: The `<Highlight>` component is still in review, so I can do this in another PR, once it's merged.

dhayab: It won't work as is, our `<Highlight>` component will only get its input from `.__highlightResult` and won't pass the value from `.highlighted`.

francoischalifour: We can always transform the hits structure before passing it to the component. We're doing this in some other libs.

dhayab: I agree, I find it better than handling this specific case in the highlighter.

sarahdayan: @dhayab It will work if we use the `<Highlight>` UI component (not the widget) which isn't coupled to this data shape.

francoischalifour:
@sarahdayan I'll start reviewing InstantSearch-related PRs from tomorrow. Feel free to merge if this is urgent.

sarahdayan:
@francoischalifour This can definitely wait for tomorrow. Thanks!

packages/react-instantsearch-hooks-dom/src/widgets/RefinementList.tsx
```diff
@@ -0,0 +1,124 @@
...
+    {
+      attribute,
+      operator,
+      limit,
+      showMore,
+      showMoreLimit,
+      sortBy,
+      escapeFacetValues,
+      transformItems,
+    },
```
Haroenv, line 52: we can change this in a different PR, but I wonder if we should pass all props to useRefinementList too. We know they won't overlap, as it's very unlikely for dom props, but it would mean a new parameter could be passed immediately, without code change

Haroenv (commented):
no big changes suggested, but would make more sense to merge this with highlight already there?

packages/react-instantsearch-hooks-dom/src/ui/RefinementList.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+
+import { cx } from './lib/cx';
+
+import type { RefinementListItem } from 'instantsearch.js/es/connectors/refinement-list/connectRefinementList';
+
+export type RefinementListProps = React.HTMLAttributes<HTMLDivElement> & {
+  items: RefinementListItem[];
+  onRefine(item: RefinementListItem): void;
+  searchBox?: React.ReactNode;
```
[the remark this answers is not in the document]
sarahdayan: No it's just a node, just like `children`. That's the "slot-based" approach I was describing in the description; instead of having the search box in the UI component directly and pass a ton of props I'm composing by passing the full thing down from the widget. Same goes for `noResults` and `showMoreButton`.

Not sure this is the right approach yet but the alternative meant passing down a ton of props, this felt cleaner.

packages/react-instantsearch-hooks-dom/src/widgets/RefinementList.tsx
sarahdayan: We're inlining the DOM props onto the UI component, so we need to separate DOM props from widget props anyway.

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -27,7 +27,7 @@ export function getServerState(
   const notifyServer: InstantSearchServerContextApi['notifyServer'] = ({
     search,
   }) => {
-    searchRef.current = search;
+    searchRef.current = search as InstantSearch;
```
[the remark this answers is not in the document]
sarahdayan: Good catch! Addressed in 85b6c6d6.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
```diff
@@ -0,0 +1,3458 @@
...
+      '.ais-RefinementList-checkbox'
+    ) as HTMLInputElement;
+
+    expect(firstCheckbox).not.toBeChecked();
+
+    userEvent.click(firstCheckbox);
+
+    await wait(0);
+
+    expect(firstCheckbox).toBeChecked();
```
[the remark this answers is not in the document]
sarahdayan: Good idea. This assertion would actually fail because there has been one call on load already.

I've added two assertions and a comment to reflect it d3d41051.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/RefinementList.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  });
+
+  test('forwards a custom class name to the root element', () => {
+    const props = createProps({ className: 'MyRefinementList' });
+    const { container } = render(<RefinementList {...props} />);
+
+    expect(document.querySelector('.ais-RefinementList')).toHaveClass(
+      'MyRefinementList'
+    );
+    expect(container).toMatchInlineSnapshot(`
```
francoischalifour, line 474: I don't think we need this snapshot.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/RefinementList.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+
+  test('forwards `div` props to the root element', () => {
+    const props = createProps({ title: 'Some custom title' });
+    const { container } = render(<RefinementList {...props} />);
+
+    expect(document.querySelector('.ais-RefinementList')).toHaveAttribute(
+      'title',
+      'Some custom title'
+    );
+    expect(container).toMatchInlineSnapshot(`
```
francoischalifour, line 543: I don't think we need this snapshot.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
```diff
@@ -0,0 +1,3462 @@
...
+      </InstantSearchHooksTestWrapper>
+    );
+
+    await wait(0);
+
+    const [checkbox1, checkbox2] = [
+      ...(container.querySelectorAll(
+        '.ais-RefinementList-checkbox'
+      ) as NodeListOf<HTMLInputElement>),
+    ].slice(0, 2);
```
francoischalifour, line 405: Could we use `getByRole()` with `name` for this?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
```diff
@@ -0,0 +1,3462 @@
...
+          attribute="brand"
+          className="MyRefinementList"
+          title="Some custom title"
+        />
+      </InstantSearchHooksTestWrapper>
+    );
+
+    await wait(0);
+
+    expect(container).toMatchInlineSnapshot(`
```
francoischalifour, line 841: I don't think we need this snapshot. We can use `toHaveAttribute()` on `container.firstChild`.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
```diff
@@ -0,0 +1,3462 @@
...
+  describe('searching', () => {
+    test('displays a search box', async () => {
+      const client = createSearchClient({ search, searchForFacetValues });
+      const { container } = render(
+        <InstantSearchHooksTestWrapper searchClient={client}>
+          <RefinementList
+            attribute="brand"
+            searchable={true}
+            searchablePlaceholder="Search brands"
+          />
```
francoischalifour, line 1268: We may want to use `limit` to not bloat the whole file with snapshots.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
```diff
@@ -0,0 +1,3462 @@
...
+              </li>
+            </ul>
+          </div>
+        </div>
+      `);
+
+      userEvent.type(searchInput, 'app');
+
+      await wait(0);
+
```
francoischalifour, line 1627: We can use `toHaveBeenCalledTimes()` here.

francoischalifour (commented):
Nice. There's too many unnecessary snapshots in the test file that makes it hard to read, which defeats their purpose. We can remove some or add the `limit` prop to shorten them.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
sarahdayan: We could, but this way we also ensure the elements with the right class names are present and they're the ones triggering a change.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
sarahdayan: Addressed in [445448c](https://github.com/algolia/react-instantsearch/pull/3381/commits/445448cb0cadfd537e1e71a6f7074db1e9789439).

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RefinementList.test.tsx
sarahdayan: I went with more focused tests on just the list of returned facets (see [445448c](https://github.com/algolia/react-instantsearch/pull/3381/commits/445448cb0cadfd537e1e71a6f7074db1e9789439))

packages/react-instantsearch-hooks-dom/src/ui/__tests__/RefinementList.test.tsx
sarahdayan: Addressed in [445448c](https://github.com/algolia/react-instantsearch/pull/3381/commits/445448cb0cadfd537e1e71a6f7074db1e9789439)

packages/react-instantsearch-hooks-dom/src/ui/RefinementList.tsx
sarahdayan: @Haroenv @dhayab Addressed in [fe10a4a](https://github.com/algolia/react-instantsearch/pull/3381/commits/fe10a4ac92bc68ba79cfca375ea520e239ab0833).

Haroenv (approved):

packages/react-instantsearch-hooks-dom/src/ui/RefinementList.tsx
```diff
@@ -0,0 +1,74 @@
...
+                  checked={item.isRefined}
+                  className="ais-RefinementList-checkbox"
+                  type="checkbox"
+                  value={item.value}
+                  onChange={() => {
+                    onRefine(item);
+                  }}
+                />
+                <span className="ais-RefinementList-labelText">
+                  <Highlight
```
francoischalifour, line 53: This means that we're always rendering highlighting DOM for the RefinementList, even when it's not searchable or that there's no query. Can we change this to support our IS.css spec?

sarahdayan: Good catch! There's not much in the [spec](https://instantsearch-css.netlify.app/widgets/refinement-list/) regarding the highlighting of searchable refinements, so I looked at the InstantSearch.js output. The only difference is that InstantSearch.js doesn't add classes to the `<mark>` elements and doesn't add tags around non-highlighted parts.

I think we should go with the default behavior of the `<Highlight>` component here, as this is more consistent with the other highlighted strings in `<Hits>`. By default, highlighted parts will have the same styles instead of showing up with the system background color for `<mark>`.

See 43e9451a.

packages/react-instantsearch-hooks-dom/src/ui/RefinementList.tsx
```diff
@@ -50,17 +52,21 @@ export function RefinementList({
...
-                      getHighlightedParts(unescape(item.highlighted || '')),
-                    ]}
-                    classNames={{
-                      root: 'ais-Highlight',
-                      highlighted: 'ais-Highlight-highlighted',
-                      nonHighlighted: 'ais-Highlight-nonHighlighted',
-                      separator: 'ais-Highlight-separator',
-                    }}
-                  />
+                  {searchBox && query.length > 0 ? (
```
francoischalifour, line 55: We just need to check for a query to allow highlighting, right?

```suggestion
                  {query.length > 0 ? (
```

sarahdayan: Addressed a07c6ede

francoischalifour (approved):

PR #3382, opened 2022-03-14
feat(hooks): implement `Breadcrumb` component

dhayab:
# Summary

This introduces the `Breadcrumb` DOM widget.

Ticket: [FX-1149](https://algolia.atlassian.net/browse/FX-1149)

# Implementation details

For the widget, I initially wanted to have test cases related to `rootPath` and `separator`, but could not find a way to set the state to have these taken into account, and other flavours don't seem to have tests for those as well, so I left them out. Wdyt?

For the UI component, I opted to have the root label (`Home`) and the separator label (`>`) as slot components, to give more flexibility to the user, but we can always change it back to label props if necessary.

# Results

- New `Breadcrumb` widget
- Replaced `Breadcrumb` with the widget in example


packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,375 @@
...
+            class="ais-Breadcrumb-list"
+          >
+            <li
+              class="ais-Breadcrumb-item"
+            >
+              <a
+                class="ais-Breadcrumb-link"
+                href="#"
+              >
+                Home
```
Haroenv, line 200: notable that the "home" item doesn't change, as it's not part of the items. Not really a problem though

Haroenv (approved):

packages/react-instantsearch-hooks-dom/src/widgets/Breadcrumb.tsx
```diff
@@ -0,0 +1,40 @@
...
+    },
+    { $$widgetType: 'ais.breadcrumb' }
+  );
+
+  return (
+    <BreadcrumbUiComponent
+      {...props}
+      createURL={createURL}
+      items={items}
+      onNavigate={refine}
```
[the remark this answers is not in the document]
dhayab: I do have a slight preference for more semantic event handler prop names, but either way, we should have consistency in all our UI components.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,276 @@
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+
+import { Breadcrumb } from '../Breadcrumb';
+
+import type { BreadcrumbProps } from '../Breadcrumb';
+
+describe('Breadcrumb UI', () => {
```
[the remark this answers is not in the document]
Haroenv: you can use `p` to filter on test files

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
```diff
@@ -0,0 +1,92 @@
...
+  rootComponent?: React.ReactNode;
+  separatorComponent?: React.ReactNode;
+};
+
+export function Breadcrumb({
+  items = [],
+  createURL,
+  onNavigate,
+  rootComponent: RootComponent = 'Home',
+  separatorComponent: SeparatorComponent = '>',
```
francoischalifour, line 22: Since those are JSX elements and not function components, we don't need to capitalize them.

francoischalifour: And since they're not components, the prop names might not be great.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Breadcrumb.test.tsx
francoischalifour: Yes, we'll have the `translations` prop to change this label.

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
dhayab: Changed to `rootElement` / `separatorElement` in f2542f4d.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,375 @@
...
+        }}
+      >
+        <VirtualHierarchicalMenu attributes={hierarchicalAttributes} />
+        <Breadcrumb attributes={hierarchicalAttributes} />
+      </InstantSearchHooksTestWrapper>
+    );
+
+    await wait(0);
+
+    search.mockClear();
```
sarahdayan, line 260: Why do you need to clear the mock?

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,375 @@
...
+          </ul>
+        </div>
+      </div>
+    `);
+
+    userEvent.click(getByText('Cameras & Camcorders'));
+
+    await wait(0);
+
+    expect(search).toHaveBeenLastCalledWith(
```
sarahdayan, line 316: You can also assert the number of calls before this assertion, so we ensure this doesn't fire more queries than expected.

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { isModifierClick } from './lib/isModifierClick';
+
+import type { CreateURL } from 'instantsearch.js';
+import type { BreadcrumbConnectorParamsItem as BreadcrumbItem } from 'instantsearch.js/es/connectors/breadcrumb/connectBreadcrumb';
+
+export type BreadcrumbProps = React.HTMLAttributes<HTMLDivElement> & {
+  items: BreadcrumbItem[];
+  createURL: CreateURL<BreadcrumbItem['value']>;
+  onNavigate: (value: BreadcrumbItem['value']) => void;
+  rootElement?: React.ReactNode;
```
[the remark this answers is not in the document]
dhayab: Yes, ultimately. I'll implement the `translations` structure in this PR.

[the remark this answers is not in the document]
dhayab: Done in 26eb22df.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Breadcrumb.test.tsx
dhayab: I did not need it originally, but now that I have a single mock for all tests I'm clearing it before each test to get an accurate number of its calls (see 773e4895).

packages/react-instantsearch-hooks-dom/src/widgets/Breadcrumb.tsx
```diff
@@ -36,6 +36,10 @@ export function Breadcrumb({
       items={items}
       hasItems={canRefine}
       onNavigate={refine}
+      translations={{
+        root: 'Home',
+        separator: '>',
```
Haroenv, line 41: do we allow these to be overridden via props?

dhayab: I don't think Home should be overridden until we have a public translations API. For `separator` I'm not sure what's the best way to handle it.

What I understand is, there are 2 "separator" props:
- one is for the hook, to specify which character is used as a separator in the facet values
- another is for the UI

I initially put the UI separator in the translations, but stricly speaking it should not be there. Is it fine like this?
Alternatively we could default the separator prop used for the hook and stop exposing it (since we can be opinionated in our default widget implementation), and instead use the prop for the UI separator. 

Wdyt?

Haroenv: I'm ok with initially having no way to customise it then

dhayab: Done in 7314cce

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
+
+import { cx } from './lib/cx';
+import { isModifierClick } from './lib/isModifierClick';
+
+import type { useBreadcrumb } from 'react-instantsearch-hooks';
+
+export type BreadcrumbTranslations = {
+  /**
+   * The label of the root element.
```
Haroenv, line 10:
```suggestion
   * The label of the root element
```

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
```diff
@@ -0,0 +1,142 @@
...
+  items = [],
+  hasItems,
+  createURL,
+  onNavigate,
+  separator = '>',
+  translations,
+  ...props
+}: BreadcrumbProps) {
+  const handleClick =
+    (value: BreadcrumbProps['items'][0]['value']) =>
```
Haroenv, line 68:
```suggestion
    (value: BreadcrumbProps['items'][number]['value']) =>
```

```suggestion
    (value: BreadcrumbItem['value']) =>
```

```suggestion
    (value: string) =>
```


I don't really mind which one of these is picked

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,271 @@
...
+        separator: 'SEPARATOR',
+        link: 'LINK',
+      },
+    });
+    const { container } = render(<Breadcrumb {...props} />);
+
+    expect(container).toMatchInlineSnapshot(`
+      <div>
+        <div
+          class="ais-Breadcrumb ROOT"
```
Haroenv, line 212: classname should be set here too, for consistency with the other tests

Haroenv (approved):

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,360 @@
...
+    Promise.resolve(
+      createMultiSearchResponse(
+        ...requests.map(() =>
+          createSingleSearchResponse({
+            facets: hierarchicalFacets,
+          })
+        )
+      )
+    )
+  );
```
sarahdayan, line 37: I think you can inline this in the search client, then use `searchClient.search` when testing what the method has been called with in your tests.

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
sarahdayan: Slight preference for option 2 (`BreadcrumbItem['value']`).

dhayab: `BreadcrumbConnectorParamsItem` is not well exported, not sure that's a good idea to import it as is:
```
import type { BreadcrumbConnectorParamsItem } from 'instantsearch.js/es/connectors/breadcrumb/connectBreadcrumb';
```

I can create a local type that maps to `UseBreadcrumbRenderState['items'][number]`, which can bring more legibility to the code, if that's ok.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Breadcrumb.test.tsx
dhayab: That's great, thanks for this suggestion!

sarahdayan (approved):

PR #3383, opened 2022-03-21
feat(hooks): implement <HitsPerPage>

francoischalifour:
## Description

This introduces the `<HitsPerPage>` UI widget in React InstantSearch Hooks.

## Usage

The API is relying on the InstantSearch.js API:

```jsx
function App(props) {
  return (
    <InstantSearch {...props}>
      <HitsPerPage
        items={[
          { label: '10', value: 10, default: true },
          { label: '20', value: 20 },
          { label: '30', value: 30 },
        ]}
      />
    </InstantSearch>
  );
}
```

It may be more idiomatic to use a `defaultValue` prop in React, but I don't think it's a strong-enough signal to diverge our APIs yet.

```jsx
function App(props) {
  return (
    <InstantSearch {...props}>
      <HitsPerPage
        defaultValue={10} // this is not supported
        items={[
          { label: '10', value: 10 },
          { label: '20', value: 20 },
          { label: '30', value: 30 },
        ]}
      />
    </InstantSearch>
  );
}
```


Haroenv (approved):

sarahdayan (approved):

sarahdayan:
You can also remove the component and its export from the `hooks-example` demo, since we pinned the sandbox in the docs.

Haroenv (changes requested):
until ci passes again :)

sarahdayan:
The failing CI is because TypeScript thinks `onChange` is either our type or `React.FormEventHandler<HTMLDivElement>` and when passing the function reference it cannot narrow.

We can fix it by changing the type slightly:

```ts
export type HitsPerPageProps = Omit<
  React.HTMLAttributes<HTMLDivElement>,
  'onChange'
> & {
  items: HitsPerPageItem[];
  onChange: (value: number) => void;
  currentValue: number;
};
```

francoischalifour:
I'll revert 05df6489395c6822c27c99f72d2db780df9fca3e if you don't mind because that's more explicit and straightforward to me.

Haroenv (approved):

sarahdayan (approved):

PR #3385, opened 2022-03-21
fix(ssr): perform initial multi-index search using a single request

dhayab:
**Summary**

This PR changes how the initial request performed by `findResultsState()` is made in React InstantSearch SSR environments.

See [FX-1256](https://algolia.atlassian.net/browse/FX-1256) / [CR-782](https://algolia.atlassian.net/browse/CR-782)


Haroenv (approved):
Great code! I guess I'm wondering whether this is a fix or a feature?

dhayab:
@Haroenv I was wondering as well since technically the previous implementation works as expected, but the fact that it has an impact on unit consumption made me flag it as a bug.

Haroenv:
sounds good

dhayab:
You're right. Not sure how I can single out the dom package, I'll probably check with you beforehand to make sure its done correctly.

PR #3386, opened 2022-03-21
feat(hooks): introduce Translations API

sarahdayan:
## Summary

This introduces the Translations API, allowing users to pass optional translations to widgets to customize labels, placeholders, and other text-based attributes.

It follows the [Translations API RFC](https://github.com/algolia/instantsearch-rfcs/blob/5414147664e12457136f50bb9511c9081705929f/accepted/translations.md) and the `templates` API in InstantSearch.js (without the HTML properties).

For now I've only translated the `ClearRefinements` widget to give an example of how the API works. Once we agree on the API I'll translate all the other existing widgets.


packages/react-instantsearch-hooks-dom/src/widgets/ClearRefinements.tsx
```diff
@@ -34,7 +35,10 @@ export function ClearRefinements(props: ClearRefinementsProps) {
   return (
     <ClearRefinementsUiComponent
       {...otherProps}
-      resetLabel={resetLabel}
+      translations={{
+        resetLabel: 'Clear refinements',
+        ...translations,
```
sarahdayan, line 40: This assumes translations are all single-level objects, which I believe is easier to grasp for users anyway.

Haroenv (approved):

PR #3387, opened 2022-03-21
feat(hooks): translate Pagination widget

sarahdayan:
## Summary

This translates the `<Pagination>` widget internally.

The translations are the same as React InstantSearch's [`Pagination`](https://www.algolia.com/doc/api-reference/widgets/pagination/react/#widget-param-translations) translations. We can change them later once we validate the Translations API RFC.


packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
```diff
@@ -52,26 +96,34 @@ export function Pagination({
...
-          const label = String(page + 1);
+          const label =
+            typeof translations.page === 'string'
+              ? translations.page
+              : translations.page(page);
+
+          const ariaLabel =
+            typeof translations.ariaPage === 'string'
+              ? translations.ariaPage
+              : translations.ariaPage(page);
```
Haroenv, line 126: the aria label for a dynamic page makes no sense to be a string. As this is for internal use only, we should accept only the function I guess?

Haroenv (changes requested):

packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
```diff
@@ -52,26 +96,34 @@ export function Pagination({
...
-            ‹
+            {translations.previous}
           </PaginationItem>
         )}
         {pages.map((page) => {
-          const label = String(page + 1);
+          const label =
+            typeof translations.page === 'string'
+              ? translations.page
+              : translations.page(page);
```
francoischalifour, line 121: We should rather check for `typeof x === 'function'` and execute it if so, instead of the opposite, because here you'll easily end up calling a `number` or any other data types.

francoischalifour, line 121: For translations it makes sense to provide the actual 1-based page number  imo (first being `1` and not `0`).

sarahdayan: We probably want a function all the time here though, similar to https://github.com/algolia/react-instantsearch/pull/3387#discussion_r833014745.

sarahdayan: Addressed in 91703928.

sarahdayan: Addressed in 91703928.

packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
```diff
@@ -5,6 +5,49 @@ import { isModifierClick } from './lib/isModifierClick';
...
+   */
+  next: string;
+  /**
+   * The label for the last page's button.
+   */
+  last: string;
+  /**
+   * The label for a page's button.
+   */
+  page(currentPage: number): string;
```
francoischalifour, line 28: At this point why don't we accept components?

```suggestion
  page(props: { currentPage: number }): JSX.Element | string;
```

sarahdayan: Initially that wasn't supposed to be in the scope of the Translations API, that was the split between translations and templates (which for example we pass as single components in `<SearchBox>`.

As you suggested during our last call on the topic, we may want to reconsider the problem space of translations when it comes to the widget API. For now since it's internal, I don't think we need it, and we can keep the discussion for when we do the new Translations API RFC. wdyt?

packages/react-instantsearch-hooks-dom/src/widgets/Pagination.tsx
```diff
@@ -45,6 +46,18 @@ export function Pagination({
...
+        first: '‹‹',
+        previous: '‹',
+        next: '›',
+        last: '››',
+        page: (currentPage: number) => String(currentPage),
+        ariaFirst: 'First',
+        ariaPrevious: 'Previous',
+        ariaNext: 'Next',
+        ariaLast: 'Last',
+        ariaPage: (currentPage: number) => `Page ${currentPage}`,
```
Haroenv, line 59: https://github.com/algolia/instantsearch-specs/pull/120 to include this

Haroenv (approved):

packages/react-instantsearch-hooks-dom/src/ui/Pagination.tsx
```diff
@@ -5,6 +5,49 @@ import { isModifierClick } from './lib/isModifierClick';
...
+  ariaNext: string;
+  /**
+   * The accessible label for the last page's button.
+   */
+  ariaLast: string;
+  /**
+   * The accessible label for a page's button.
+   */
+  ariaPage(currentPage: number): string;
+};
```
dhayab, line 49: Following https://github.com/algolia/instantsearch-specs/pull/120, should we have a `ariaCurrent` string for the current page accessible label? Or maybe `ariaPage()` could accept a `isCurrent` flag to compose the string differently.

dhayab (approved):
Looks good to me. I just have a small optional suggestion, but I think it's fine for now either way.

PR #3388, opened 2022-03-21
feat(hooks): translate SearchBox widget

sarahdayan:
## Summary

This translates the `<SearchBox>` widget internally.

The translations are the same as React InstantSearch's [`SearchBox`](https://www.algolia.com/doc/api-reference/widgets/search-box/react/#widget-param-translations) translations, **meaning I replaced the `placeholder` prop with `translations.placeholder`.** We can change them later once we validate the Translations API RFC.

I think for now we should simply not expose `placeholder` on the widget until we agree on whether this is part of translations or not. If we make it part of translations, we won't have to deprecate a prop. If we don't make it part of translations, we can introduce the new prop later. Either way, it avoids breaking changes after we go in beta.


packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
```diff
@@ -9,18 +9,32 @@ import type {
...
+  placeholder: string;
+};
+
 export type SearchBoxProps = React.HTMLAttributes<HTMLDivElement> & {
   inputRef: RefObject<HTMLInputElement>;
   isSearchStalled: boolean;
   onChange(event: ChangeEvent): void;
   onReset(event: FormEvent): void;
   onSubmit?(event: FormEvent): void;
-  placeholder?: string;
```
Haroenv, line 18: this change makes it so the placeholder is no longer customizable by users, as translations aren't exposed. IMO a regular dom prop shouldn't be a translation

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
```diff
@@ -9,18 +9,32 @@ import type {
...
+   * The alternative text of the reset button.
+   */
+  resetTitle: string;
+  /**
+   * The placeholder of the search input.
+   */
+  placeholder: string;
+};
+
 export type SearchBoxProps = React.HTMLAttributes<HTMLDivElement> & {
```
Haroenv, line 27:
```suggestion
export type SearchBoxProps = React.HTMLAttributes<HTMLDivElement> & Pick<React.HTMLAttributes<HTMLInputElement, 'placeholder' | 'value' | 'onChange'>> & Pick<React.HTMLAttributes<HTMLFormElement, 'onSubmit' | 'onReset'>>  {
```

(and then removed those props from the regular object too of course)

Haroenv (changes requested):

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
sarahdayan: Fair point. I changed it back in 6d2940b0.

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
```diff
@@ -2,26 +2,35 @@ import React from 'react';
...
+export type SearchBoxProps = Omit<
+  React.HTMLAttributes<HTMLDivElement>,
+  'onSubmit' | 'onReset'
+> &
+  Pick<React.HTMLAttributes<HTMLFormElement>, 'onSubmit' | 'onReset'> &
+  Pick<
+    React.InputHTMLAttributes<HTMLInputElement>,
+    'placeholder' | 'onChange'
+  > & {
+    value: string;
```
sarahdayan, line 25: I didn't pick it from `React.InputHTMLAttributes<HTMLInputElement>` because the types are too wide (`string | number | string[] | undefined`) and we're always passing a string anyway, so this avoids extra and useless runtime code.

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
sarahdayan: Done in https://github.com/algolia/react-instantsearch/commit/6d2940b090a96d6f595706b909d369561429d3f6, see https://github.com/algolia/react-instantsearch/pull/3388#discussion_r833289453.

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
```diff
@@ -17,7 +17,8 @@ export type SearchBoxProps = Omit<
   React.HTMLAttributes<HTMLDivElement>,
   'onSubmit' | 'onReset'
 > &
-  Pick<React.HTMLAttributes<HTMLFormElement>, 'onSubmit' | 'onReset'> &
+  Pick<React.HTMLAttributes<HTMLFormElement>, 'onSubmit'> &
+  Required<Pick<React.HTMLAttributes<HTMLFormElement>, 'onReset'>> &
```
Haroenv, line 21:
```suggestion
  Required<Pick<React.HTMLAttributes<HTMLFormElement>, 'onSubmit' | 'onReset'>> &
```

This is simpler, no?

sarahdayan: We don't require `onSubmit`, it's a forwarded prop but our API doesn't depend on it.

packages/react-instantsearch-hooks-dom/src/widgets/SearchBox.tsx
```diff
@@ -3,17 +3,17 @@ import { useSearchBox } from 'react-instantsearch-hooks';
...
 import { SearchBox as SearchBoxUiComponent } from '../ui/SearchBox';
 
+import type { Translatable } from '../types';
 import type { SearchBoxProps as SearchBoxUiComponentProps } from '../ui/SearchBox';
-import type { ChangeEvent } from 'react';
 import type { UseSearchBoxProps } from 'react-instantsearch-hooks';
 
 export type SearchBoxProps = Omit<
-  SearchBoxUiComponentProps,
+  Translatable<SearchBoxUiComponentProps>,
```
Haroenv, line 11: I thought we weren't going to make the translations available top-level yet?

sarahdayan: Absolutely, I think I effed up a rebase. Addressed 423f4775.

sarahdayan:
Waiting for https://github.com/algolia/react-instantsearch/pull/3394 to be merged to rebase.

dhayab (approved):

Haroenv (approved):

packages/react-instantsearch-hooks-dom/src/widgets/__testutils__/all-widgets.tsx
```diff
@@ -56,6 +56,9 @@ function Widget<TWidget extends SingleWidget>({
     case 'RefinementList': {
       return <widget.Component attribute="brand" {...props} />;
     }
+    case 'SearchBox': {
+      return <widget.Component onSubmit={undefined} {...props} />;
```
Haroenv, line 60: I wonder if something is wrong with the props that makes this onSubmit required to be passed in a way compatible with both div and form, we should look into this after merging

PR #3390, opened 2022-03-23
chore: mark react-instantsearch-hooks-dom package as private

dhayab:
**Summary**

This PR marks the react-instantsearch-hooks-dom package as private to exclude it from being published until it is ready.


francoischalifour (approved):

dhayab:
I don't see detailed logs, but I'm wondering if the **ci/codesandbox** job fails because the hooks-dom package is now private. I guess it's fine since I'll remove this just after performing release of the other packages.

francoischalifour:
Marking the package as private shouldn't be necessary anyway since Ship.js won't try to release it based on the config.

packages/react-instantsearch-hooks-dom/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "react-instantsearch-hooks-dom",
+  "private": true,
```
Haroenv, line 3:
```suggestion
```

Haroenv (approved):

PR #3391, opened 2022-03-23
feat(hooks): implement PoweredBy component

sarahdayan:
## Summary

This introduces the `<PoweredBy>` DOM widget.

Ticket: [FX-1159](https://algolia.atlassian.net/browse/FX-1159)

## Implementation details

Now that we have an API for `classNames`, I've implemented it in this widget.

Unlike the way it's been implemented in [`<Highlight>`](https://github.com/algolia/react-instantsearch/blob/de980324a1891a5f7c6b358f9d91e49b473315a4/packages/react-instantsearch-hooks-dom/src/widgets/Highlight.tsx#L42-L50) and [`<Snippet>`](https://github.com/algolia/react-instantsearch/blob/de980324a1891a5f7c6b358f9d91e49b473315a4/packages/react-instantsearch-hooks-dom/src/widgets/Snippet.tsx#L42-L50), I've added the default classes directly in the UI component (instead of passing them from the widget). This doesn't change anything for the user, but it avoids splitting the default UI implementation between the widget and UI component. **This way, the code doesn't have to be repeated across all flavors, which minimizes the odds of inconsistencies across them.**

## Results

- New `<PoweredBy>` widget
- Replaced `<PoweredBy>` with the widget in example


packages/react-instantsearch-hooks-dom/src/ui/PoweredBy.tsx
```diff
@@ -1,21 +1,57 @@
...
+   */
+  link: string;
+  /**
+   * Class names to apply to the SVG logo
+   */
+  logo: string;
+};
+
+export type PoweredByProps = React.HTMLAttributes<HTMLDivElement> & {
+  classNames: Partial<PoweredByClassNames>;
```
Haroenv, line 21: if we follow the same pattern as highlight, in the ui component the class names are required, and passed directly from the prop; in the widget they are optional and get the default value.

sarahdayan: See my [explanation](https://github.com/algolia/react-instantsearch/pull/3391#issue-1178092284) for why.

packages/react-instantsearch-hooks-dom/src/ui/PoweredBy.tsx
```diff
@@ -1,21 +1,57 @@
...
+  theme = 'light',
+  ...props
+}: PoweredByProps) {
   return (
-    <div className="ais-PoweredBy">
+    <div
+      {...props}
+      className={cx(
+        'ais-PoweredBy',
+        theme === 'dark' ? 'ais-PoweredBy--dark' : 'ais-PoweredBy--light',
```
Haroenv, line 37:
```suggestion
        theme === 'dark' ? 'ais-PoweredBy--dark' : 'ais-PoweredBy--light',
        theme === 'dark' ? classNames.dark : classNames.light,
```

sarahdayan: There's an inconsistency between [InstantSearch.js](https://www.algolia.com/doc/api-reference/widgets/powered-by/js/?client=vue#widget-param-cssclasses) and [Vue InstantSearch](https://www.algolia.com/doc/api-reference/widgets/powered-by/vue/?client=react#widget-param-class-names) regarding those CSS classes. **What should we follow as the source of truth?**

The benefit of following Vue InstantSearch is that we perform the condition for the user, that's less code for them. If we agree on this, I'll also make the change in InstantSearch.js to avoid further drifts between flavors.

Haroenv:
```suggestion
        theme === 'dark' ? cx(classNames.dark, 'ais-PoweredBy--dark') : cx(classNames.light, 'ais-PoweredBy--light'),
```

alternative option to that condition.



IMO we should make sure every class name is modifiable, which is useful for eg. tailwind

sarahdayan: Yes I much prefer this option rather than doing the condition twice. Addressed in 2f8451e3.

Haroenv (approved):

packages/react-instantsearch-hooks-dom/src/ui/PoweredBy.tsx
```diff
@@ -1,21 +1,67 @@
...
+};
+
+export type PoweredByProps = React.HTMLAttributes<HTMLDivElement> & {
+  classNames: Partial<PoweredByClassNames>;
+  url: string;
+  theme?: 'light' | 'dark';
+};
+
+export function PoweredBy({
+  classNames,
```
dhayab, line 35:
```suggestion
  classNames = {},
```

packages/react-instantsearch-hooks-dom/src/ui/PoweredBy.tsx
```diff
@@ -1,21 +1,67 @@
...
+   */
+  link: string;
+  /**
+   * Class names to apply to the SVG logo
+   */
+  logo: string;
+};
+
+export type PoweredByProps = React.HTMLAttributes<HTMLDivElement> & {
+  classNames: Partial<PoweredByClassNames>;
```
dhayab, line 29:
```suggestion
  classNames?: Partial<PoweredByClassNames>;
```

packages/react-instantsearch-hooks-dom/src/widgets/PoweredBy.tsx
```diff
@@ -0,0 +1,15 @@
...
+import type { ClassNames } from '../types';
+import type { PoweredByProps as PoweredByUiComponentProps } from '../ui/PoweredBy';
+
+export type PoweredByProps = Omit<ClassNames<PoweredByUiComponentProps>, 'url'>;
+
+export function PoweredBy({ classNames = {}, ...props }: PoweredByProps) {
+  const { url } = usePoweredBy();
+
+  return <PoweredByUiComponent {...props} url={url} classNames={classNames} />;
+}
```
dhayab, line 15: In conjunction with the other suggestions, we could now remove the defaulted `classNames` prop here and let it be passed to the UI component with `...props`:

```suggestion
export function PoweredBy(props: PoweredByProps) {
  const { url } = usePoweredBy();

  return <PoweredByUiComponent {...props} url={url} />;
}
```

packages/react-instantsearch-hooks-dom/src/widgets/PoweredBy.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { usePoweredBy } from 'react-instantsearch-hooks';
+
+import { PoweredBy as PoweredByUiComponent } from '../ui/PoweredBy';
+
+import type { ClassNames } from '../types';
+import type { PoweredByProps as PoweredByUiComponentProps } from '../ui/PoweredBy';
+
+export type PoweredByProps = Omit<ClassNames<PoweredByUiComponentProps>, 'url'>;
```
dhayab, line 9: Not sure we need the `ClassNames` type anymore since it's already a Partial in the UI component.

dhayab (commented):
A few suggestions, mainly on shifting the default value of `classNames` to the UI component.

packages/react-instantsearch-hooks-dom/src/widgets/PoweredBy.tsx
Haroenv: yes, ClassNames can be removed in #3394

packages/react-instantsearch-hooks-dom/src/widgets/PoweredBy.tsx
sarahdayan: Addressed in e84cf31a.

dhayab (approved):

PR #3399, opened 2022-03-24
fix(hooks): compute initial search parameters from widget

francoischalifour:
## Description

This changes how we compute the initial Helper state that we give to `widget.getWidgetRenderState()` to retrieve the initial widget React state.

The problem originally surfaced when server-side rendering the `<Menu>` component:

> Error: Cannot refine the undeclared facet categories; it should be added to the helper options facets, disjunctiveFacets or hierarchicalFacets

This happens because the `helper.state` and `state` used to call `widget.getWidgetRenderState()` mismatched. In algolia/instantsearch.js#4960, we tried to fix this problem by always relying on `state`. But this constraint is hard to respect, and hard to implement in practice because of stale closures (in `refine()` and `createURL()`).

I explain in that PR why we had to do it this way, essentially because the helper state was the resolved helper state, including all the inherited search parameters from the parent indices. With a resolved helper state, InstantSearch considers the parameters as controlled and therefore doesn't get notified of new changes coming from other indices. The part that we were missing is that there's a way to compute the local index search parameters without inheriting the parent index via `index.getWidgetSearchParameters()`.

## Solution

This PR is another cleaner approach directly in React InstantSearch Hooks. We assign the widget search parameters gotten from `widget.getWidgetSearchParameters()` to the helper state so that the widget is directly aware of its search parameters.

## Result

- It fixes the original problem with `<Menu>` throwing about the undeclared facet.
- It also fixes `useClearRefinements()` and `useCurrentRefinements()` that used to initially render with an empty state. They're now directly aware of their refinements during first render.
- We'll be able to rollback some of the changes introduced in algolia/instantsearch.js#4960

### DIFF
--- a/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
+++ b/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
@@ -89,17 +89,19 @@ export function useConnector<
     if (widget.getWidgetRenderState) {
       // The helper exists because we've started InstantSearch.
       const helper = parentIndex.getHelper()!;
+      const uiState = parentIndex.getWidgetUiState({})[
+        parentIndex.getIndexId()
+      ];
+      helper.state =
+        widget.getWidgetSearchParameters?.(helper.state, { uiState }) ||
+        helper.state;
       const results =
         // On SSR, we get the results injected on the Index.
         parentIndex.getResults() ||
         // On the browser, we create fallback results based on the widget's
         // `getWidgetSearchParameters()` method to inject the initial UI state,
         // or fall back to the helper state.
-        createSearchResults(
-          widget.getWidgetSearchParameters?.(helper.state, {
-            uiState: parentIndex.getWidgetUiState({})[parentIndex.getIndexId()],
-          }) || helper.state
-        );
+        createSearchResults(helper.state);
       const scopedResults = parentIndex
         .getScopedResults()
         .map((scopedResult) => {
@@ -124,7 +126,7 @@ export function useConnector<
         instantSearchInstance: search,
         results,
         scopedResults,
-        state: results._state,
+        state: helper.state,
         renderState: search.renderState,
         templatesConfig: search.templatesConfig,
         createURL: parentIndex.createURL,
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch-hooks/src/hooks/useConnector.ts
```diff
@@ -89,17 +89,19 @@ export function useConnector<
     if (widget.getWidgetRenderState) {
       // The helper exists because we've started InstantSearch.
       const helper = parentIndex.getHelper()!;
+      const uiState = parentIndex.getWidgetUiState({})[
+        parentIndex.getIndexId()
+      ];
+      helper.state =
+        widget.getWidgetSearchParameters?.(helper.state, { uiState }) ||
+        helper.state;
```
Haroenv, line 97: this means that if this part of the function gets called, but the widget never actually mounts, there will be stray search parameters not associated to a widget (as we don't call dispose). Could this ever be a real issue?

Haroenv (approved):
I'll approve this, because I'm not seeing how what I described can cause real issues.

Maybe we should call `widget.dispose` after the initial getWidgetRenderState and before it "really gets mounted"? Or some other way to set helper.state back to the previous value after fake rendering (`cache state; getWidgetRenderState; reset state;`)?

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
francoischalifour: Why wouldn't the widget mount at this point?

And `getWidgetSearchParameters()` returns a value without triggering side effects, so I'm not sure to understand what you're saying.

Haroenv: getWidgetSearchParameters doesn't have any side-effects, but `helper.state = getWidgetSearchParameters()` is a side-effect. 

I'm not sure why it wouldn't be mounted, but I'm not 100% sure whether the contract of react allows a memo to be called before/after the component is already mounted? Are there any specs of that? Is it possible a memo is called without the rest of the component being called?

sarahdayan (approved):

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
francoischalifour: Assigning `helper.state` is indeed a side effect. I think it's localized in that case because each widget overrides it for its initial rendering, and then the InstantSearch.js lifecycle takes over and sets it.

React can call memo anytime that it wants to call it (for memory/optimization purposes), so there's no certainty that it'll be called only once. I don't think that the new code introduced competes with the existing logic of `useConnector()` and `useMemo()` though. We already had this consideration before this change.

PR #3400, opened 2022-03-24
feat(hooks): implement ToggleRefinement component

sarahdayan:
## Summary

This introduces the `<ToggleRefinement>` DOM widget.

Ticket: [FX-1165](https://algolia.atlassian.net/browse/FX-1165)

## Results

- New `<ToggleRefinement>` widget
- Replaced `<ToggleRefinement>` with the widget in example


packages/react-instantsearch-hooks-dom/src/ui/ToggleRefinement.tsx
```diff
@@ -0,0 +1,64 @@
...
+      {...props}
+      className={cx('ais-ToggleRefinement', classNames.root, props.className)}
+    >
+      <label className={cx('ais-ToggleRefinement-label', classNames.label)}>
+        <input
+          className={cx('ais-ToggleRefinement-checkbox', classNames.checkbox)}
+          type="checkbox"
+          checked={checked}
+          onChange={(event) => {
+            onChange(event.target.checked);
```
Haroenv, line 52: in InstantSearch.js this is:

```suggestion
            onChange({ checked: event.target.checked });
```

I guess typescript explaining what the boolean can also help, but a loose boolean can be confusing

packages/react-instantsearch-hooks-dom/src/ui/__tests__/ToggleRefinement.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    expect(props.onChange).toHaveBeenLastCalledWith(true);
+  });
+
+  test('customizes the class names', () => {
+    const props = createProps({
+      classNames: {
+        root: 'MyCustomToggleRefinement',
+        label: 'MyCustomToggleRefinementLabel',
+        checkbox: 'MyCustomToggleRefinementCheckbox',
+        labelText: 'MyCustomToggleRefinementLabelText',
```
Haroenv, line 90: in the others I have this as the uppercase version of the label (so it's easily to visually see in the snapshot). I'm not sure if we need to be consistent here though, it's just a test

packages/react-instantsearch-hooks-dom/src/widgets/ToggleRefinement.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { refine, value } = useToggleRefinement(
+    { attribute, on, off },
+    {
+      $$widgetType: 'ais.toggleRefinement',
+    }
+  );
+
+  return (
+    <ToggleRefinementUiComponent
+      {...props}
```
Haroenv, line 30:
```suggestion
      label={attribute}
      {...props}
```

In InstantSearch.js we have the equivalent as this, meaning the label isn't required. Not sure what's best though

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/ToggleRefinement.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+  });
+
+  test('customizes the class names', async () => {
+    const { container } = render(
+      <InstantSearchHooksTestWrapper>
+        <ToggleRefinement
+          attribute="free_shipping"
+          label="Free shipping"
+          classNames={{
+            root: 'MyCustomToggleRefinement',
```
Haroenv, line 183: same question about the capitals

Haroenv (commented):
👍

packages/react-instantsearch-hooks-dom/src/ui/ToggleRefinement.tsx
sarahdayan: Yes, I kept this part in the widget so we're not leaking the API of `useToggleRefinement`. I think this is fine since this is a UI component, users don't interact with it.

packages/react-instantsearch-hooks-dom/src/ui/__tests__/ToggleRefinement.test.tsx
sarahdayan: I like consistency too, and I agree that capitals are better because they stick out.

Updated in db1b19a0.

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/ToggleRefinement.test.tsx
sarahdayan: Addressed in db1b19a0.

packages/react-instantsearch-hooks-dom/src/widgets/ToggleRefinement.tsx
sarahdayan: The problem is that `attribute` is the name of the record property, but you may need to specify a display name. In InstantSearch.js it's solved with the [`labelText`](https://www.algolia.com/doc/api-reference/widgets/toggle-refinement/js/#templates) template, but since we're not opening translations yet, it would mean users can't customize the display name.

Haroenv: you would be able to, as props is still spread right after it, which would override the label, if given, this is just giving a default value to the label I'm suggesting :)

Haroenv (approved):
remaining comment not a blocker

packages/react-instantsearch-hooks-dom/src/widgets/ToggleRefinement.tsx
sarahdayan: Good point! Addressed in 0b96386b.

packages/react-instantsearch-hooks-dom/src/widgets/__testutils__/all-widgets.tsx
```diff
@@ -55,6 +55,9 @@ function Widget<TWidget extends SingleWidget>({
     case 'RefinementList': {
       return <widget.Component attribute="brand" {...props} />;
     }
+    case 'ToggleRefinement': {
+      return <widget.Component attribute="free_shipping" />;
+    }
```
Haroenv, line 60:
```suggestion
    case 'ToggleRefinement':
    case 'RefinementList': {
      return <widget.Component attribute="brand" {...props} />;
    }
```

dhayab (approved):
Looks good to me :)

Haroenv (approved):

PR #3403, opened 2022-03-25
docs(hooks): cleanup unused components in example

sarahdayan:
## Summary

This cleans up the Hooks example by removing the components that have already been converted to widgets.


Haroenv:
I did this too in https://github.com/algolia/react-instantsearch/pull/3398, but it makes the hooks-next and hooks-ssr examples fail too on /master, is that ok?

Haroenv (approved):

PR #3405, opened 2022-03-28
feat(hooks): implement Menu component

Haroenv:
**Summary**

Implements the Menu component and does a small refactor to RefinementList to keep the widgets classname-free

**Result**

FX-1156


packages/react-instantsearch-hooks-dom/src/ui/__tests__/Menu.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+          count: 633,
+          isRefined: false,
+        },
+      ],
+      onRefine: jest.fn(),
+      createURL: (value) => `#${value}`,
+      canToggleShowMore: true,
+      toggleShowMore: () => {
+        isShowingMore = !isShowingMore;
+      },
```
sarahdayan, line 32: Not sure this is necessary to reimplement the behavior of the function that the widget will pass. I'd make this a mock function without any implementation, then have a test that clicks the button and ensures it has been called. wdyt?

sarahdayan (commented):
Clean job ✨ And thanks for the refactor!

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Menu.test.tsx
```diff
@@ -0,0 +1,1746 @@
...
+        'OtterBox',
+        'Metra',
+        'LG',
+        'KitchenAid',
+        'Insignia™',
+      ]);
+    });
+  });
+
+  describe('showMore', () => {
```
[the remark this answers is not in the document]
Haroenv: toggling sounds like toggling the facet values / refining, no?

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Menu.test.tsx
Haroenv: It seems simpler to me to have a correct implementation than a mock, as the real rendering state is based on isShowingMore, so that would still need to be updated. This is closer to real-life usage :)

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/Menu.test.tsx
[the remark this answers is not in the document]
sarahdayan: It could yes. Otherwise we can go with "showing more/less" (and change it in the `RefinementList` test) wdyt?

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Menu.test.tsx
sarahdayan: If you want to test both (function called + updated state) then go ahead! It can still be in a mock though 🙂

packages/react-instantsearch-hooks-dom/src/ui/__tests__/Menu.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+          count: 633,
+          isRefined: false,
+        },
+      ],
+      onRefine: jest.fn(),
+      createURL: (value) => `#${value}`,
+      canToggleShowMore: true,
+      onToggleShowMore: () => {
+        isShowingMore = !isShowingMore;
+      },
```
sarahdayan, line 32: Then you can add a test where you click the button and assert:
- That the function was called
- That the UI updated accordingly

```suggestion
      onToggleShowMore: jest.fn(() => {
        isShowingMore = !isShowingMore;
      }),
```

Haroenv: added a test for just the function being called, because the state really is the responsibility of a wrapping component, so it doesn't need to be tested in detail here

sarahdayan (approved):
Just need to use `userEvent` instead of `fireEvent` and LGTM!

PR #3412, opened 2022-03-31
feat(refinements): use escaped value for refining

Haroenv:
**Summary**

The helper exposes a `value` to getFacetValues too now, which is the same as the name/label, except escaped. We can use this instead of `name` in all relevant widgets.

relies on https://github.com/algolia/algoliasearch-helper-js/pull/889 and https://github.com/algolia/algoliasearch-helper-js/pull/900 to show the escaped value (leading `-` gets transformed into `\-`

**Result**

Facet values that start with `-` are automatically escaped, but you can still refine on a literal value starting with `-`

FX-1135

same as https://github.com/algolia/instantsearch.js/pull/5039

### DIFF
--- a/packages/react-instantsearch-core/package.json
+++ b/packages/react-instantsearch-core/package.json
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch-helper": "^3.7.4",
+    "algoliasearch-helper": "https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/493f99a3/algoliasearch-helper",
     "prop-types": "^15.6.2",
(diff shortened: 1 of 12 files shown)


packages/react-instantsearch-core/package.json
```diff
@@ -37,7 +37,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch-helper": "^3.7.4",
+    "algoliasearch-helper": "https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/493f99a3/algoliasearch-helper",
```
Haroenv, line 40: these must be updated before merged

sarahdayan (approved):
Oy!

Issue #3407, opened 2022-03-30
`allSearchResults` is updated too many times after searching


ksntcrq:
## 🐛 Bug description

When doing a multiple index query with several `Index`, `allSearchResults` is updated as many times as there are indices after searching instead of at once.

## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Open the sandbox
2. Type a letter in the search field
3. See the console:
  1. Searching becomes `true`
  2. Searching becomes `false` and the `nbHits` of the main index are updated
  3. Only then the `nbHits` of the other index is updated

**Live reproduction:**

https://codesandbox.io/s/kind-varahamihira-srcdhn?file=/src/App.js

## 💭 Expected behavior

I would expect all the `nbHits` to be updated at the same time, when `searching` becomes false, because this means that searching is done and that everything was retrieved.

## Environment

- OS: macOS
- Browser: Chrome
- Version: 6.23.0

## Additional context

I need everything to be updated at once because I need to send a segment event when we're done searching, with the number of results for each index. For this, I'm looking for a `searching` changing from `true` to `false` and need the number of hits to be updated at the same time.

dhayab:
Hi @ksntcrq , thanks for filing this. It seems like an issue indeed, as `searching` is set to `false` as soon as the state is updated with the first index's results.

We're going to look at this, and in the meantime, you can work around this by thoroughly checking that the search parameters are identical in all indices. Here is a sandbox that checks for the query parameter for example: https://codesandbox.io/s/focused-nobel-jiket5?file=/src/App.js.

ksntcrq:
@dhayab Thanks! 

I see that your Pull Request was already merged. Any ETA on a release?

Haroenv:
It will be released automatically once this PR merges, so in 10-20 minutes, thanks again for reporting! https://github.com/algolia/react-instantsearch/pull/3421

ksntcrq:
Just tried with the new version, I can confirm it's fixed. Thanks!

PR #3419, opened 2022-04-04
fix(multi-index): correctly set `searching` prop in multi-index result states

dhayab:
**Summary**

Currently, search results update the state with the `searching` prop set inconditionnally to `false`, which don't represent the reality of multi-index searches. This PR counts the number of expected search results to set the property to its right value.

See details and sandbox in issue https://github.com/algolia/react-instantsearch/issues/3407.

**Result**

The `searching` prop is set to `true` for all but the last expected request.


Haroenv (approved):
👍 love this fix!

PR #3423, opened 2022-04-05
feat(hooks): implement <RangeInput>

dhayab:
# **Summary**

This introduces the `<RangeInput>` DOM widget.

Ticket: [FX-1162](https://algolia.atlassian.net/browse/FX-1162)

# **Result**

- New `<RangeInput>` widget
- DOM tests for UI component and widget
- Replaces `<RangeInput>` with the widget in the hooks example


packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RangeInput.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+            },
+          },
+        })
+      )
+    )
+  );
+});
+
+beforeEach(() => {
+  search.mockClear();
```
francoischalifour, line 36: How about we create a new function in each test so that we don't need to care about resetting the mock?

examples/hooks/App.tsx
```diff
@@ -15,6 +15,7 @@ import {
   InfiniteHits,
   Menu,
   Pagination,
+  RangeInput,
```
francoischalifour, line 18: The `<RangeInput>` doesn't display because the `<NumericMenu>` takes precedence over it. I think we can remove `<NumericMenu>` now.

francoischalifour (commented):
The behavior is not identical with InstantSearch.js.

There are HTML form checks `> 0`, and the `0` prefix isn't replaced. See gifs.

**InstantSearch.js**
rangeinput-is

**React InstantSearch Hooks**
rangeinput-rish

Haroenv (dismissed):
same question as Francois around inconsistent DOM with InstantSearch.js

dhayab:
Applied suggestions in 0d48847 including notably:

- on input change, remove `0` as prefix for another digit
- wrap inputs in labels
- new `step` property in inputs

packages/react-instantsearch-hooks-dom/src/widgets/__tests__/RangeInput.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+import {
+  createMultiSearchResponse,
+  createSearchClient,
+  createSingleSearchResponse,
+} from '../../../../../test/mock';
+import { InstantSearchHooksTestWrapper, wait } from '../../../../../test/utils';
+import { RangeInput } from '../RangeInput';
+
+function getSearchClient() {
```
francoischalifour, line 13: It doesn't make much sense to have `get*` call `create*`. We could call it something like `createSearchClientWithFacetsStats()`.

packages/react-instantsearch-hooks-dom/src/ui/RangeInput.tsx
```diff
@@ -0,0 +1,174 @@
...
+  React.useEffect(() => {
+    setRange({ from: values.min, to: values.max });
+  }, [values.min, values.max]);
+
+  const onInput =
+    (key: string) =>
+    ({ currentTarget }: React.SyntheticEvent<HTMLInputElement>) => {
+      const value = Number(currentTarget.value);
+
+      currentTarget.value = value.toString();
```
francoischalifour, line 109: What's this line for? Tests don't fail without it.

packages/react-instantsearch-hooks-dom/src/ui/RangeInput.tsx
```diff
@@ -0,0 +1,174 @@
...
+      minValue !== -Infinity && minValue !== min
+        ? minValue
+        : unsetNumberInputValue,
+    max:
+      maxValue !== Infinity && maxValue !== max
+        ? maxValue
+        : unsetNumberInputValue,
+  };
+
+  const [{ from, to }, setRange] = React.useState({
```
francoischalifour, line 95: We use this syntax everywhere else.

```suggestion
  const [{ from, to }, setRange] =useState({
```

packages/react-instantsearch-hooks-dom/src/ui/RangeInput.tsx
```diff
@@ -0,0 +1,174 @@
...
+    from: values.min,
+    to: values.max,
+  });
+
+  React.useEffect(() => {
```
francoischalifour, line 100: We use this syntax everywhere else.

```suggestion
  useEffect(() => {
```

packages/react-instantsearch-hooks-dom/src/ui/RangeInput.tsx
```diff
@@ -0,0 +1,174 @@
...
+              cx('ais-RangeInput-input--min', classNames.inputMin)
+            )}
+            type="number"
+            min={min}
+            max={max}
+            value={from}
+            step={step}
+            placeholder={min?.toString()}
+            disabled={disabled}
+            onInput={onInput('from')}
```
francoischalifour, line 143: Why this indirection? We could inline it like this:

```suggestion
            onInput={(event) => {
              setRange({ from: Number(event.currentTarget.value), to });
            }}
```

francoischalifour (changes requested):

packages/react-instantsearch-hooks-dom/src/ui/RangeInput.tsx
dhayab: This line is what's allowing the overwrite of the `0` when it prefixes other digits (see [sandbox](https://codesandbox.io/s/recursing-golick-9hn904?file=/src/App.js)).

CleanShot 2022-04-11 at 11 59 42

I have issue with finding a good way to test that as the `.toHaveValue()` assertion expects a number, and directly checking the `value` of the input returns the expected value, whether or not the line is present.

Do you have any ideas as to why?

francoischalifour: Instead of mutating the target value, we can rely on the `value` HTML attribute to stringify:

```tsx
<input
  // ...
  value={from?.toString()}
/>
```

As for the test, it's indeed quite a browser behavior so I'm not sure without digging into internals. I would suggest adding a comment as an alternative.

dhayab: Thanks, that works! I'm guessing its due to the fact that now React checks the value as a string instead of a number.

packages/react-instantsearch-hooks-dom/src/widgets/RangeInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+  max,
+  precision,
+  ...props
+}: RangeInputProps) {
+  const { range, start, canRefine, refine } = useRange(
+    { attribute, min, max, precision },
+    { $$widgetType: 'ais.rangeInput' }
+  );
+
+  const step = 1 / Math.pow(10, precision || 0);
```
francoischalifour, line 27: We don't have any tests for `step`?

dhayab: Added tests for `precision` and `step` in b71fd75.

francoischalifour (approved):

packages/react-instantsearch-hooks-dom/src/widgets/RangeInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { RangeInput as RangeInputUiComponent } from '../ui/RangeInput';
+
+import type { RangeInputProps as RangeInputUiProps } from '../ui/RangeInput';
+import type { UseRangeProps } from 'react-instantsearch-hooks';
+
+export type RangeInputProps = Omit<
+  RangeInputUiProps,
+  'disabled' | 'onSubmit' | 'range' | 'start' | 'step' | 'translations'
+> &
```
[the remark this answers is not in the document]
Haroenv: I think the main reason omit is used is because of the root div props, it should indeed be different than this

Issue #3426, opened 2022-04-07
Hooks with React Native: pragma and pragmaFrag cannot be set when runtime is automatic


Antonhansel:
## 🐛 Bug description

Following this example: https://github.com/algolia/doc-code-samples/pull/358 by @sarahdayan 
I'm unable to run my React Native Application using Expo once I import InstantSearch from `react-instantsearch-hooks`

```
iOS Bundling failed 4201ms
node_modules/react-instantsearch-hooks/node_modules/instantsearch.js/cjs/components/CurrentRefinements/CurrentRefinements.js: /app/node_modules/react-instantsearch-hooks/node_modules/instantsearch.js/cjs/components/CurrentRefinements/CurrentRefinements.js: pragma and pragmaFrag cannot be set when runtime is automatic.
> 1 | "use strict";
    | ^
  2 |
  3 | Object.defineProperty(exports, "__esModule", {
  4 |   value: true
SyntaxError: /app/node_modules/react-instantsearch-hooks/node_modules/instantsearch.js/cjs/components/CurrentRefinements/CurrentRefinements.js: pragma and pragmaFrag cannot be set when runtime is automatic.
> 1 | "use strict";
    | ^
  2 |
  3 | Object.defineProperty(exports, "__esModule", {
  4 |   value: true
    at File.buildCodeFrameError (/app/node_modules/@babel/core/lib/transformation/file/file.js:249:12)
    at NodePath.buildCodeFrameError (/app/node_modules/@babel/traverse/lib/path/index.js:143:21)
    at PluginPass.enter (/app/node_modules/@babel/plugin-transform-react-jsx/lib/create-plugin.js:153:28)
    at newFn (/app/node_modules/@babel/traverse/lib/visitors.js:177:21)
    at NodePath._call (/app/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/app/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/app/node_modules/@babel/traverse/lib/path/context.js:100:31)
    at TraversalContext.visitQueue (/app/node_modules/@babel/traverse/lib/context.js:103:16)
    at TraversalContext.visitSingle (/app/node_modules/@babel/traverse/lib/context.js:77:19)
```

## 🔍 Bug reproduction

```
import {
  InstantSearch
} from 'react-instantsearch-hooks';
import algoliasearch from 'algoliasearch';

const searchClient = algoliasearch(
  'appId',
[...]

PR #3437, opened 2022-04-19
fix(hooks): forbid importing from instantsearch.js root path

Haroenv:
**Summary**

The root import of InstantSearch in the CJS build contains everything, including the components and widgets. These are not useful, and thus shouldn't be imported.

**Result**

closes https://github.com/algolia/react-instantsearch/issues/3426

as React Native, through Metro, imports the CJS build, it ends up trying to build the InstantSearch.js components, which it can't as they're preact.

On the individual level this could be solved by putting "module" in the resolverMainFields: https://facebook.github.io/metro/docs/configuration#resolvermainfields, but we can't suggest people to make a Metro change, as for example Expo hides the Metro config.

This likely also slightly improves bundlesize, as the keys that are added to the `instantsearch` factory are no longer needed here, EDIT: yes, the umd for example goes from 38.28 -> 37.85


sarahdayan (approved):
Can confirm this change solves the import issue in React Native.

Small suggestion on the comment for extra clarity, otherwise LGTM.

francoischalifour (approved):
Nice one

PR #3440, opened 2022-04-20
chore(hooks): type definition consistency

dhayab:
**Summary**

This PR makes sure all of our components and widgets follow the same conventions and structure.

Ticket: [FX-1339](https://algolia.atlassian.net/browse/FX-1339)


packages/react-instantsearch-hooks-dom/src/widgets/Breadcrumb.tsx
```diff
@@ -6,11 +6,9 @@ import { Breadcrumb as BreadcrumbUiComponent } from '../ui/Breadcrumb';
 import type { BreadcrumbProps as BreadcrumbUiProps } from '../ui/Breadcrumb';
 import type { UseBreadcrumbProps } from 'react-instantsearch-hooks';
 
-export type BreadcrumbProps = Omit<
-  BreadcrumbUiProps,
-  'items' | 'hasItems' | 'createURL' | 'onNavigate' | 'translations'
```
Haroenv, line 11: the omit is longer, but seems more correct. I wonder if we somehow can have a more static object of the "props we will pass to the component" rather than listing manually though, because it's indeed hard to make sure this stays correct. Maybe we can omit the required props (not sure if that's a good heuristic)

packages/react-instantsearch-hooks-dom/src/widgets/Breadcrumb.tsx
```diff
@@ -6,11 +6,9 @@ import { Breadcrumb as BreadcrumbUiComponent } from '../ui/Breadcrumb';
...
-> &
-  Omit<UseBreadcrumbProps, 'separator'>;
+export type BreadcrumbProps = React.HTMLAttributes<HTMLDivElement> &
```
francoischalifour, line 9: Why don't we use this? It's a bit easier to read.

```suggestion
export type BreadcrumbProps = React.ComponentProps<'div'> &
```

packages/react-instantsearch-hooks-dom/src/widgets/Breadcrumb.tsx
dhayab: I separated the internal and exposed props on components, to make the latter easy to use in widgets.

Not set on the naming of the internal props though (`Breadcrumbs_Internal_Props`), so feel free to jump in if you have ideas. I am also not fond of exporting the type but I need them for the ui tests.

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
```diff
@@ -47,9 +47,12 @@ type UseBreadcrumbRenderState = ReturnType<typeof useBreadcrumb>;
 
 type BreadcrumbItem = UseBreadcrumbRenderState['items'][number];
 
-export type BreadcrumbProps = React.HTMLAttributes<HTMLDivElement> &
+export type BreadcrumbProps = React.ComponentProps<'div'> & {
+  classNames?: Partial<BreadcrumbClassNames>;
+};
+
+export type BreadcrumbInternalProps = BreadcrumbProps &
```
Haroenv, line 54: should these two names be swapped? the ui component props is everything, but I find it hard to find a name for them. I still think that everything which is not required should be allowed, like 

```ts
type RequiredKeys<T> = { [K in keyof T]-?: {} extends Pick<T, K> ? never : K }[keyof T];
type OptionalKeys<T> = { [K in keyof T]-?: {} extends Pick<T, K> ? K : never }[keyof T];

// in widget:
export type BreadcrumbProps = OptionalKeys<BreadcrumbUiComponentProps> & UseBreadcrumbProps
```

This way what we'll need to be careful for is that the optionality of the keys is correct, so separator here for example should be required, but have `string | undefined` as its value

What do you think?

dhayab: To make sure I understood, we would keep all props in a single type, in which we have:
- required props → specific to the ui component
- optional props → passed to extend the widget props

To illustrate, the type for `<Breadcrumb>` would be: 
```ts
type BreadcrumbProps = React.ComponentProps<'div'> &
  Required<Pick<UseBreadcrumbRenderState, 'items' | 'createURL'>> & {
    classNames?: Partial<BreadcrumbClassNames>;
    hasItems: boolean;
    onNavigate: UseBreadcrumbRenderState['refine'];
    separator: string | undefined;
    translations: BreadcrumbTranslations;
  };
```

If this is what you had in mind, I like the idea, but there is a risk of forgetting this custom convention in the future, because it is not really self-explanatory.

Haroenv: An alternative could be:

```tsx
type WrapperProps = React.ComponentProps<'div'>;
type RequiredProps = Required<
  Pick<UseBreadcrumbRenderState, 'items' | 'createURL'> & {
    hasItems: boolean;
    onNavigate: UseBreadcrumbRenderState['refine'];
    separator: string | undefined;
    translations: BreadcrumbTranslations;
  }
>;
type OptionalProps = Partial<{
  classNames: Partial<BreadcrumbClassNames>;
}>;

export type BreadcrumbProps = WrapperProps & RequiredProps & OptionalProps;
```


This puts a little more structure, and less likely to have it wrong. What do you think? 


The part of that I'm not sure of is using `UseBreadcrumbProps. Once ui components is standalone later, it won't have access to the React InstantSearch props, so it might be better to retype it there. But we don't have to make that last decision now.

dhayab: Thanks for the suggestion!

dhayab: This structure necessited to change the signature of some components that were previously reviewed. Specifically, props that would not be passed on the widget but were optional and defaulted, are now required. I think this is fine since we don't expose this. An example of this type of change can be seen in [ui/CurrentRefinements.tsx](https://github.com/algolia/react-instantsearch/pull/3440/files#diff-fe8163ba90a2676bee0ac9f1d8754ea8654d0776783d0212a0b81004498ab55d).

I also added a `WidgetProps<UiComponentProps>` utility type to make props forwarding easier.

packages/react-instantsearch-hooks-dom/src/types/OptionalKeys.ts
```diff
@@ -0,0 +1,3 @@
+export type OptionalKeys<T> = {
+  [K in keyof T]-?: {} extends Pick<T, K> ? K : never;
+}[keyof T];
```
Haroenv, line 3:
```suggestion
export type OptionalKeys<TObject> = {
  [TKey in keyof TObject]-?: {} extends Pick<TObject, TKey> ? TKey : never;
}[keyof TObject];
```

Haroenv: I thought we had an eslint rule that banned single-letter variables, surprised this wasn't flagged

Haroenv (commented):
✔️ once generics have more than one letter each 😄

packages/react-instantsearch-hooks-dom/src/types/WidgetProps.ts
```diff
@@ -0,0 +1,3 @@
+import type { OptionalKeys } from './OptionalKeys';
+
+export type WidgetProps<T> = Pick<T, OptionalKeys<T>>;
```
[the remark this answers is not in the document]
dhayab: `OptionalKeys<T>` doesn't return the filtered type, it just returns the optional keys of the type. This is then used in the `Pick<>` utility type to get the filtered type.

packages/react-instantsearch-hooks-dom/src/types/OptionalKeys.ts
dhayab: Just did a check and could not find a rule for that. [`id-length`](https://github.com/algolia/eslint-config-algolia/blob/master/packages/eslint-config-algolia/rules/base.js#L184) is turned off by default and does not account for type parameters anyway. I can have that behavior checked with a custom regex in `@typescript-eslint/naming-convention`:

```js
      {
        selector: 'typeParameter',
        format: ['PascalCase'],
        // prefix: ['T', 'K'],
        custom: {
          regex: '^[T|K].+',
          match: true,
        },
      },
```

Wdyt?

packages/react-instantsearch-hooks-dom/src/types/WidgetProps.ts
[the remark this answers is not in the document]
Haroenv: Ah yes I see. It may be combinable in a single type, as we probably don't want to expose these helper types, but good for me

Haroenv (dismissed):

francoischalifour (approved):

packages/react-instantsearch-hooks-dom/src/ui/InfiniteHits.tsx
```diff
@@ -4,16 +4,23 @@ import { cx } from './lib/cx';
...
-  onShowPrevious?: () => void;
+  onShowPrevious: (() => void) | undefined;
   onShowMore: () => void;
-  classNames?: Partial<InfiniteHitsClassNames>;
   translations: InfiniteHitsTranslations;
-};
+}>;
+type OptionalProps<THit> = Partial<{
+  classNames: Partial<InfiniteHitsClassNames>;
+  hitComponent: React.JSXElementConstructor<{ hit: THit }>;
```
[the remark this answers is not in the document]
dhayab: Since it is defaulted in the UI component below, it warranted an optional type declaration. Does marking it as a required prop brings other benefits?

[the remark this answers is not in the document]
francoischalifour: Indeed. That's argument against this solution. I think that this heuristic is too fragile and too easy to break without control.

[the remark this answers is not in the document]
Haroenv: I'd then vouch for separately exporting:

1. XXXComponentProps for all props
2. XXXComponentOverridableProps for the ones we use in the widget (name up for debate)

This still will be organised in mostly the same way, but of course reintroduces the problem of a prop being overridden by the widget, without it being removed from the type. That's really hard to solve so I guess it's ok. 

Does that work for you @francoischalifour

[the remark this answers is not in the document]
dhayab: The 2 exports works for me, that's similar to a previous implementation. As for naming, I can suggest `XXXComponentForwardedProps` or even `XXXComponentWidgetProps`.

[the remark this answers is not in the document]
Haroenv: actually we can leave out the `component` part, right? and `XXXProps` (the one if you use the ui component directly, passed to the arguments) and `XXXWidgetProps` for the ones overridable by the user

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
```diff
@@ -56,20 +56,25 @@ export type SearchBoxTranslations = {
...
     submitIconComponent?: React.JSXElementConstructor<IconProps>;
     loadingIconComponent?: React.JSXElementConstructor<IconProps>;
     classNames?: Partial<SearchBoxClassNames>;
+  };
+
+export type SearchBoxProps = SearchBoxWidgetProps &
+  Required<
+    Pick<React.ComponentProps<'form'>, 'onReset'> &
+      Pick<React.ComponentProps<'input'>, 'onChange'>
+  > & {
```
Haroenv, line 74: as these overlap, is the end result required or optional? are we able to use a similar pattern for infinite hits?

Haroenv (commented):
👍

packages/react-instantsearch-hooks-dom/src/ui/SearchBox.tsx
dhayab: They are required and omitted from `WrapperProps` with which they overlapped.

Haroenv: there's no omitting happening in this one, in InfiniteHits/Hits there is

dhayab: The definition on L59 seems to be working though:

```ts
type WrapperProps = Omit<React.ComponentProps<'div'>, 'onChange' | 'onReset'>;
```

The only overlap I see is with `onSubmit` which I'll fix but the other one are correctly marked as required from `SearchBoxProps`. Am I missing something?

Haroenv: 🤦 I didn't see that line! It seems good to go with for now then, but maybe with override keys or a similar utility it would be simpler to read?

https://github.com/algolia/react-instantsearch/blob/51e8818fce224819230c8bf6dea2a08d71d9be29/test/mock/createSearchClient.ts#L19-L21

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
```diff
@@ -47,12 +47,17 @@ type UseBreadcrumbRenderState = ReturnType<typeof useBreadcrumb>;
 
 type BreadcrumbItem = UseBreadcrumbRenderState['items'][number];
 
-export type BreadcrumbProps = React.HTMLAttributes<HTMLDivElement> &
+type WrapperProps = React.ComponentProps<'div'>;
+
+export type BreadcrumbWidgetProps = WrapperProps & {
```
francoischalifour, line 52: The UI components shouldn't be aware of widgets, even conceptually.

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
```diff
@@ -47,12 +47,17 @@ type UseBreadcrumbRenderState = ReturnType<typeof useBreadcrumb>;
...
+  classNames?: Partial<BreadcrumbClassNames>;
+};
+
+export type BreadcrumbProps = BreadcrumbWidgetProps &
```
francoischalifour, line 56: Why do we have all this indirection types and not this?

```tsx
export type BreadcrumbProps = React.ComponentProps<'div'> &
  Pick<UseBreadcrumbRenderState, 'items' | 'createURL'> & {
    hasItems: boolean;
    onNavigate: UseBreadcrumbRenderState['refine'];
    separator: string | undefined;
    translations: BreadcrumbTranslations;
    classNames?: Partial<BreadcrumbClassNames>;
  };
```

packages/react-instantsearch-hooks-dom/src/ui/InfiniteHits.tsx
```diff
@@ -4,17 +4,26 @@ import { cx } from './lib/cx';
...
   classNames?: Partial<InfiniteHitsClassNames>;
-  translations: InfiniteHitsTranslations;
+  hitComponent?: React.JSXElementConstructor<{ hit: THit }>;
 };
 
+export type InfiniteHitsProps<THit> = Omit<
+  InfiniteHitsWidgetProps<THit>,
+  'hitComponent'
+> &
+  Required<Pick<InfiniteHitsWidgetProps<THit>, 'hitComponent'>> & {
```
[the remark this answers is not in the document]
francoischalifour: The type could come from the UI component as optional and we use it as required in the widget. This is more respectful of the contract.

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
Haroenv: in the widget, how would you then decide which props to expose from the users' POV?

packages/react-instantsearch-hooks-dom/src/ui/InfiniteHits.tsx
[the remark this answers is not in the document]
Haroenv: According to your suggestion, it actually is required though

packages/react-instantsearch-hooks-dom/src/ui/Breadcrumb.tsx
dhayab: Could that be solved with a change of naming? Having types in a single place addresses an earlier feedback, making it less difficult to have correct typing for a component.

With a less specific name like `BreadcrumbForwardedProps` or something like that, we remove the vocabulary specific to widget, while retaining type consistency for the component. Wdyt?

packages/react-instantsearch-hooks-dom/src/ui/InfiniteHits.tsx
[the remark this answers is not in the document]
francoischalifour: I meant to other way around indeed. The point is that this layer shouldn't be aware of widgets.

francoischalifour:
Perhaps we can have a quick sync together to take a more informed decision?

dhayab:
> Perhaps we can have a quick sync together to take a more informed decision?

From our sync we have identified 3 types of structure:

- Omitting non-relevant props in widgets (which is basically what we currently have, see [SearchBox.ts](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx#L9))
- Duplicating types in widgets → https://github.com/algolia/react-instantsearch/pull/3458
- Composing types → pending

sarahdayan:
The current option (omitting irrelevant props) still looks the most straightforward to me.

I'm not sure the pros of these new options bring enough to the table. In my opinion, the two new options add needless complexity and hurt readability. One of the valuable aspects of our components and widgets in RISH is the minimal indirection we managed to have, and I think it's worth preserving.

Note that we don't write new components every day. Whenever we do, it will be different dynamics than us rushing to port everything to achieve feature parity within a couple of weeks. We'll likely have a sharper eye to spot typing issues, so tools like code review could be enough. wdyt?

Haroenv:
I personally think the option in https://github.com/algolia/react-instantsearch/pull/3459 is the best, as it enforces the types we expose to users are correct, also when a ui component would be used in multiple widgets, and also still avoids double writing every type and avoids the ui component to know about the widget.

dhayab:
I've thought about it as well, and one comment from @sarahdayan struck me as interesting (emphasis mine):

> In my opinion, the two new options add needless complexity and _hurt readability_.

It doesn't solve the single reason [the JIRA ticket](https://algolia.atlassian.net/browse/FX-1339) was created for:  improving readability.

I also agree that the third solution adds type safety, but as evoked earlier code review can be sufficient for the amount and frequency of changes that will be done here.

For those reasons I think we could step back and hold on to what we currently have. Would that be acceptable for you @Haroenv @francoischalifour ?

FabienMotte:
I would personally go for the first/current option for simplicity/readability.

Haroenv:
I don't think the existing code is more readable personally, as part of readability is making sure you don't need to think much. Now you need to manually make sure the list in `omit` is the same as the keys you override, which is something we found out multiple times only during code review or when another widget was written. We should use typescript for its strengths of correctness too IMO

PR #3446, opened 2022-04-25
feat(hooks): remove experimental warning

francoischalifour:
## Description

This removes the experimental warning in React InstantSearch Hooks, as well as the `suppressExperimentalWarning` prop on `<InstantSearch>`.

## Related

- https://github.com/algolia/doc/pull/6911

### DIFF
--- a/packages/react-instantsearch-hooks/src/lib/useInstantSearch.ts
+++ b/packages/react-instantsearch-hooks/src/lib/useInstantSearch.ts
@@ -19,21 +19,9 @@ const defaultUserAgents = [
   `react-instantsearch-hooks (${version})`,
 ];
 
-export type UseInstantSearchProps = InstantSearchOptions & {
-  /**
-   * Removes the console warning about the experimental version.
-   *
-   * Note that this warning is only displayed in development mode.
-   *
-   * @default false
-   */
-  suppressExperimentalWarning?: boolean;
-};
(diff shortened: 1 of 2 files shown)


sarahdayan (commented):
The `warn` import is no longer needed and is causing the pipeline to fail because of ESLint.

Haroenv (approved):

dhayab (approved):

packages/react-instantsearch-hooks/src/lib/useInstantSearch.ts
```diff
@@ -7,7 +7,6 @@ import version from '../version';
 
 import { useForceUpdate } from './useForceUpdate';
 import { useStableValue } from './useStableValue';
-import { warn } from './warn';
```
Haroenv, line 10: this was the only use of warn, it can be removed, unless we plan on having it again for another reason

Haroenv (approved):

packages/react-instantsearch-hooks/src/lib/useInstantSearch.ts
francoischalifour: I think we can keep it since it gets out of the final bundle anyway. It'll likely be useful for other purposes.

sarahdayan (approved):

PR #3455, opened 2022-05-02
fix(hooks): prevent infinite loops from render state

francoischalifour:
This optimizes when we update the React state in `useConnector()` to skip function reference changes in the render state. This lets us and users create unstable function references in `getWidgetRenderState()` (our own `connectSearchBox()` doesn't expose stable functions).

This **brings support for inline functions used in widgets**, like users are used to in RIS v6.

```jsx
function App(props) {
  const [favorites, setFavorites] = useState([]);

  return (
    <InstantSearch {...props}>
      <Hits
        hitComponent={Hit}
        transformItems={(hits) =>
          hits.map((hit) => ({
            ...hit,
            isFavorite: favorites.find((x) => x.objectID === hit.objectID),
          }))
        }
      />
    </InstantSearch>
  );
}
```

Prior to this change, **such an app would be prone to trigger an infinite loop**. Now, it triggers only two renders. In the future, we might be able to optimize it further to trigger a single render.

When using the Hook, users still need to stabilize their function references.

```jsx
function Search() {
  const [favorites, setFavorites] = useState([]);
  const transformItems = useCallback((items) => items, [favorites]);
  const { hits } = useHits({ transformItems });

  return <>{/* ... */}</>;
}

function App(props) {
  const [favorites, setFavorites] = useState([]);
  const transformItems = useCallback((items) => items, [favorites]);
  const { hits } = useHits({ transformItems });

  return (
    <InstantSearch {...props}>
      <Search />
    </InstantSearch>
  );
}
```

### DIFF
--- a/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
+++ b/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
@@ -1,3 +1,4 @@
+import { dequal } from 'dequal/lite';
 import { useMemo, useRef, useState } from 'react';
 
 import { createSearchResults } from '../lib/createSearchResults';
@@ -27,6 +28,9 @@ export function useConnector<
     additionalWidgetProperties
   );
   const shouldSetStateRef = useRef(true);
+  const previousRenderStateRef = useRef<TDescription['renderState'] | null>(
+    null
+  );
 
   const widget = useMemo(() => {
     const createWidget = connector(
@@ -54,8 +58,29 @@ export function useConnector<
           const { instantSearchInstance, widgetParams, ...renderState } =
             connectorState;
 
-          // eslint-disable-next-line @typescript-eslint/no-use-before-define
-          setState(renderState);
+          const renderStateWithoutFns = Object.entries(renderState).reduce(
+            (acc, [key, value]) => {
+              if (typeof value === 'function') {
+                return acc;
+              }
+
+              acc[key] = value;
+
+              return acc;
+            },
+            {}
+          );
+
+          // We only update the state if the widget render state has changed.
+          // This avoids infinite loops when a function prop reference changes.
+          // We base the check only on non-function parameters because functions
+          // cannot be compared. It's safe to omit them because they get updated
+          // every time another render param changes.
+          if (!dequal(renderStateWithoutFns, previousRenderStateRef.current)) {
+            // eslint-disable-next-line @typescript-eslint/no-use-before-define
+            setState(renderState);
+            previousRenderStateRef.current = renderStateWithoutFns;
+          }
         }
       },
       () => {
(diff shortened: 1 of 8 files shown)


Haroenv (approved):
it's too bad I can't find where in dequal's code a "functions are always identical" would live. I'd prefer that over this filtering here, which results in a few extra iterations. At the end of the day, this should work, as long as non-stable functions only use references that can't get stale

francoischalifour:
@Haroenv Here's an [alternative solution patching `dequal` to use a compare function](https://github.com/algolia/react-instantsearch/compare/feat/support-unstable-fns-from-render-state-dequal).

This has the benefit of generalizing the solution and not looping over the whole render state at every state change. I think I'm more in favor of that one, which would need more tests to propose as a PR.

Wdyt?

Haroenv:
I think it's an interesting option to fork dequal, but then we need to make sure it's used everywhere of course. 

Any reason you went with comparing the contstructor vs. typeof? 

At the end of the day, I'm ok with either the current solution or the one proposed, as they both have tradeoffs @francoischalifour

dhayab:
@francoischalifour these are 2 great options! I'm more in favor of the dequal fork since it doesn't require preparing a filtered version of the render state.

francoischalifour:
@Haroenv I went with constructor because [that's what dequal is doing](https://github.com/lukeed/dequal/blob/9aa73181ac7e081cd330cac67d313632ac04bb02/src/lite.js) for the other data types. I [saw online](https://www.measurethat.net/Benchmarks/Show/5536/1/check-function-typeof-vs-constructor-null-check) that `typeof` is faster though, so I'm not sure what's best here.

Haroenv:
I think the constructor comparison is faster as they're comparing both, so don't need to call typeof twice (and Date for example has a different constructor than function). In our case we do want to know when things are functions in the most simple way, so I'd go with that approach :)

francoischalifour:
In 4042731a57f368de8e63bfb05c831187750b22db I use the change I linked above.

- We now fork `dequal` to be able to use a compare function
- We disallow importing the `dequal` package directly with an ESLint rule
- We don't have code specific to removing functions from the render state in `useConnector()`, we skip their reference check with the compare function

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
```diff
@@ -58,28 +58,21 @@ export function useConnector<
...
+          // except for functions. We ignore function reference changes to avoid
+          // infinite loops. It's safe to omit them because they get updated
           // every time another render param changes.
-          if (!dequal(renderStateWithoutFns, previousRenderStateRef.current)) {
+          if (
+            !dequal(
+              renderState,
+              previousRenderStateRef.current,
+              (a, b) =>
+                a?.constructor === Function && b?.constructor === Function
```
Haroenv, line 70:
```suggestion
                typeof a === "function"
```

could we shortcut here, as we know you wouldn't switch from function to something else? or is that too risky?

dhayab (approved):
👏

francoischalifour:
In 9b8975b75de2f6c5578293dffb65135eb448fca3 I remove the `dequal` dependency.

PR #3461, opened 2022-05-06
fix(numericmenu): include range values in comparison with minmax bounds

dhayab:
**Summary**

This PR addresses an issue encountered by a customer, where an item on a `NumericMenu` having its `start` value set to the maximum value of a facet would wrongly be disabled in the widget.

Fixes [FX-1376](https://algolia.atlassian.net/browse/FX-1376).

**Result**

- The items are now correctly enabled/disabled
- Tests are updated to include these specific cases

### DIFF
--- a/packages/react-instantsearch-core/src/connectors/__tests__/connectNumericMenu.js
+++ b/packages/react-instantsearch-core/src/connectors/__tests__/connectNumericMenu.js
@@ -268,6 +268,8 @@ describe('connectNumericMenu', () => {
             { label: '2', start: 400 },
             { label: '3', end: 200 },
             { label: '4', start: 100, end: 200 },
+            { label: '5', start: 300 },
+            { label: '6', start: 300, end: 300 },
           ],
           contextValue,
         },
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-core/src/connectors/__tests__/connectNumericMenu.js
```diff
@@ -268,6 +268,8 @@ describe('connectNumericMenu', () => {
             { label: '2', start: 400 },
             { label: '3', end: 200 },
             { label: '4', start: 100, end: 200 },
+            { label: '5', start: 300 },
+            { label: '6', start: 300, end: 300 },
```
Haroenv, line 272: I think I'd rather see a separate test for this too, so it can easily be ported to InstantSearch.js (where it already works)

dhayab: This test specifically verifies the `noRefinement` value for each item range, which is not implemented in IS.js, so I don't see how it could be ported. Do you have something specific in mind?

Haroenv (approved):
misread the test, lgtm!

PR #3471, opened 2022-05-13
fix(hooks-highlight): make sure highlight and snippet don't show html-escaped content

Haroenv:
**Summary**

instantsearch.js escapes the html and replaces the original tags with <mark> in Hits. This is not desirable in React as it doesn't decode literal html literals.

**Result**

apostrophe, quote, lesser and greater signs show up as authored in the index. 

https://algolia.atlassian.net/browse/CR-1142


dhayab (approved):
Good catch

PR #3472, opened 2022-05-16
feat(react-instantsearch-core): allow widgets to set their $$widgetType

Haroenv:
**Summary**

This is is used in metadata to distinguish between a built-in component and a custom one.

This commit only shows how it could be implemented in HierarchicalMenu, not yet the tests / other widgets to open the conversation

FX-917

**Result**

widgets created by createConnector accept an optional second argument that has `{ $$widgetType: string }` to specify the identifier for the widget


packages/react-instantsearch-core/src/core/createConnector.tsx
```diff
@@ -351,8 +356,15 @@ export function createConnectorWithoutContext(
 }
 
 const createConnectorWithContext =
-  (connectorDesc: ConnectorDescription) => (Composed: ReactType) => {
-    const Connector = createConnectorWithoutContext(connectorDesc)(Composed);
+  (connectorDesc: ConnectorDescription) =>
+  (
+    Composed: ReactType,
+    meta: WidgetMeta = { $$widgetType: 'custom-widget' }
```
Haroenv, line 362: this is debatable, whether we either give a string default or an explicit null. I think a non-optional that allows string or undefined can be interesting too

dhayab: Can we get usage as easily when `$$widgetType` is undefined? If so, I think it's fine if we don't default it.

sarahdayan: I'd be in favor of requiring to fill it in, but if we do this, anyone using `createConnector` already and updating the library will have a TypeScript violation. That's arguably breaking.

Haroenv: we can have it required internally, and still discuss on DefinitelyTyped what the type for externally is

packages/react-instantsearch-dom/src/widgets/MenuSelect.js
```diff
@@ -46,4 +46,4 @@ const MenuSelectWidget = (props) => (
   </PanelCallbackHandler>
 );
 
-export default connectMenu(MenuSelectWidget);
+export default connectMenu(MenuSelectWidget, { $$widgetType: 'ais.menu' });
```
dhayab, line 49: I have `ais.menuSelect` in IS.js. Do we want to merge it with the menu widget?

```suggestion
export default connectMenu(MenuSelectWidget, { $$widgetType: 'ais.menuSelect' });
```

packages/react-instantsearch-dom/src/widgets/QueryRuleCustomData.tsx
```diff
@@ -13,4 +13,6 @@ const QueryRuleCustomDataWidget: React.FC<
   </PanelCallbackHandler>
 );
 
-export default connectQueryRules(QueryRuleCustomDataWidget);
+export default connectQueryRules(QueryRuleCustomDataWidget, {
+  $$widgetType: 'ais.queryRules',
```
dhayab, line 17:
```suggestion
  $$widgetType: 'ais.queryRuleCustomData',
```

packages/react-instantsearch-dom/src/widgets/RatingMenu.js
```diff
@@ -57,4 +57,4 @@ const RatingMenuWidget = (props) => (
   </PanelCallbackHandler>
 );
 
-export default connectRange(RatingMenuWidget);
+export default connectRange(RatingMenuWidget, { $$widgetType: 'ais.range' });
```
dhayab, line 60:
```suggestion
export default connectRange(RatingMenuWidget, { $$widgetType: 'ais.ratingMenu' });
```

dhayab (changes requested):
A few widgets have unexpected values for `$$widgetType`.

dhayab (approved):

PR #3475, opened 2022-05-17
test(server): retry module tests

Haroenv:
**Summary**

These tests are flaky, as sometimes Jest chooses to cache the modules accross tests, even if they shouldn't be. isolatedModules or resetModules could help, but then react doesn't work anymore https://github.com/facebook/jest/issues/8987

**Result**

tests pass consistently

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,22 @@ jobs:
           command: yarn type-check
       - run:
           name: Unit Tests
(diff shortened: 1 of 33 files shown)


Haroenv:
ugh, still flaky, and not sure if it actually retried? I feel like it's something cache-related

dhayab (approved):
😍

PR #3481, opened 2022-05-25
fix(types): support React 18 types

francoischalifour:
> **Note**
> The CI is expected to fail, only the [**Type checking**](https://app.circleci.com/pipelines/github/algolia/react-instantsearch/2842/workflows/4d48e58b-30e1-438b-8b6d-c3540483d5e5/jobs/55175/parallel-runs/0/steps/0-108) workflow matters.

## Description

This updates the types in our codebase to support React 18.

## Related

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210
- [JIRA ticket](https://algolia.atlassian.net/browse/FX-1451)

### DIFF
--- a/packages/react-instantsearch-core/src/widgets/Index.tsx
+++ b/packages/react-instantsearch-core/src/widgets/Index.tsx
@@ -12,6 +12,7 @@ function getIndexContext(props: Props): IndexContext {
 type Props = {
   indexName: string;
   indexId: string;
+  children?: React.ReactNode;
 };
 
 type InnerProps = Props & { contextValue: InstantSearchContext };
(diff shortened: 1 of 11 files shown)


packages/react-instantsearch-core/src/widgets/Index.tsx
```diff
@@ -12,6 +12,7 @@ function getIndexContext(props: Props): IndexContext {
 type Props = {
   indexName: string;
   indexId: string;
+  children?: React.ReactNode;
 };
```
dhayab, line 16: An alternative would be:

```suggestion
type Props = React.PropsWithChildren<{
  indexName: string;
  indexId: string;
}>;
```

I don't know if one way is better than the other, though, so feel free to disregard if you're satisfied with the current type definition.

dhayab (approved):
👍

FabienMotte (approved):

PR #3484, opened 2022-05-27
docs(github): add React version in issue template

francoischalifour:
This adds the React version in the GitHub issue template to better track on what React version the InstantSearch bug appears. (We're seeing more bugs happen on React 18 only.)


FabienMotte (approved):
Good idea 👍

Issue #3482, opened 2022-05-26
Server-rendered, Controlled, InstantSearch fails after re-render


bgschiller:
## 🐛 Bug description

We have set up a server-rendered instant search, including the `onStateChange` prop making it a controlled component. When the parent component re-renders, any state change will fail with the error:

> Error: The `start` method needs to be called before `setUiState`.

## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Clone down https://github.com/bgschiller/instantsearch-rerender-bug `git clone git@github.com:bgschiller/instantsearch-rerender-bug.git`
2. Set environment variables for `ALGOLIA_APP_ID` and `ALGOLIA_SEARCH_KEY`. I'm happy to provide mine if you need, as they're just for a demo environment for us.
3. `npm install` and `npm run dev`. Visit http://localhost:3000
4. In any order, click the "rerender" button and type something in the search box.
5. See error "Error: The `start` method needs to be called before `setUiState`."

**Live reproduction:**

https://codesandbox.io/s/gallant-banach-2zbyil

## 💭 Expected behavior

I expected this to work without error

## 🖥 Screenshots

![a traceback starting with "Error: The `start` method needs to be called before `setUiState`."](https://clip.brianschiller.com/efGbz0x-2022-05-26.png)

## Environment

- OS: verified on Windows, Ubuntu, and codesandbox
- Browser: Chrome
- Version: react-instantsearch-hooks-{web,server} 6.26.0

## Additional context

I came across this issue in the course of my work at DEPT, but I'll be leaving that job soon. I'm tagging @wintersieck to answer any follow-up questions.

francoischalifour:
Thank you the bug report details and reproduction. I was able to reproduce this bug in an isolated test environment as well.

chan-fullstack:
Same issue I'm encountering when closing a modal through `next/router` change. It seems like the InstantSearch instance becomes null after re-render.
![image](https://user-images.githubusercontent.com/61461187/171844955-c84cfdf8-55b1-44b2-8745-5ed80ca9bd5a.png)

francoischalifour:
@bgschiller Migrating our internals with [`useSyncExternalStore()`](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) in #3489 fixed your original issue (this is not released yet).

[...]

PR #3485, opened 2022-05-27
test(hooks): use widgets in `<InstantSearchSSRProvider>` tests

francoischalifour:
This refactors the tests for `<InstantSearchSSRProvider>` to use our widgets.

This is a preparation of [a test](https://github.com/algolia/react-instantsearch/commit/63ad30fd76ee776939f037cdda0951c2d60aa30b) that shows the bug reported in #3482, to use our actual `<SearchBox>` widget, closer to a real use case.


Haroenv (approved):

Haroenv:
I'm ok with the React InstantSearch hooks core tests using React InstantSearch hooks web as it's simpler and more realistic, but I wonder if they maybe should live in React InstantSearch serve or React InstantSearch hooks web packages, what do you think?

francoischalifour:
Since the `<InstantSearchSSRProvider>` tests are in `-hooks` I think we can leave this test here.

PR #3489, opened 2022-05-31
feat(hooks): migrate to `useSyncExternalStore()`

francoischalifour:
> **Note**
> This is in the PR series that brings support for React 18.

## Description

This updates the implementation in `<InstantSearch>` from an effect to the [`useSyncExternalStore()`](https://reactjs.org/docs/hooks-reference.html#usesyncexternalstore) API. This Hook is meant to support new concurrent and time slicing features of React 18.

This is using a shim that backports the behavior to all React versions from 16.8.0 (the one that provides Hooks).

Together with algolia/instantsearch.js#5056, we'll should have the certainty to support concurrent features with a minimum amount of network requests.

## Related

- algolia/instantsearch.js#5056

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -50,6 +50,8 @@ module.exports = (api) => {
             'instantsearch.js',
             // React-DOM also fails if the paths are incomplete
             'react-dom',
+            // `use-sync-external-store` also fails if the paths are incomplete
+            'use-sync-external-store/shim',
           ],
         },
       ],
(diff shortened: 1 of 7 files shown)


babel.config.js
```diff
@@ -50,6 +50,8 @@ module.exports = (api) => {
             'instantsearch.js',
             // React-DOM also fails if the paths are incomplete
             'react-dom',
+            // `use-sync-external-store` also fails if the paths are incomplete
+            'use-sync-external-store/shim',
```
Haroenv, line 54:
```suggestion
            'use-sync-external-store',
```

I think it may just be like this

PR #3494, opened 2022-06-01
feat(hooks): introduce `useInstantSearch()`

Haroenv:
**Summary**

useInstantSearch is a new hook to access advanced functionality of InstantSearch without creating custom widgets:

- access to complete search results
- access to complete ui state
- functions to change the ui state
- refreshing the search results
- using middleware

**Result**

- implements useInstantSearch
- implements two underlying, non-exposed connectors for search results and ui state
- uses useInstantSearch for a refresh button in the example


packages/react-instantsearch-hooks/src/hooks/useInstantSearch.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return {
+    results,
+    scopedResults,
+    uiState,
+    setUiState,
+    indexUiState,
+    setIndexUiState,
+    use,
+    refresh: () => search.refresh(),
```
francoischalifour, line 53: We should expose a stable reference for this function.

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,396 @@
...
+
+      await wait(0);
+
+      expect(container.innerHTML).toMatchInlineSnapshot(
+        `"<button type=\\"button\\">hey</button>hey"`
+      );
+    });
+  });
+
+  describe('middleware', () => {
```
francoischalifour, line 262: Can we add a test with a real use case where you use `use()` in a component, to make sure that the middleware subscribes from the start, and doesn't miss any initial events?

packages/react-instantsearch-hooks/src/hooks/useInstantSearch.ts
```diff
@@ -0,0 +1,56 @@
...
+type InstantSearchApi = {
+  scopedResults: ScopedResult[];
+  results: SearchResults<any>;
+  uiState: UiState;
+  setUiState: InstantSearch['setUiState'];
+  indexUiState: IndexUiState;
+  setIndexUiState: (indexUiState: IndexUiState) => void;
+  use: (...middlewares: Middleware[]) => void;
+  refresh: InstantSearch['refresh'];
+  search: InstantSearch;
```
francoischalifour, line 25: We don't want to provide access to `search` in this API until it's proven to be useful, which it shouldn't.

Haroenv: I feel like there's other keys that don't need to be "exposed as much", but can still be useful, like `search.renderState` for accessing refinement functions without mounting that widget again. If we feel very strongly it can be removed again, but I saw the use in it

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchState.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+      wrapper: ({ children }) =>
+        wrapper({
+          children: (
+            <>
+              <SearchBox />
+              {children}
+            </>
+          ),
+        }),
+    });
```
francoischalifour, line 70: The `<SearchBox>` is not needed in this test.

```suggestion
    const { result, waitForNextUpdate } = renderHook(() => useSearchState(), {
      wrapper: ({ children }) => wrapper({ children }),
    });
```

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchState.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+    );
+
+    userEvent.click(container.querySelector('button')!);
+
+    await wait(0);
+
+    expect(container.innerHTML).toMatchInlineSnapshot(
+      `"<button type=\\"button\\">hey</button>hey"`
+    );
+  });
```
francoischalifour, line 133: The test is easier to read this way, there's no escaping and we understand the intent of clicking the button.

```suggestion
  test('returns a function to modify the whole state', async () => {
    function App() {
      const { uiState, setUiState } = useSearchState();

      return (
        <>
          <button
            type="button"
            onClick={() => {
              setUiState({ indexName: { query: 'new query' } });
            }}
          >
            {uiState.indexName.query}
          </button>
          <SearchBox />
        </>
      );
    }

    const { getByRole } = render(
      <InstantSearchHooksTestWrapper>
        <App />
      </InstantSearchHooksTestWrapper>
    );
    const button = getByRole('button');

    await wait(0);

    expect(button).toHaveTextContent('');

    userEvent.click(button);

    await wait(0);

    expect(button).toHaveTextContent('new query');
  });
```

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchState.test.tsx
```diff
@@ -0,0 +1,174 @@
...
```
francoischalifour, line 173:
```suggestion
  test('returns a function to modify the index state', async () => {
    function App() {
      const { indexUiState, setIndexUiState } = useSearchState();

      return (
        <>
          <button
            type="button"
            onClick={() => {
              setIndexUiState({ query: 'new query' });
            }}
          >
            {indexUiState.query}
          </button>
          <SearchBox />
        </>
      );
    }

    const { getByRole } = render(
      <InstantSearchHooksTestWrapper>
        <App />
      </InstantSearchHooksTestWrapper>
    );
    const button = getByRole('button');

    await wait(0);

    expect(button).toHaveTextContent('');

    userEvent.click(button);

    await wait(0);

    expect(button).toHaveTextContent('new query');
  });
```

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchResults.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+      results: expect.any(SearchResults),
+      scopedResults: [
+        expect.objectContaining({
+          helper: expect.any(AlgoliaSearchHelper),
+          indexId: 'indexName',
+          results: expect.any(SearchResults),
+        }),
+      ],
+    });
+
```
francoischalifour, line 25: We can also add this check, because we'll use this to detect the "no results" scenario.

```suggestion
    expect(result.current.results.__isArtificial).toEqual(true);

```

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchResults.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+    expect(result.current).toEqual({
+      results: expect.any(SearchResults),
+      scopedResults: [
+        expect.objectContaining({
+          helper: expect.any(AlgoliaSearchHelper),
+          indexId: 'indexName',
+          results: expect.any(SearchResults),
+        }),
+      ],
+    });
```
francoischalifour, line 38:
```suggestion
    });
    expect(result.current.results.__isArtificial).toBeUndefined();
```

packages/react-instantsearch-hooks/src/hooks/useInstantSearch.ts
```diff
@@ -0,0 +1,60 @@
...
+} from 'instantsearch.js';
+
+type InstantSearchApi = {
+  scopedResults: ScopedResult[];
+  results: SearchResults<any>;
+  uiState: UiState;
+  setUiState: InstantSearch['setUiState'];
+  indexUiState: IndexUiState;
+  setIndexUiState: (indexUiState: IndexUiState) => void;
+  use: (...middlewares: Middleware[]) => void;
```
francoischalifour, line 23:
```suggestion
  use: (...middlewares: Middleware[]) => () => void;
```

packages/react-instantsearch-hooks/src/hooks/useInstantSearch.ts
```diff
@@ -0,0 +1,60 @@
...
+  search: InstantSearch;
+};
+
+export function useInstantSearch(): InstantSearchApi {
+  const search = useInstantSearchContext();
+  const { uiState, setUiState, indexUiState, setIndexUiState } =
+    useSearchState();
+  const { results, scopedResults } = useSearchResults();
+
+  const use = useCallback(
```
francoischalifour, line 34:
```suggestion
  const use: InstantSearchApi['use'] = useCallback(
```

packages/react-instantsearch-hooks/src/hooks/useInstantSearch.ts
```diff
@@ -0,0 +1,60 @@
...
+      search.use(...middlewares);
+
+      return () => {
+        search.unuse(...middlewares);
+      };
+    },
+    [search]
+  );
+
+  const refresh = useCallback(() => {
```
francoischalifour, line 45:
```suggestion
  const refresh: InstantSearchApi['refresh'] = useCallback(() => {
```

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+        'scopedResults',
+        'uiState',
+        'setUiState',
+        'indexUiState',
+        'setIndexUiState',
+        'use',
+        'refresh',
+        'search',
+      ]);
+    });
```
francoischalifour, line 51: Do you think this has value on top of our typings?

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+            wrapper({
+              children: (
+                <>
+                  <SearchBox />
+                  {children}
+                </>
+              ),
+            }),
+        }
+      );
```
francoischalifour, line 153:
```suggestion
      const { result, waitForNextUpdate } = renderHook(
        () => useInstantSearch(),
        {
          wrapper: ({ children }) => wrapper({ children }),
        }
      );
```

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+            indexName: { query: 'iphone' },
+          },
+          indexUiState: { query: 'iphone' },
+          setUiState: expect.any(Function),
+          setIndexUiState: expect.any(Function),
+        })
+      );
+    });
+
+    test('returns a function to modify the whole state', async () => {
```
francoischalifour, line 182: Same suggestion as in the `useSearchState()` test.

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+      userEvent.click(container.querySelector('button')!);
+
+      await wait(0);
+
+      expect(container.innerHTML).toMatchInlineSnapshot(
+        `"<button type=\\"button\\">hey</button>hey"`
+      );
+    });
+
+    test('returns a function to modify the index state', async () => {
```
francoischalifour, line 222: Same suggestion as in the `useSearchState()` test.

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+      await wait(0);
+
+      expect(container.innerHTML).toMatchInlineSnapshot(
+        `"<button type=\\"button\\">hey</button>hey"`
+      );
+    });
+  });
+
+  describe('middleware', () => {
+    test('gives access use function', async () => {
```
francoischalifour, line 264:
```suggestion
    test('gives access to the use function', async () => {
```

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+      expect(subscribe).toHaveBeenCalledTimes(1);
+      expect(onStateChange).toHaveBeenCalledTimes(1);
+      // unsubscribe is first called by the parent InstantSearch unmounting
+      // 🚨 dispose doesn't remove middleware (because otherwise routing would break)
+      // then unuse is called by this widget itself unmounting.
+      // if only the component with useInstantSearch is unmounted, unsubscribe is called once.
+      expect(unsubscribe).toHaveBeenCalledTimes(2);
+
+      // middleware was removed
+      expect(result.current.middleware).toEqual([]);
```
francoischalifour, line 338: This is testing InstantSearch.js, not the public API that it exposes. So I don't think we should have it here.

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+    });
+
+    test('provides a stable reference', async () => {
+      const wrapper = createInstantSearchTestWrapper();
+      const { result, waitForNextUpdate, rerender } = renderHook(
+        () => useInstantSearch(),
+        { wrapper }
+      );
+
+      expect(result.current.refresh).toEqual(expect.any(Function));
```
francoischalifour, line 384:
```suggestion
      expect(result.current.refresh).toBeInstanceOf(Function);
```

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+
+      const ref = result.current.refresh;
+
+      await waitForNextUpdate();
+
+      // reference has not changed
+      expect(result.current.refresh).toEqual(ref);
```
francoischalifour, line 391: We want to test the reference equality.

```suggestion
      expect(result.current.refresh).toBe(ref);
```

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+
+      rerender();
+
+      // reference has not changed
+      expect(result.current.refresh).toEqual(ref);
```
francoischalifour, line 396:
```suggestion
      expect(result.current.refresh).toBe(ref);
```

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+    });
+  });
+
+  describe('other instantsearch access', () => {
```
francoischalifour, line 400: We can remove this once we remove the `search` access from the API.

francoischalifour (commented):
We can rename `src/lib/useInstantSearch.ts` to `src/lib/useInstantSearchApi.ts` (as well as the Hook name) to avoid confusion with the new `useInstantSearch()` public API.

examples/hooks/components/Refresh.css
```diff
@@ -0,0 +1,27 @@
...
+  line-height: 2rem;
+  cursor: pointer;
+}
+
+.Refresh:hover {
+  background: linear-gradient(-180deg, #fff, #f5f5fa);
+}
+
+.Refresh svg {
+  transition: transform .5s ease;
```
francoischalifour, line 22: This has a smoother ending.

```suggestion
  transition: transform .5s ease-out;
```

examples/hooks/components/Refresh.css
```diff
@@ -0,0 +1,27 @@
...
+  font-family: inherit;
+  font-size: 0.875rem;
+  color: #23263b;
+  background: linear-gradient(-180deg, #fff, #fcfcfd);
+  border: 1px solid #d6d6e7;
+  border-radius: 3px;
+  box-shadow: 0 1px 0 0 rgb(35 38 59 / 5%);
+  text-align: center;
+  line-height: 2rem;
+  cursor: pointer;
```
francoischalifour, line 14:
```suggestion
  line-height: 2rem;
  cursor: pointer;
  align-items: center;
  justify-content: center;
  display: flex;
```

RobbyUitbeijerse:
> What I wanted to put the light on in https://github.com/algolia/react-instantsearch/pull/3494#discussion_r886660022 is that it's not possible to recreate a router from the use() API because the middleware seems to be added after the first render.

@Haroenv might be interesting to provide initial middleware like so?

```
const {} = useInstantSearch({ 
  use: [middleware]
})
```

similar to how it's been implemented in swr:
https://swr.vercel.app/docs/middleware

Haroenv:
Sounds interesting, but I wonder if those also would be added too late. Maybe it should be a prop to the InstantSearch component instead of a use function?

francoischalifour:
@RobbyUitbeijerse @Haroenv Thinking back at it, it makes sense that we cannot recreate a complete router from `use()`, and this is because InstantSearch.js [treats routing as a special middleware internally](https://github.com/algolia/instantsearch.js/blob/0a517609de103eef4f8edfefe6e28a1d79a14209/src/lib/InstantSearch.ts#L267-L270), which subscribes before `start()`. So this is not a blocker for `useInstantSearch().use`. The `routing` prop would be the way to plug a routing system.

A more conventional middleware that I tried to reimplement is the [Insights middleware](https://www.algolia.com/doc/api-reference/widgets/insights/js/) with `<InsightsMiddleware>` in [this sandbox](https://codesandbox.io/s/react-instantsearch-hooks-insights-middleware-o5yrtl?file=/InsightsMiddleware.tsx). You can see that from the `useInstantSearch().use` standpoint, it works correctly.

So to me this is enough to validate this API.

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchState.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+      return (
+        <>
+          <button
+            type="button"
+            onClick={() => {
+              setUiState({ indexName: { query: 'new query' } });
+            }}
+          >
+            {uiState.indexName.query}
```
francoischalifour, line 99: Could we assert on the whole `uiState` to make sure to have the correct object?

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchState.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+      return (
+        <>
+          <button
+            type="button"
+            onClick={() => {
+              setIndexUiState({ query: 'new query' });
+            }}
+          >
+            {indexUiState.query}
```
francoischalifour, line 136: Could we assert on the whole `uiState` to make sure to have the correct object?

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -0,0 +1,341 @@
...
+      unmount();
+
+      expect(middleware).toHaveBeenCalledTimes(1);
+      expect(subscribe).toHaveBeenCalledTimes(1);
+      expect(onStateChange).toHaveBeenCalledTimes(1);
+      // unsubscribe is first called by the parent InstantSearch unmounting
+      // 🚨 dispose doesn't remove middleware (because otherwise routing would break)
+      // then unuse is called by this widget itself unmounting.
+      // if only the component with useInstantSearch is unmounted, unsubscribe is called once.
+      expect(unsubscribe).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
Haroenv: is the addition of this line sufficient? https://github.com/algolia/react-instantsearch/pull/3494/commits/8ece99ee66a0dcb35f68f7f1e52fe0fffe3fb157#diff-1514cc8cd3c08b8855382d2d168f136e3f1bff8ec79130f262cfb1c360b43946R280

packages/react-instantsearch-hooks/src/lib/useSearchState.ts
```diff
@@ -0,0 +1,74 @@
...
+          setIndexUiState = (nextUiState) => {
+            setUiState((prevUiState) => ({
+              ...prevUiState,
+              [indexId]:
+                typeof nextUiState === 'function'
+                  ? nextUiState(prevUiState[indexId])
+                  : nextUiState,
+            }));
+          };
+        }
```
dhayab, line 48: Are there advantages in defining those setters in the `init()` method of the connector, instead of doing this conditionally here?

dhayab (approved):

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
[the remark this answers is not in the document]
francoischalifour: Yes

francoischalifour (approved):
Great API addition!

packages/react-instantsearch-hooks/src/lib/useSearchState.ts
Haroenv: init gets called after getWidgetRenderState in some cases :)

PR #3500, opened 2022-06-06
test(hooks): add test for re-renders

francoischalifour:
This adds several test cases that should cover #3482 by re-rendering an InstantSearch app in multiple scenarios.

The test in `<InstantSearchSSRProvider>` covers the initial post in #3482. The `<InstantSearch>` tests already worked before introducing `useSyncExternalStore()` in #3489, but still makes sure that these scenarios work.


Haroenv (approved):
even though we want to change this, it's good to have the tests that will be changed at a later point

PR #3502, opened 2022-06-06
feat(hooks): upgrade to InstantSearch.js 4.41.0

francoischalifour:
This upgrades the InstantSearch.js to [v4.41.0](https://github.com/algolia/instantsearch.js/releases/tag/v4.41.0) in the Hooks package.

This includes algolia/instantsearch.js#5056 which skips the initial search request to only rely on the one triggered by `addWidgets()`. This eliminates problems in React 18 where effects in `useInstantSearch()` and `useConnector()` get executed one after the other, triggering sometimes 2 network requests.

### DIFF
--- a/packages/react-instantsearch-hooks/package.json
+++ b/packages/react-instantsearch-hooks/package.json
@@ -47,8 +47,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch-helper": "^3.7.4",
-    "instantsearch.js": "^4.40.1",
+    "algoliasearch-helper": "^3.8.0",
+    "instantsearch.js": "^4.41.0",
     "use-sync-external-store": "^1.0.0"
   },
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch-hooks/package.json
```diff
@@ -47,8 +47,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "algoliasearch-helper": "^3.7.4",
-    "instantsearch.js": "^4.40.1",
+    "algoliasearch-helper": "^3.8.0",
```
Haroenv, line 50: .2 please, .0 is broken/deprecated

francoischalifour: I updated the peer deps for all packages in f6031759c30f7aa0cbf363b1bf305b44663c3511 then.

Haroenv (approved):

PR #3514, opened 2022-06-13
feat(hooks): support React 18 Strict Mode

francoischalifour:
This brings **support for React 18's Strict Mode** in React InstantSearch Hooks.

## Implementation

The code that supports Strict Mode is in `useWidget()`, shared across `useConnector()` and `<Index>`. We now need to break the scenario of a widget render in two:

(1) **The widget is initially rendered**. We simply call `addWidgets()`, and schedule a call to `removeWidets()`. We don't call `removeWidgets()` right away anymore, because we want to be able to cancel it in the next effect execution. This lets us minimize the amount of remove/add widgets and save the routing mechanism.

(2) **The widget is rerendered or updated**. In that case, we cancel the `removeWidget()` scheduled call because we don't want to wait for the scheduled call to finish to execute our logic. We then check if the props are equal. If they are, it means that it's just a rerender, and we stop the effect execution. If props aren't equal, it means that the widget needs to be updated, so we call `removeWidget()` right away, and directly call `addWidgets()`.

So the main change is that **the effect cleanup function is now scheduled, and not called immediately**. When we need to run it immediately, we cancel the scheduled one and call it right away.

## Test changes

- I refactored all the lifecycle tests to run in `<StrictMode>` to make sure that the lib runs exactly as expected in Strict Mode.
- I refactored the custom InstantSearch listeners in the tests to an `<InstantSearchSpy>` component. It lets us track what underlying InstantSearch.js method is called during the lifecycle.

> 💡 **Tip**
> Click "Hide whitespaces" on GitHub to minimize review changes in the tests.


packages/react-instantsearch-hooks/src/components/__tests__/Index.test.tsx
```diff
@@ -29,9 +30,11 @@ describe('Index', () => {
     const searchClient = createSearchClient({});
 
     const { container } = render(
-      <InstantSearch indexName="indexName" searchClient={searchClient}>
-        <Index indexName="childIndex">Children</Index>
-      </InstantSearch>
+      <StrictMode>
```
Haroenv, line 33: do the tests pass without strict mode too? Do we need a separate non-strict version of the tests, or can it not cause an issue?

packages/react-instantsearch-hooks/src/lib/useWidget.ts
```diff
@@ -0,0 +1,70 @@
+import { useEffect, useRef } from 'react';
+
+import { dequal } from './dequal';
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
+
+import type { Widget } from 'instantsearch.js';
+import type { IndexWidget } from 'instantsearch.js/es/widgets/index/index';
+
+export function useWidget<TWidget extends Widget | IndexWidget, TProps>(
```
Haroenv, line 9: should we have unit tests for useWidget too, or rely only on integrated tests?

Haroenv (approved):
this is obviously a big change, but I can't think of any scenarios in which it would not work, unless React chooses to have a different component instance

packages/react-instantsearch-hooks/src/components/__tests__/Index.test.tsx
francoischalifour: They do pass without Strict Mode.

At first I ran each with both regular mode and Strict Mode, but since we expect behaviors to be strictly identical, I figured we better just use Strict Mode.

packages/react-instantsearch-hooks/src/lib/useWidget.ts
francoischalifour: It's only extracted because we use the same logic for `useConnector()` and `<Index>`, but it's tested at the public API level.

packages/react-instantsearch-hooks/src/hooks/__tests__/useConnector.test.tsx
```diff
@@ -526,19 +540,111 @@ describe('useConnector', () => {
...
+    const { rerender } = render(<App callback={() => {}} />);
+
+    await waitFor(() => expect(searchClient.search).toHaveBeenCalledTimes(1));
+    expect(indexContext.current!.addWidgets).toHaveBeenCalledTimes(1);
+
+    rerender(<App callback={() => {}} />);
 
-    // This checks that InstantSearch doesn't re-render endlessly. We should
-    // still be able to optimize this render count to `1`, but `2` is acceptable
-    // for now compared to an infinite loop.
```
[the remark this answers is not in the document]
francoischalifour: Indeed, I didn't mean to remove it, good catch. See 72fb73df99f1baa35ece3e88aeac9600a735cbc7.

dhayab (approved):
Amazing refactoring 👏 and thanks for detailing the behavior of `useWidget()` with comments. That helps a lot!

PR #3515, opened 2022-06-13
fix(hooks-server): import react server via an expression

### LINKED ISSUE #3512: React Instantsearch Hooks Server: Module not found react-dom error is thrown when usign webpack
## 🐛 Bug description

When importing the `getServerState` function from `react-instantsearch-hooks-server` the following error is thrown:

> Module not found: Package path ./server.js is not exported from package ./node_modules/react-dom (see exports field in ./node_modules/react-dom/package.json)

## 🔍 Bug reproduction

1. Clone the `hooks-ssr` example: https://github.com/algolia/react-instantsearch/tree/v6.22.0/examples/hooks-ssr
2. Update package.json with the following dependencies (note that React 18 is being used):
```json
 "dependencies": {
    "algoliasearch": "4.11.0",
    "instantsearch.css": "7.4.5",
    "next": "12.1.6",
    "react": "18.1.0",
    "react-dom": "18.1.0",
    "react-instantsearch-hooks": "6.22.0",
    "react-instantsearch-hooks-server": "6.22.0"
  },
  "devDependencies": {
    "@types/node": "17.0.36",
    "@types/react": "18.0.9",
    "@types/react-dom": "18.0.5",
    "eslint": "8.16.0",
    "eslint-config-next": "12.1.6",
    "typescript": "4.5.4"
  }
```
3. Install the dependencies
4. Start the application
5.  See the error stated above being thrown

**Live reproduction:**

https://codesandbox.io/s/delicate-https-ekihqc?file=/package.json

## 💭 Expected behavior

The application should load normally.

## Environment

- React InstantSearch version: 6.22.0
- React version: 18.1.0

## Additional context

I got this issue on a Next.js application. Next.js uses webpack and it seems that webpack performs some static analysis of imports at build time ([ref](https://github.com/webpack/webpack/issues/6680#issuecomment-370800037)). Since there are [two dynamic imports](https://github.com/algolia/react-instantsearch/blob/68072324cf302801502a1b4c3d06703e57b55a97/packages/react-instantsearch-hooks-server/src/getServerState.tsx#L189-L190) on the `getServerState` function to try to import the `react-dom/server` module from different paths, webpack will throw an error at build time because the `react-dom/server.js` does not exist. It seems that this might be what is causing the problem.

Haroenv:
**Summary**

This is still WIP, but can be tried out via codesandbox now.

According to https://github.com/webpack/webpack/issues/13865 an expression avoids the shortcut of importing all items

**Result**

No errors in Next.js example

fixes #3512

### DIFF
--- a/packages/react-instantsearch-hooks-server/src/getServerState.tsx
+++ b/packages/react-instantsearch-hooks-server/src/getServerState.tsx
@@ -182,22 +182,21 @@ function getInitialResults(rootIndex: IndexWidget): InitialResults {
 }
 
 function importRenderToString() {
-  return Promise.all([
-    // React pre-18 doesn't use `exports` in package.json, requiring a fully resolved path
-    // Thus, only one of these imports is correct
-    // eslint-disable-next-line import/extensions
-    import('react-dom/server.js').catch(() => {}),
-    import('react-dom/server').catch(() => {}),
-  ]).then((imports) => {
-    const ReactDOMServer = imports.find(
-      (mod): mod is { renderToString: typeof RenderToString } =>
-        mod !== undefined
-    );
+  // React pre-18 doesn't use `exports` in package.json, requiring a fully resolved path
+  // Thus, only one of these imports is correct
+  const modules = ['react-dom/server.js', 'react-dom/server'];
+  return Promise.all(modules.map((mod) => import(mod).catch(() => {}))).then(
+    (imports: unknown[]) => {
+      const ReactDOMServer = imports.find(
+        (mod): mod is { renderToString: typeof RenderToString } =>
+          mod !== undefined
+      );
 
-    if (!ReactDOMServer) {
-      throw new Error('Could not import ReactDOMServer.');
+      if (!ReactDOMServer) {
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -182,22 +182,21 @@ function getInitialResults(rootIndex: IndexWidget): InitialResults {
...
-    import('react-dom/server.js').catch(() => {}),
-    import('react-dom/server').catch(() => {}),
-  ]).then((imports) => {
-    const ReactDOMServer = imports.find(
-      (mod): mod is { renderToString: typeof RenderToString } =>
-        mod !== undefined
-    );
+  // React pre-18 doesn't use `exports` in package.json, requiring a fully resolved path
+  // Thus, only one of these imports is correct
+  const modules = ['react-dom/server.js', 'react-dom/server'];
```
francoischalifour, line 187: Can we also mention why we need this intermediary variable to avoid regressions?

francoischalifour (approved):

PR #3518, opened 2022-06-14
fix(hooks): mount widgets in SSR to retrieve HTML

### LINKED ISSUE #3511: React Instantsearch Hooks: SSR not working on v6.27.0
## 🐛 Bug description

SSR with Next.js is not working on the latest version of `react-instantsearch-hooks-web` and `react-instantsearch-hooks-server` (6.27.0). Every element that is rendered inside the `<InstantSearch />` component is only rendered on the client side,  

## 🔍 Bug reproduction

1. Clone the `hooks-ssr` example: https://github.com/algolia/react-instantsearch/tree/v6.22.0/examples/hooks-ssr
2. Change the `react-instantsearch-hooks`, `react-instantsearch-hooks-server` and `react-instantsearch-hooks-web` dependencies version from 6.22.0 to 6.27.0
3. Install the npm dependencies
4. Start the server
5. Verify that no search results were server side rendered by checking the html source

**Live reproduction:**

https://codesandbox.io/s/github/algolia/react-instantsearch/tree/v6.27.0/examples/hooks-next?file=/package.json

## 💭 Expected behavior

The widgets inside the `<InstantSearch />` should be server side rendered

## Environment

- React InstantSearch Hooks version: 6.27.0
- React version: 17.0.2

## Additional context

Tried to downgrade to version 6.26.0 and the search results are server side rendered, so the issue should have been introduced on the last release

francoischalifour:
This fixes our SSR APIs as reported in #3511.

During the `useSyncExternalStore()` migration in #3489, we omitted some SSR checks by removing the `serverState.initialResults` condition. This was responsible for mounting the widgets during rendering, to then render them on the server. We now also cover the generated HTML with a test.

Note that this solution doesn't work with Strict Mode, as it wasn't prior to the bug. This is just a fix to release early, and then we'll work on Strict Mode support.

Closes #3511

### DIFF
--- a/packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
+++ b/packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
@@ -3,16 +3,16 @@
  */
 
 import React, { version as ReactVersion } from 'react';
+import { renderToString } from 'react-dom/server';
 import {
   InstantSearch,
   InstantSearchSSRProvider,
   Index,
   DynamicWidgets,
-  useHits,
-  useRefinementList,
-  useSearchBox,
   version,
+  useSearchBox,
 } from 'react-instantsearch-hooks';
+import { Hits, RefinementList } from 'react-instantsearch-hooks-web';
 
 import { createSearchClient } from '../../../../test/mock';
 import { getServerState } from '../getServerState';
@@ -20,8 +20,19 @@ import { getServerState } from '../getServerState';
 import type {
   InstantSearchServerState,
   InstantSearchProps,
-  UseRefinementListProps,
 } from 'react-instantsearch-hooks';
+
+function SearchBox() {
+  const { query } = useSearchBox();
+
+  return (
+    <div className="ais-SearchBox">
+      <form action="" className="ais-SearchBox-form" noValidate>
+        <input className="ais-SearchBox-input" type="search" value={query} />
+      </form>
+    </div>
+  );
+}
 
 type CreateTestEnvironmentProps = {
   searchClient: InstantSearchProps['searchClient'];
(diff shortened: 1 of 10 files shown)


packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
```diff
@@ -362,19 +378,58 @@ describe('getServerState', () => {
...
+                 value="iphone"
+          >
+        </form>
+      </div>
+      <h2>
+        instant_search
+      </h2>
+      <div class="ais-Hits ais-Hits--empty">
+        <ol class="ais-Hits-list">
+        </ol>
```
Haroenv, line 410: isn't this meant to have li's? or is the search client returning empty hits?

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
```diff
@@ -145,13 +145,12 @@ export function useConnector<
...
-  // On the server, we add the widget early to retrieve its search parameters
-  // in the render pass.
-  if (serverContext && !parentIndex.getWidgets().includes(widget)) {
-    parentIndex.addWidgets([widget]);
-  }
+  useWidget({
+    widget,
+    parentIndex,
+    props: stableProps,
+    shouldSsr: Boolean(serverContext),
```
Haroenv, line 152: not serverContext || initialResults?

packages/react-instantsearch-hooks/src/lib/useWidget.ts
```diff
@@ -22,6 +28,8 @@ export function useWidget<TWidget extends Widget | IndexWidget, TProps>(
   }, [widget]);
 
   const cleanupTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const shouldSsrRender =
```
Haroenv, line 31:
```suggestion
  const shouldAddWidget =
```

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
Haroenv: not needed because of the getWidgetRenderState branch

packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
francoischalifour: It returns no hits, I added some.

packages/react-instantsearch-hooks/src/hooks/useConnector.ts
francoischalifour: `initialResults` is only needed for `<Index>`.

Haroenv (approved):

PR #3517, opened 2022-06-14
feat(hooks): support React 18

francoischalifour:
This is the aggregated PR that brings support for React 18. All changes have been reviewed individually.

PR #3526, opened 2022-06-20
chore(ci): run React 17 tests

francoischalifour:
This runs our unit test suite with React 17 on the CI.

Since #3517, our codebase uses React 18, but we want to make sure that our libraries work on React 17 as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,6 +92,44 @@ jobs:
             # retry introduced in https://github.com/algolia/react-instantsearch/pull/3475
             retry yarn test --maxWorkers=4
 
+  test_unit_react_v17:
+    <<: *defaults
+    steps:
+      - checkout
+      - run: *install_yarn_version
+      - restore_cache: *restore_yarn_cache
+      - run: *run_yarn_install
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Install React v17 dependencies
+          command: |
+            yarn add --force -DW \
+              react@17.0.2 \
+              @types/react@17.0.47 \
+              react-dom@17.0.2 \
+              @types/react-dom@17.0.17 \
+              react-test-renderer@17.0.2 \
+              @types/react-test-renderer@17.0.2 \
+              @testing-library/react@12.1.4
+      - run:
+          name: Unit Tests
+          command: |
+            retry() {
+              MAX_RETRY=2
+              n=0
+              until [ $n -ge $MAX_RETRY ]
+              do
+                  "$@" && break
+                  n=$[$n+1]
+              done
+              if [ $n -ge $MAX_RETRY ]; then
+                exit 1
+              fi
+            }
+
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
```diff
@@ -421,7 +417,6 @@ describe('getServerState', () => {
               novalidate
         >
           <input class="ais-SearchBox-input"
-                 type="search"
```
francoischalifour, line 424: I removed this attribute because the attributes order differed in the React 17 and React 18 snapshots.

.circleci/config.yml
```diff
@@ -92,6 +92,44 @@ jobs:
...
+      - checkout
+      - run: *install_yarn_version
+      - restore_cache: *restore_yarn_cache
+      - run: *run_yarn_install
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Install React v17 dependencies
+          command: |
+            yarn add --force -DW \
+              react@17.0.2 \
```
Haroenv, line 107: this seems like a decent solution for now, but I wonder if we could have a separate package with those dependencies, and then run the tests:

```
└ tests-react-17
  └ package.json with react 17 dependencies
  └ jest ../jest.config.js
```

That way the node_modules used should be the one from react 17 (maybe also similar possible with algoliasearch v3), but tbh I'm not fully sure if that would work

Haroenv (approved):

.circleci/config.yml
```diff
@@ -92,6 +92,44 @@ jobs:
...
+              n=0
+              until [ $n -ge $MAX_RETRY ]
+              do
+                  "$@" && break
+                  n=$[$n+1]
+              done
+              if [ $n -ge $MAX_RETRY ]; then
+                exit 1
+              fi
+            }
```
dhayab, line 128: Maybe this can be made as an alias and instead could be referenced before this command as something like: 
```yml
- add_retry_util: *add_retry_util
```

This would allow us to only have one definition for both jobs that use it. It might be necessary to export the function though, I'm not sure if the job steps are run in isolation or not with Circle CI.

dhayab (approved):

.circleci/config.yml
francoischalifour: They seem to be in insolation according to the [**CI output**](https://app.circleci.com/pipelines/github/algolia/react-instantsearch/3021/workflows/0d5d919a-6a45-4895-b081-3a38252e5268/jobs/56244/parallel-runs/0/steps/0-110).

That's the commit I tried: https://github.com/algolia/react-instantsearch/compare/chore/ci-alias-retry.

Haroenv: alternatively we could isolate the whole `retry jest tests`, but that seems a bit much

dhayab: Thanks for trying! It's fine as is for me.

PR #3529, opened 2022-06-21
test(hooks): test rerenders with unstable routing

francoischalifour:
This adds a test for using an unstable `routing` prop on `<InstantSearch>` to make sure that the state is recovered.

Since this test passes, it means that we can simplify the [routing APIs docs](https://www.algolia.com/doc/api-reference/widgets/history-router/react-hooks/#widget-param-getlocation) to remove the `useMemo()`.


Haroenv (approved):
I'm guessing this is solved more or less accidentally because a change in routing doesn't cause InstantSearch to remount?

PR #3551, opened 2022-07-06
refactor(hooks): remove `useEffect()` from `<SearchBox>`

francoischalifour:
## Description

This refactors the `<SearchBox>` component to remove the `useEffect()`.

You can learn more in [**You Might Not Need an Effect**](https://beta.reactjs.org/learn/you-might-not-need-an-effect), notably these two sections:

- [Adjusting some state when a prop changes](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)
- [Sharing logic between event handlers](https://beta.reactjs.org/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers)

## Related

- #3552
- #3553

### DIFF
--- a/packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
+++ b/packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
@@ -33,7 +33,7 @@ export function SearchBox({ queryHook, ...props }: SearchBoxProps) {
   }
 
   function onChange(event: React.ChangeEvent<HTMLInputElement>) {
-    const targetValue = event.target.value;
+    const targetValue = event.currentTarget.value;
     setValue(targetValue);
     refine(targetValue);
   }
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
```diff
@@ -29,33 +29,22 @@ export function SearchBox({ queryHook, ...props }: SearchBoxProps) {
 
   function onReset() {
     setValue('');
+    refine('');
   }
 
   function onChange(event: React.ChangeEvent<HTMLInputElement>) {
-    setValue(event.currentTarget.value);
+    const targetValue = event.target.value;
```
Haroenv, line 36: we're changing from currentTarget to target, could that have any impact? I don't think so as the listener is directly on the input

Haroenv (approved):
much cleaner flow!

packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
francoischalifour: That wasn't intentional, I'll revert this.

francoischalifour, line 36:
```suggestion
    const targetValue = event.currentTarget.value;
```

PR #3554, opened 2022-07-07
fix(hooks): provide state and results APIs from "render" event

francoischalifour:
This updates the [`useInstantSearch()`](https://www.algolia.com/doc/api-reference/widgets/use-instantsearch/react-hooks/) implementation for the search state and search results APIs (`uiState`, `setUiState()`, `indexUiState`, `setIndexUiState()`, `results`, `scopedResults`).

The **previous implementation relied on adding widgets** to the instance to retrieve the search state and the search results. This added footprint on the instance, and triggered unnecessary renders.

This **new implementation subscribes to the InstantSearch `render` event** to update the React state. Since we rely on this event, the React state will get updated at the same time as all the other widgets' render, resulting in a batch update.


Haroenv (approved):
🥇

Issue #3537, opened 2022-06-23
Nexjts SSR broken functionality with reactStrictMode enabled


abusada:
## 🐛 Bug description

I'm following [this guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react-hooks/?client=index.tsx#with-nextjs) trying to achieve server side rendering with nextjs and react-instantsearch-hooks.

once i wrap my code with `InstantSearchSSRProvider`, search functionality simple stops working,
![screenshot- 2022-06-23 at 15 57 31](https://user-images.githubusercontent.com/2572320/175317280-8d5260bc-4011-41ab-be58-e5e7e6c296d0.gif)

**Live reproduction:**

https://codesandbox.io/s/hungry-stitch-b070o2?file=/pages/index.tsx:990-1014

## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Go to the provided sandbox: https://codesandbox.io/s/hungry-stitch-b070o2?file=/pages/index.tsx:990-1014
2. Try using the search box

## 💭 Expected behavior

Expect search to work and hits to update, notice how in the [withour-ssr page](https://b070o2.sse.codesandbox.io/without-ssr) the search functionality is working properly.

![screenshot- 2022-06-23 at 16 03 39](https://user-images.githubusercontent.com/2572320/175318566-9a3e3c7c-a257-49f9-ac6b-1a482490f35b.gif)

## Environment

- React InstantSearch Hooks version: [6.29.0]
- React version: [18.2.0]
- Nextjs version [12.1.6]

please **note** the following configs in `next.config.js`:

```
module.exports = {
  reactStrictMode: true,
  swcMinify: true
};
```

**update**
setting `reactStrictMode` to false fixes the issue, here is a fork of the above sandbox with strict mode turned off: https://codesandbox.io/s/lucid-cache-n7h0no?file=/next.config.js:19-45

abusada:
Could you possibly point me to the code where you suspect the issue is happening? perhaps i could be of help!

Haroenv:
There's some tests:

- https://github.com/algolia/react-instantsearch/pull/3518/files#diff-7accb274a0dfc2e4d20af5c09f1c9b6f800efaedcc568ffa1fd06dc5f8de55eeR160
[...]

PR #3561, opened 2022-07-13
fix(hooks): use single instance in <InstantSearch>

### LINKED ISSUE #3530: NextJS static page with rewrite rule -> error on refine
## 🐛 Bug description

Refine in a nextJS static page throw this error if you defined a next.js rewrite rule:
`Error: Cannot refine the undeclared facet brand; it should be added to the helper options facets, disjunctiveFacets or hierarchicalFacets`

## 🔍 Bug reproduction

1. Put a rewite rule in your `next.config.js`
2. Browse a static page including algolia search
3. Click on a refinement in the left sidebar

**Live reproduction:**

https://codesandbox.io/s/quizzical-night-mjlw22

## 💭 Expected behavior

The refinement should be applied correctly

## 🖥 Screenshots

image

## Environment

- React InstantSearch Hooks version: 6.28.0
- React version: 18.1.0

## Additional context

I think that the problem is started with the migration to `useSyncExternalStore` in `useInstantSearchApi` because I noticed that in the case mentioned above `search.dispose()` function is called after the page load. Is it possible that migrating from `useEffect` to `useSyncExternalStore+useCallback` you forgot to remove the old cleanup function and actually it's invoked accidentally in some cases?

image

### LINKED ISSUE #3550: Conflict using InstantSearchSSRProvider with routing and Remix Run useFetcher hook
## 🐛 Bug description

In a Remix Run project I have some difficulties running Instant Search React Hook with the SSR Provider. The problem occurs when I try to keep the UIState sync with the browser url via the ```routing``` attribute that require a getLocation method when render with SSR.

```typescript
 <InstantSearch
      searchClient={searchClient}
      indexName="REMIX_TEST"
      routing={{
        router: history({
          getLocation: () => {
            return typeof window === "undefined"
              ? (new URL(url) as unknown as Location)
              : window.location;
          },
        }),
      }}
    />
```

This work well until a fetch request is triggered somewhere in the app via the Remix hook ```useFetcher``` even if the hook is used outside of ```<InstantSearch />```.

This cause the history router to push an empty state to the browser url and "broke" any search related component. 

## 🔍 Bug reproduction

1. Clone this repo
2. run ```npm install``` then ```npm dev```
3. Open localhost:300
4. Click the With Hook SSR link
5. Enter some characters and select some filters
6. Click the details button on any result
7. Click a filters checkbox
8. See the Error in the console as well as log of state push via the history

**Live reproduction:**

https://codesandbox.io/s/remix-algolia-hook-ulbwcw

## 💭 Expected behavior

The history should not push an empty state to the browser url and search component (widget from the hook library as well as custom widget using hook) should be usable.

## 🖥 Screenshots

Screen Shot 2022-07-05 at 21 55 15

## Environment

- React InstantSearch Hooks version: 6.29.0
- React version: 17.0.2
- Browser: 103.0.5060.53
- OS: macOS

## Additional context

The linked repo install a patch for the instantsearch.js library to log state push to the url

francoischalifour:
This addresses a lot of issues related to the `<InstantSearch>` lifecycle that happened when remounting:

- Closes #3530
- Closes #3550
- Maybe #3537

### Updating `<InstantSearch>` props

Prior to this PR, any prop change would create a new instance of InstantSearch.js, in an Effect. Now, we track props changes during rendering, and update them with the underlying InstantSearch.js internals. That's similar to [how it's done in Vue InstantSearch](https://github.com/algolia/vue-instantsearch//blob/71c25d460e1b077b62ca41a59a5c4df7b713ec56/src/util/createInstantSearchComponent.js#L16-L36).

InstantSearch.js doesn't provide a way to override `routing`, so this prop cannot be changed. If we find cases where we need `routing` to be dynamic, we can provide an internal API in InstantSearch.js.

But long term, I think we should have an update API in InstantSearch.js. There are many corner cases right now, like you can't have a `searchFunction`, and then remove it. InstantSearch.js will throw because it still expects a `searchFunction` to exist (because it was attached in `start()`).

### Unmounting `<InstantSearch>`

The `dispose()` call now happens in a `setTimeout()`, as in `useConnector()`. This lets us prevent InstantSearch from being recreated, and therefore to manipulate a single instance in all the widgets.

### Supporting SSR in Strict Mode

The single InstantSearch.js instance created is now compatible with SSR in Strict Mode.

### Routing tests

I copied the [InstantSearch.js routing tests](https://github.com/algolia/instantsearch.js/tree/cf738b50d36c99c12995c79157920748094b4a7c/src/lib/__tests__/routing) to make sure that React InstantSearch Hooks behaves similarly to InstantSearch.js. This is useful because we skip the `removeWidgets()` call on `dispose()`, as I explained in the comments. This lets React unmount all the widgets and not rely on `<InstantSearch>` to remove them.

### DIFF
--- a/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
+++ b/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
@@ -542,8 +542,7 @@ describe('InstantSearch', () => {
 
     await waitFor(() => {
       expect(searchClient1.search).toHaveBeenCalledTimes(1);
-      // @TODO: can this call count go down to 1 in Strict Mode?
-      expect(searchClient2.search).toHaveBeenCalledTimes(2);
+      expect(searchClient2.search).toHaveBeenCalledTimes(1);
       expect(searchClient2.addAlgoliaAgent).toHaveBeenCalledWith(
         `react (${ReactVersion})`
       );
@@ -559,8 +558,8 @@ describe('InstantSearch', () => {
 
     await waitFor(() => {
       expect(searchClient1.search).toHaveBeenCalledTimes(1);
-      expect(searchClient2.search).toHaveBeenCalledTimes(2);
-      expect(searchClient3.search).toHaveBeenCalledTimes(2);
+      expect(searchClient2.search).toHaveBeenCalledTimes(1);
+      expect(searchClient3.search).toHaveBeenCalledTimes(1);
       expect(searchClient3.addAlgoliaAgent).toHaveBeenCalledWith(
         `react (${ReactVersion})`
       );
@@ -575,7 +574,7 @@ describe('InstantSearch', () => {
     userEvent.type(screen.getByRole('searchbox'), 'iphone');
 
     await waitFor(() => {
-      expect(searchClient3.search).toHaveBeenCalledTimes(8);
+      expect(searchClient3.search).toHaveBeenCalledTimes(7);
       expect(searchClient3.search).toHaveBeenLastCalledWith([
         {
           indexName: 'indexName',
@@ -611,11 +610,12 @@ describe('InstantSearch', () => {
       ]);
     });
 
+    expect(searchClient.search).toHaveBeenCalledTimes(1);
+
     rerender(<App indexName="indexName2" />);
 
     await waitFor(() => {
-      // @TODO: can this call count go down in Strict Mode?
-      expect(searchClient.search).toHaveBeenCalledTimes(3);
+      expect(searchClient.search).toHaveBeenCalledTimes(2);
       expect(searchClient.search).toHaveBeenLastCalledWith([
         expect.objectContaining({
           indexName: 'indexName2',
@@ -626,8 +626,7 @@ describe('InstantSearch', () => {
     rerender(<App indexName="indexName3" />);
 
     await waitFor(() => {
-      // @TODO: can this call count go down in Strict Mode?
-      expect(searchClient.search).toHaveBeenCalledTimes(5);
+      expect(searchClient.search).toHaveBeenCalledTimes(3);
       expect(searchClient.search).toHaveBeenLastCalledWith([
         expect.objectContaining({
           indexName: 'indexName3',
@@ -638,7 +637,7 @@ describe('InstantSearch', () => {
     userEvent.type(screen.getByRole('searchbox'), 'iphone');
 
     await waitFor(() => {
-      expect(searchClient.search).toHaveBeenCalledTimes(11);
+      expect(searchClient.search).toHaveBeenCalledTimes(9);
       expect(searchClient.search).toHaveBeenLastCalledWith([
         {
           indexName: 'indexName3',
--- a/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
+++ b/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
@@ -147,8 +147,11 @@ describe('InstantSearch', () => {
 
     unmount();
 
-    expect(searchContext.current!.dispose).toHaveBeenCalledTimes(1);
-    expect(searchContext.current!.started).toEqual(false);
+    await waitFor(() => {
+      expect(searchContext.current!.dispose).toHaveBeenCalledTimes(1);
+      expect(searchContext.current!.started).toEqual(false);
+      expect(searchContext.current!.mainIndex.getWidgets()).toEqual([]);
+    });
   });
 
   test('triggers a single network request on mount with widgets', async () => {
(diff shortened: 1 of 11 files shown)


packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
```diff
@@ -147,8 +147,11 @@ describe('InstantSearch', () => {
 
     unmount();
 
-    expect(searchContext.current!.dispose).toHaveBeenCalledTimes(1);
-    expect(searchContext.current!.started).toEqual(false);
+    await waitFor(() => {
```
francoischalifour, line 150: The `dispose()` call now happens in a `setTimeout()`.

packages/react-instantsearch-hooks/src/components/__tests__/InstantSearchSSRProvider.test.tsx
```diff
@@ -379,14 +379,15 @@ describe('InstantSearchSSRProvider', () => {
       },
     };
 
-    // @TODO: this test doesn't work in Strict Mode
```
francoischalifour, line 382: It does now!

packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
```diff
@@ -452,9 +455,10 @@ describe('InstantSearch', () => {
...
 
   // This test shows that giving an unstable `onStateChange` reference (or any
-  // unstable prop) remounts the <InstantSearch> component and therefore resets
-  // the state after the remount.
-  // Users need to provide stable references for rerenders to keep the state.
+  // unstable prop) does not remount the <InstantSearch> component and therefore
+  // keeps the state.
+  // Prior to the current implementation, we created a new InstantSearch.js
+  // instance at each prop change, which made this test fail.
   test('recovers the state on rerender with an unstable onStateChange', async () => {
```
francoischalifour, line 462: We can remove this test if you think it doesn't make sense to have it now.

Haroenv: it makes sense to keep for when the "update" method moves to InstantSearch.js

packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
```diff
@@ -508,11 +510,244 @@ describe('InstantSearch', () => {
...
+          indexName: 'indexName1',
+        }),
+      ]);
+    });
+
+    rerender(<App indexName="indexName2" />);
+
+    await waitFor(() => {
+      // @TODO: can this call count go down in Strict Mode?
+      expect(searchClient.search).toHaveBeenCalledTimes(3);
```
Haroenv, line 618: do we know why one extra search happened with index 1? is it the same parameters?

packages/react-instantsearch-hooks/src/components/__tests__/routing/dispose-start.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    // 3. Dispose: '/'
+    // 4. Refine: '/'
+    // 5. Start: '/'
+    // 6. Refine: '/?indexName[query]=Apple'
+
+    const pushState = jest.spyOn(window.history, 'pushState');
+    const searchClient = createSearchClient({});
+
+    let setQuery = (_query: string) => {};
+    function QueryMiddleware() {
```
Haroenv, line 26: wonder if this is clearer with a component using useSearchBox?

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -84,8 +66,102 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
...
+      setPrevProps(props);
+    }
+
+    // Updating the `routing` prop is not supported because InstantSearch.js
+    // doesn't let us change it. This might not be a problem though, because `routing`
+    // shouldn't need to be dynamic.
+    // If we find scenarios where `routing` needs to change, we can always expose
+    // it privately on the InstantSearch instance. Another way would be to
+    // manually inject the routing middleware in this library, and not rely
+    // on the provided `routing` prop.
```
Haroenv, line 116: still not sure whether we should let people know routing will be stale, in theory if the stale routing is the same as the newer ones (but just not referentially) there shouldn't be an issue, but some weird edge cases like using react state in the router would break

Haroenv (approved):

packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
francoischalifour: These are the `searchClient.search()` mock calls at this step:

```json
[
      [
        [
          {
            "indexName": "indexName1",
            "params": {
              "facets": [],
              "query": "",
              "tagFilters": ""
            }
          }
        ]
      ],
      [
        [
          {
            "indexName": "indexName2",
            "params": {
              "facets": [],
              "query": "",
              "tagFilters": ""
            }
          }
        ]
      ],
      [
        [
          {
            "indexName": "indexName2",
            "params": {
              "facets": [],
              "query": "",
              "tagFilters": ""
            }
          }
        ]
      ]
    ]
```

So at least the client is able to cache it, but I'm not yet sure why there's this extra request.

francoischalifour: Storing the prev props in a ref rather than in state triggers the correct amount of searches. See 2bb00c6220dae6ceb9a05615ff255840f8d38214.

packages/react-instantsearch-hooks/src/components/__tests__/routing/dispose-start.test.tsx
francoischalifour: See aa1644aa9627bfc770ec5cd0b2cd3c5eaf6cbec0.

Haroenv (approved):

dhayab (approved):

PR #3565, opened 2022-07-15
feat(SearchBox): expose formRef

esaborit4code:
**Summary**
This change is motivated from a very specific use case:

I have a custom form input component which includes a `SearchBox`, so I my own `<form>` tag is wrapping the `<form>` that's within `SearchBox`. This leads to an unexpected form submission when pressing the <kbd>Enter</kbd> key in the search box text input. 

I've already tried `preventDefault` and `stopPropagation` in the `onSubmit` callback available for `SearchBox` but it seems the way React handles events makes this useless. Getting a `ref` for the form allowed me to stop the submit event by using `addEventListener` on the `ref.current` element.
 
Since there's already an `inputRef` I think having a `formRef` won't hurt and would help others that may find themselves in a similar form nesting situation.

**Result**
A new `formRef` prop analog to current `inputRef` is added to `SearchBox`.

### DIFF
--- a/packages/react-instantsearch-dom/src/components/SearchBox.js
+++ b/packages/react-instantsearch-dom/src/components/SearchBox.js
@@ -82,6 +82,11 @@ class SearchBox extends Component {
     isSearchStalled: PropTypes.bool,
     showLoadingIndicator: PropTypes.bool,
 
+    formRef: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.exact({ current: PropTypes.object }),
+    ]),
+
     inputRef: PropTypes.oneOfType([
       PropTypes.func,
       PropTypes.exact({ current: PropTypes.object }),
@@ -249,6 +254,7 @@ class SearchBox extends Component {
     return (
       <div className={classNames(cx(''), className)}>
         <form
+          ref={this.props.formRef}
           noValidate
           onSubmit={this.props.onSubmit ? this.props.onSubmit : this.onSubmit}
           onReset={this.onReset}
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch-dom/src/components/SearchBox.js
```diff
@@ -249,6 +264,7 @@ class SearchBox extends Component {
     return (
       <div className={classNames(cx(''), className)}>
         <form
+          ref={this.onFormMount}
```
Haroenv, line 267:
```suggestion
          ref={this.props.formRef}
```

as this.form isn't used, there's no need for the function dance I think

Haroenv (commented):
It makes sense to me, as we'd like people to mainly use the official SearchBox component, however; this should have some added tests too.

Thanks for your PR already!

packages/react-instantsearch-hooks-web/src/ui/SearchBox.tsx
```diff
@@ -63,6 +63,7 @@ export type SearchBoxProps = Omit<
   Pick<React.ComponentProps<'form'>, 'onSubmit'> &
   Required<Pick<React.ComponentProps<'form'>, 'onReset'>> &
   Pick<React.ComponentProps<'input'>, 'placeholder' | 'onChange'> & {
+    formRef: React.RefObject<HTMLFormElement>;
```
dhayab, line 66: In this form it forces all implementations to specify `formRef`. You should make it optional to prevent that.

```suggestion
    formRef?: React.RefObject<HTMLFormElement>;
```

dhayab (changes requested):
Hi, we'll be able to move forward after you address the comments and add a test case for `formRef`.

Thanks again for your contribution!

packages/react-instantsearch-dom/src/components/SearchBox.js
esaborit4code: Yes, that makes sense. I just mimicked `onInputMount`, but there was no real need for that. Thanks!

Haroenv (approved):

PR #3583, opened 2022-07-29
fix(hooks): use `label` CSS class in <HierarchicalMenu>

FabienMotte:
**Summary**

This PR fixes the `label` CSS class (instead of `labelText`) used in the `HierarchicalMenu` widget of React InstantSearch Hooks.

According to the [InstantSearch specs](https://instantsearch-css.netlify.app/widgets/hierarchical-menu/), the correct CSS class is `.ais-HierarchicalMenu-label`.
For reference, here is the [implementation of the CSS class in InstantSearch.js](https://github.com/algolia/instantsearch.js/blob/master/src/widgets/hierarchical-menu/hierarchical-menu.tsx#L335).

**Note:** This fix may break current user implementations that depend on the `labelText` CSS class so we want to make sure the release note mentions it clearly.

**Results**

When a `HierarchicalMenu` item is selected, [the correct style](https://github.com/algolia/instantsearch-specs/blob/master/src/scss/themes/satellite.scss#L266-L273) is applied to it.

### DIFF
--- a/packages/react-instantsearch-hooks-web/src/ui/HierarchicalMenu.tsx
+++ b/packages/react-instantsearch-hooks-web/src/ui/HierarchicalMenu.tsx
@@ -108,7 +108,7 @@ function HierarchicalList({
             }}
           >
             <span
-              className={cx('ais-HierarchicalMenu-labelText', classNames.label)}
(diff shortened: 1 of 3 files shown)


dhayab (approved):
Thanks for noticing and fixing it 👏

PR #3585, opened 2022-08-02
feat(hooks): add `searchAsYouType` option to <SearchBox>

aymeric-giraudet:
**Summary**

As mentioned in discussion #3523, the `searchAsYouType` prop is missing from `react-instantsearch-hooks-web`'s `SearchBox` component, while it is present in [the non-hooks version](https://www.algolia.com/doc/api-reference/widgets/search-box/react/#widget-param-searchAsYouType).

**Notes**:
- I think we can do better on the widget/ui components splitting, so we won't have to duplicate default search behaviour for each JSX UI library (Preact or Solid for example)
- I'm not sure whether it's fine or not that we now refine on submit by default even if `searchAsYouType` is true, my guess would be it's okay as we won't trigger a search if query does not change.

**Result**

As the tests show, when `searchAsYouType` is set to `false`, it does not change internal InstantSearch state. Pressing enter or pushing the submit button will however trigger a search.

Ticket: [FX-1571](https://algolia.atlassian.net/browse/FX-1571)

### DIFF
--- a/packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
+++ b/packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
@@ -60,10 +60,12 @@ export function SearchBox({
   }
 
   function onSubmit(event: React.FormEvent<HTMLFormElement>) {
+    if (!searchAsYouType) {
+      refine(inputValue);
+    }
+
     if (props.onSubmit) {
       props.onSubmit(event);
-    } else if (!searchAsYouType) {
-      refine(inputValue);
     }
   }
 
--- a/packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
+++ b/packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
@@ -12,14 +12,30 @@ type UiProps = Pick<
   | 'isSearchStalled'
   | 'onChange'
   | 'onReset'
+  | 'onSubmit'
   | 'value'
   | 'translations'
 >;
 
-export type SearchBoxProps = Omit<SearchBoxUiComponentProps, keyof UiProps> &
-  UseSearchBoxProps;
+export type SearchBoxProps = Omit<
+  SearchBoxUiComponentProps,
+  Exclude<keyof UiProps, 'onSubmit'>
+> &
+  UseSearchBoxProps & {
+    /**
+     * If false, triggers the search only on submit.
+     * @type {boolean}
+     * @default true
+     * @optional
+     */
+    searchAsYouType?: boolean;
+  };
 
-export function SearchBox({ queryHook, ...props }: SearchBoxProps) {
+export function SearchBox({
+  queryHook,
+  searchAsYouType = true,
+  ...props
+}: SearchBoxProps) {
   const { query, refine, isSearchStalled } = useSearchBox(
     { queryHook },
     { $$widgetType: 'ais.searchBox' }
@@ -29,7 +45,10 @@ export function SearchBox({ queryHook, ...props }: SearchBoxProps) {
 
   function setQuery(newQuery: string) {
     setInputValue(newQuery);
-    refine(newQuery);
+
+    if (searchAsYouType) {
+      refine(newQuery);
+    }
   }
 
   function onReset() {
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
```diff
@@ -40,6 +59,14 @@ export function SearchBox({ queryHook, ...props }: SearchBoxProps) {
     setQuery(event.currentTarget.value);
   }
 
+  function onSubmit(event: React.FormEvent<HTMLFormElement>) {
+    if (props.onSubmit) {
+      props.onSubmit(event);
+    } else {
+      refine(inputValue);
+    }
```
FabienMotte, line 67:
```suggestion
    } else if(!searchAsYouType) {
      refine(inputValue);
    }
```
I guess it could make sense to follow [InstantSearch.js' implementation](https://github.com/algolia/instantsearch.js/blob/master/src/components/SearchBox/SearchBox.tsx#L114-L116)

FabienMotte (commented):
Good job! 🎉 👏

packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
aymeric-giraudet: Makes sense ! I updated this and the test name accordingly 😄

FabienMotte: Yes you're right! I think we should call the `onSubmit` prop if it exists and only call `refine` if `searchAsYouType` is `false`, like what we're doing on InstantSearch.js (decoupled conditions).
This way, `refine` still gets called even if the user passes the `onSubmit` prop.

packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
```diff
@@ -12,14 +12,30 @@ type UiProps = Pick<
...
+export type SearchBoxProps = Omit<
+  SearchBoxUiComponentProps,
+  Exclude<keyof UiProps, 'onSubmit'>
+> &
+  UseSearchBoxProps & {
+    /**
+     * If false, triggers the search only on submit.
+     * @type {boolean}
+     * @default true
+     * @optional
```
francoischalifour, line 29: We rely on TypeScript for type and optional indications. Otherwise we get duplications and it's harder to maintain.

```suggestion
     * Whether to trigger the search only on submit.
     * @default true
```

packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
aymeric-giraudet: Alright, I'm just running it after `refine` so that they can later refine with a custom value if they want :)

packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
```diff
@@ -40,6 +59,16 @@ export function SearchBox({ queryHook, ...props }: SearchBoxProps) {
...
   }
 
+  function onSubmit(event: React.FormEvent<HTMLFormElement>) {
+    if (!searchAsYouType) {
+      refine(inputValue);
+    }
+
+    if (props.onSubmit) {
+      props.onSubmit(event);
+    }
```
FabienMotte, line 69:
```suggestion
    props.onSubmit?.(event);
```
Small suggestion

aymeric-giraudet: Seems like the ESLint configuration does not allow it, I get a [no-unused-expressions](https://eslint.org/docs/latest/rules/no-unused-expressions) error.
And I think it's best to stay consistent with [the rest here](https://github.com/algolia/react-instantsearch/blob/c7c07f30ac1e7af5559b3fa74855ea001707390f/packages/react-instantsearch-hooks-web/src/ui/SearchBox.tsx#L164-L166)
We neither have `onSubmit?.(event)` nor `inputRef.current?.focus()`

packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
aymeric-giraudet: Makes sense, thanks 😄

packages/react-instantsearch-hooks-web/src/widgets/SearchBox.tsx
FabienMotte: No worries, I didn’t tested it, let’s keep it as it is, thanks!

dhayab (approved):
👏 Looks great, thanks!

FabienMotte (approved):

aymeric-giraudet:
Thanks to all of you 😄

Issue #3586, opened 2022-08-02
Multiple requests hit on component unmounts / route change in react


iamdhrooov:
## 🐛 Bug description

When changing routes in react or unmounting the components, the library hits the queries multiple times. (number of times specific refinements are added). This is happening on latest version v6.30.3 and not happening for v6.29.0

## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Go to [Codepen link](https://codesandbox.io/s/strange-flower-13wnc5?file=/App.tsx)
2. Click on `Change route on top`
3. Check API query calls in the network tab.
4. Multiple queries hit on unmount.
5. Change version to v6.29.0 -> works fine

**Live reproduction:**

https://codesandbox.io/s/strange-flower-13wnc5?file=/App.tsx

## 💭 Expected behavior

It should not hit queries multiple times.

## Environment

- React InstantSearch Hooks version: [6.30.3]

FabienMotte:
Hello @iamdhrooov,

Thanks for reporting this issue, I'm working on PR #3590 to fix it.
It should be released this week.

Thanks for your patience!

FabienMotte:
Hey @iamdhrooov! This issue should be fixed in the latest release of React InstantSearch Hooks [v6.31.1](https://github.com/algolia/react-instantsearch/releases/tag/v6.31.1).

Feel free to reopen if that's not the case, thank you!

iamdhrooov:
Thanks for the quick turnaround on this @FabienMotte Really appreciate it. :)

ccchapman:
Would these excess network requests result in additional charges?

Haroenv:
Yes, every request which happens will be charged. If this had an impact on your pricing fixed since 6.31.1, you can contact support@algolia.com

PR #3590, opened 2022-08-03
fix(hooks): prevent widget cleanup on `<InstantSearch>` unmount

FabienMotte:
#### Context

Starting from [v6.30.2](https://github.com/algolia/react-instantsearch/releases/tag/v6.30.2), when the `<InstantSearch>` component is unmounted, one search request per widget is triggered. The PR https://github.com/algolia/react-instantsearch/pull/3561 introduces this regression.

This issue was originally reported here: https://github.com/algolia/react-instantsearch/issues/3586.

#### Details

Currently, when the `<InstantSearch>` component is unmounted, a `cleanup` is scheduled using a `setTimeout`.

https://github.com/algolia/react-instantsearch/blob/affcbb58ee5c63345cc9424ad8ab566f3a01c9f9/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts#L156-L162

The widgets also rely on the same strategy when they unmount.

https://github.com/algolia/react-instantsearch/blob/affcbb58ee5c63345cc9424ad8ab566f3a01c9f9/packages/react-instantsearch-hooks/src/lib/useWidget.ts#L75-L77

And since the widgets are unmounted first, the `removeWidgets()` function is called and triggers a new search.

https://github.com/algolia/react-instantsearch/blob/affcbb58ee5c63345cc9424ad8ab566f3a01c9f9/packages/react-instantsearch-hooks/src/lib/useWidget.ts#L40-L42

#### Solution

This PR introduces a temporary fix that relies on a boolean flag `_preventWidgetCleanup` on the InstantSearch instance:

- This flag is set to `true` when the `<InstantSearch>` component is unmounted, thus preventing the widgets `cleanup` function to be executed.
- This flag is set to `false` when the `<InstantSearch>` component updates.

This solution works but should be considered temporary because in the future the InstantSearch.js update API should solve this issue.

#### Tests

This PR adds two tests, covering these use-cases:

- When the `<InstantSearch>` component is unmounted, no search request is made.
- When a widget is unmounted (we want to test this cause we introduce a new logic in the `useWidget` hook `cleanup` function), a new search request is made.

### Result

When the `<InstantSearch>` component is unmounted, no search request is made.

### DIFF
--- a/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
+++ b/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
@@ -750,7 +750,7 @@ describe('InstantSearch', () => {
     });
   });
 
-  test('triggers no more than one search on unmount', async () => {
+  test('triggers no search on unmount', async () => {
     const searchClient = createSearchClient({});
 
     function App() {
--- a/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
+++ b/packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
@@ -6,7 +6,7 @@ import React, { StrictMode, Suspense, version as ReactVersion } from 'react';
 import { SearchBox } from 'react-instantsearch-hooks-web';
 
 import { createSearchClient } from '../../../../../test/mock';
-import { createInstantSearchSpy } from '../../../../../test/utils';
+import { createInstantSearchSpy, wait } from '../../../../../test/utils';
 import { useRefinementList } from '../../connectors/useRefinementList';
 import version from '../../version';
 import { Index } from '../Index';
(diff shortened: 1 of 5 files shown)


packages/react-instantsearch-hooks/src/components/__tests__/InstantSearch.test.tsx
```diff
@@ -749,4 +749,55 @@ describe('InstantSearch', () => {
       expect(searchFunction2).toHaveBeenCalledTimes(5);
     });
   });
+
+  test('triggers no more than one search on unmount', async () => {
```
dhayab, line 753: The only triggered search request is on render, so when unmounting we should expect the number of calls not to grow.

```suggestion
  test('triggers no search on unmount', async () => {
```

dhayab (commented):
It's good for me, thanks @FabienMotte 👏 .

Since we've worked on it a bit together, I'll let @francoischalifour give his comments / approval.

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -159,7 +152,8 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
         // Executing the cleanup function in a `setTimeout()` lets us cancel it
         // in the next effect.
         // (There might be better ways to do this.)
-        cleanupTimerRef.current = setTimeout(cleanup, 0);
+        cleanupTimerRef.current = setTimeout(cleanup);
+        search._preventWidgetCleanup = true;
```
francoischalifour, line 156: Can we add a comment here to explain why this is needed?

packages/react-instantsearch-hooks/src/lib/useWidget.ts
```diff
@@ -31,13 +32,16 @@ export function useWidget<TWidget extends Widget | IndexWidget, TProps>({
...
+  const search = useInstantSearchContext();
+
   // This effect is responsible for adding, removing, and updating the widget.
   // We need to support scenarios where the widget is remounted quickly, like in
   // Strict Mode, so that we don't lose its state, and therefore that we don't
   // break routing.
   useIsomorphicLayoutEffect(() => {
     const previousWidget = prevWidgetRef.current;
     function cleanup() {
+      if (search._preventWidgetCleanup) return;
```
francoischalifour, line 44: Strictly speaking, we probably need to have `search` as a dependency of the effect now. You could temporarily change `useIsomorphicLayoutEffect()` to `useEffect()` to respect the ESLint rule.

francoischalifour (approved):
Great dig up!

packages/react-instantsearch-hooks/src/lib/useWidget.ts
FabienMotte: Good catch! I customized the ESLint rule to include the custom hook

```js
'react-hooks/exhaustive-deps': [
  'warn',
  {
    additionalHooks: '(useIsomorphicLayoutEffect)',
  },
],
```

Now it's warning about `props`, should we add it to the array of deps?

dhayab: I think we should add `props` to the dependency array as well, we expect to delete/add the widget when only the props change.

PR #3597, opened 2022-08-18
fix(useInstantSearch): `results` is `null` when first search is stalled

aymeric-giraudet:
**Summary**

[FX-1595](https://algolia.atlassian.net/browse/FX-1595)

`useSearchResults` returns a null `results` when the first search is stalled.
This is because the `render` event is triggered when a search is stalled, in which case `lastResults` from the helper will be null.

**Result**

Added a test with a simulated stalled search which fails with the previous code, but is valid now


packages/react-instantsearch-hooks/src/lib/useSearchResults.ts
```diff
@@ -21,10 +21,15 @@ export function useSearchResults(): SearchResultsApi {
...
   useEffect(() => {
     function handleRender() {
-      setSearchResults({
-        results: searchIndex.getResults()!, // Results can't be `null` after the first render.
-        scopedResults: searchIndex.getScopedResults(),
-      });
+      const results = searchIndex.getResults();
+
+      // results can be null when first search is stalled, in this case we skip the update
+      if (results !== null) {
```
aymeric-giraudet, line 27: I hesitated to also check if results are different from the previous, which could be the case when a subsequent search is stalled, which currently would trigger a rerender for nothing.
However, `useSearchResults` is only used in `useInstantSearch` which would anyway trigger rerender even if `useSearchResults` did not change, so I opted to keep the code simpler.

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchResults.test.tsx
```diff
@@ -49,4 +53,47 @@ describe('useSearchResults', () => {
     });
     expect(result.current.results.__isArtificial).toBeUndefined();
   });
+
+  test('should not return null results when first search is stalled', async () => {
```
sarahdayan, line 57:
```suggestion
  test('does not return `null` results when first search is stalled', async () => {
```

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchResults.test.tsx
```diff
@@ -49,4 +53,47 @@ describe('useSearchResults', () => {
...
+      results: expect.any(SearchResults),
+      scopedResults: [
+        expect.objectContaining({
+          helper: expect.any(AlgoliaSearchHelper),
+          indexId: 'indexName',
+          results: expect.any(SearchResults),
+        }),
+      ],
+    });
+    expect(result.current.results.__isArtificial).toEqual(true);
```
sarahdayan, line 92: This is covered in the first test already and unrelated to the test scenario, I don't think we need it twice even if the parameters are slightly different.

packages/react-instantsearch-hooks/src/lib/useSearchResults.ts
```diff
@@ -21,10 +21,15 @@ export function useSearchResults(): SearchResultsApi {
 
   useEffect(() => {
     function handleRender() {
-      setSearchResults({
-        results: searchIndex.getResults()!, // Results can't be `null` after the first render.
-        scopedResults: searchIndex.getScopedResults(),
-      });
+      const results = searchIndex.getResults();
+
+      // results can be null when first search is stalled, in this case we skip the update
```
sarahdayan, line 26: 1. Can you add a permalink to the relevant code from the Helper? This will be useful to understand this portion of code later on when stumbling upon it.

packages/react-instantsearch-hooks/src/lib/useSearchResults.ts
sarahdayan: Why would `useInstantSearch` re-render anyway even if the data returned from `useSearchResults` didn't change?

aymeric-giraudet: Because it [also uses `useSearchState`](https://github.com/algolia/react-instantsearch/blob/a416fc9ec2d28ccc4ea039cbad3cb2c9185afe77/packages/react-instantsearch-hooks/src/hooks/useInstantSearch.ts#L21-L22) which [sets state on every `render` event](https://github.com/algolia/react-instantsearch/blob/a416fc9ec2d28ccc4ea039cbad3cb2c9185afe77/packages/react-instantsearch-hooks/src/lib/useSearchState.ts#L56-L60)

packages/react-instantsearch-hooks/src/lib/useSearchResults.ts
aymeric-giraudet: I think it has to do with `instantsearch.js` and how the index widget works, I added a link to that piece of code which made me understand the issue : https://github.com/algolia/instantsearch.js/blob/20996c7a159988c58e00ff24d2d2dc98af8b980f/src/widgets/index/index.ts#L652-L657

dhayab (approved):
Nice fix!

packages/react-instantsearch-hooks/src/lib/useSearchResults.ts
```diff
@@ -21,10 +21,16 @@ export function useSearchResults(): SearchResultsApi {
...
+      // see: https://github.com/algolia/instantsearch.js/blob/20996c7a159988c58e00ff24d2d2dc98af8b980f/src/widgets/index/index.ts#L652-L657
```
sarahdayan, line 27: We usually wrap comments at 80 chars (unless there's a link) for readability.

```suggestion
      // Results can be `null` when the first search is stalled.
      // In this case, we skip the update.
      // See: https://github.com/algolia/instantsearch.js/blob/20996c7a159988c58e00ff24d2d2dc98af8b980f/src/widgets/index/index.ts#L652-L657
```

packages/react-instantsearch-hooks/src/lib/__tests__/useSearchResults.test.tsx
```diff
@@ -49,4 +53,34 @@ describe('useSearchResults', () => {
     });
     expect(result.current.results.__isArtificial).toBeUndefined();
   });
+
+  test('does not return `null` results when first search is stalled', async () => {
```
sarahdayan, line 57:
```suggestion
  test('does not return `null` results when the first search is stalled', async () => {
```

PR #3599, opened 2022-08-22
feat(SearchBox): introduce `autoFocus` prop

sarahdayan:
## Summary

This PR introduces the `autoFocus` prop on the `<SearchBox>` widget in React InstantSearch Hooks.

This prop exists in all flavors but was missing in React InstantSearch Hooks.

The prop capitalization is different from other flavors (JS, Vue and Angular) because we're relying on React's custom `autoFocus` prop and not the native [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus). This is more natural for React developers who are used to this prop.

Ticket: [FX-1629](https://algolia.atlassian.net/browse/FX-1629)

### DIFF
--- a/examples/hooks/App.tsx
+++ b/examples/hooks/App.tsx
@@ -107,7 +107,7 @@ export function App() {
             ]}
           />
 
-          <SearchBox placeholder="Search" />
+          <SearchBox placeholder="Search" autoFocus />
 
           <div className="Search-header">
             <PoweredBy />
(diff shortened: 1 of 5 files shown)


examples/hooks/App.tsx
```diff
@@ -107,7 +107,7 @@ export function App() {
             ]}
           />
 
-          <SearchBox placeholder="Search" />
+          <SearchBox placeholder="Search" autoFocus />
```
aymeric-giraudet, line 110: Should the example have `autoFocus` ? Seems like the other ones don't

aymeric-giraudet (approved):
Looks good to me !

examples/hooks/App.tsx
sarahdayan: Since this example is the Hooks playground, I find it useful because you can start typing right away, no need to tab to the input.

No strong opinions on this one.

aymeric-giraudet: Well it seems handy so I vote to keep it 😄

dhayab (approved):

Issue #3601, opened 2022-08-22
Documentation and examples for NextJs


Mar0cas:
## 🐛 Bug description

Your example on [CodeSandbox](https://codesandbox.io/s/github/algolia/react-instantsearch/tree/master/examples/next?file=/components/app.js:1036-1046) in this [GitHub examples next doc](https://github.com/algolia/react-instantsearch/tree/master/examples/next), isn't working.

## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Go to https://codesandbox.io/s/github/algolia/react-instantsearch/tree/master/examples/next?file=/components/app.js
2. Click in any filter or input search
3. See console errors

**Live reproduction:**

https://codesandbox.io/s/github/algolia/react-instantsearch/tree/master/examples/next?file=/components/app.js

## 🖥 Screenshots

![image](https://user-images.githubusercontent.com/59584721/185966916-54829aa8-5cb9-4f3a-8bc0-4630f6241847.png)

## Environment

- React InstantSearch version: [e.g. 6.24.0]
- React version: [e.g. 18.1.0]
- Browser: [e.g. Chrome 102.0.5005.61, Safari 15.2]
- OS: [e.g. Windows / Linux / macOS / iOS / Android]

dhayab:
Hi, thanks for your feedback! The sample works well locally, but it is currently broken in Codesandbox because it uses node 10 and some dependencies are not compatible.

I've added some specific settings for Codesandbox so it switches to the right node version.

PR #3602, opened 2022-08-24
chore: explicitly set node version for codesandbox

dhayab:
This PR adds a CodeSandbox config file that explicitly sets node version to 12, to comply with some dependencies requirements.

This PR fixes https://github.com/algolia/react-instantsearch/issues/3601.

[**Sandbox →**](https://codesandbox.io/s/github/algolia/react-instantsearch/tree/fix/example-next-codesandbox/examples/next)

### DIFF
--- a/examples/next/sandbox.config.json
+++ b/examples/next/sandbox.config.json
@@ -1,0 +1,8 @@
+{
+  "infiniteLoopProtection": true,
+  "hardReloadOnChange": false,
+  "view": "browser",
+  "container": {
+    "node": "12"
+  }
+}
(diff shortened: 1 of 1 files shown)


examples/next/sandbox.config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "infiniteLoopProtection": true,
+  "hardReloadOnChange": false,
+  "view": "browser",
```
sarahdayan, line 4: Is this necessary? These are the [defaults](https://codesandbox.io/docs/configuration#sandbox-configuration).

sarahdayan (approved):

examples/next/sandbox.config.json
dhayab: Ah, this is not then, I simply copied the generated config from codesandbox. I'll remove the unnecessary keys, thanks!

dhayab, line 4:
```suggestion
```

PR #3611, opened 2022-09-06
ci(shipjs): update peerDependency on version bump

Haroenv:
This was disappeared when we moved to lerna for bumping the packages.

discovered in #2897


dhayab (approved):
Great, thanks for the fix. Is it the only package that exhibits this behavior (having our lib as a peer dep) or should we glob all the package.json files in the project and look for this?

Haroenv:
There's no other peer dependencies, otherwise the packages have direct dependencies

PR #3612, opened 2022-09-06
ci(circle): remove integration

Haroenv:
These tests don't report correctly, and the service no longer accepts screenshots. We've disabled the argos check a while ago, but apparently didn't remove the step in the pipeline.

This should make the CI use slightly less resources


.circleci/config.yml
```diff
@@ -253,7 +239,6 @@ workflows:
       - test_unit
       - test_unit_react_v17
       - test_unit_algoliasearch_v3
-      - test_integration
```
dhayab, line 256: (unrelated to this specific line but will allow us to have a threaded discussion)

Should we remove the scripts in `package.json` as well as any lingering dependencies (argos-cli for example)?

Haroenv: oh good one, I didn't think of that

dhayab (approved):
Looks good!

PR #3615, opened 2022-09-08
chore(renovate): remove config

Haroenv:
We haven't merged any Renovate PR in a long time, it existing still only wastes bandwidth


sarahdayan (approved):
I'm fine to remove for now but we'll need to find a solution.

PR #3618, opened 2022-09-09
instantsearch-hooks-server: use literal imports

danielbeardsley:
Next.js build process (using pnpm) ends up not seeing these import statements cause the import statements didn't have string literals.

The result was that it didn't put these dependencies in the nested `node_modules` directory and thus this whole package breaks. Notably this only fails in the production build and not the dev build. See [reported bug](https://github.com/vercel/next.js/issues/40067) at next.js.

This change is functionally equivalent, we're just reformatting the code so the literal import statements can be found by whatever tool Next.js is using.

### DIFF
--- a/packages/react-instantsearch-hooks-server/src/getServerState.tsx
+++ b/packages/react-instantsearch-hooks-server/src/getServerState.tsx
@@ -185,21 +185,23 @@ function getInitialResults(rootIndex: IndexWidget): InitialResults {
 function importRenderToString() {
   // React pre-18 doesn't use `exports` in package.json, requiring a fully resolved path
   // Thus, only one of these imports is correct
-  const modules = ['react-dom/server.js', 'react-dom/server'];
+  const modules = [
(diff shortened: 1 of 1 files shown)


Haroenv:
Argh, it's annoying that this code breaks again, as you can see in the linked issue https://github.com/webpack/webpack/issues/13865. We'll need to investigate this deeply to find the right solution, or possibly make the import user-injected (dependency injection) so it can't have any of these issues.

Haroenv:
@danielbeardsley, for a workaround you can use patch-package for now I think, is there a way we can reproduce this outside of Vercel?

DB-Alex:
Maybe my pullrequest will work:

https://github.com/algolia/react-instantsearch/pull/3626

danielbeardsley:
> for a workaround you can use patch-package for now I think,

Yeah, we started using `pnpm patch` and it seems to be working.

PR #3626, opened 2022-09-15
Fix for webpack warning: Critical dependency: the request of a depend…

DB-Alex:
…ency is an expression

Use a static dependency import for react-dom using the same code as next.js:

https://github.com/vercel/next.js/blob/6108f10799b46bc0ef97373c913e23714e135942/packages/next/server/render.tsx#L92

**Summary**

**Result**


Haroenv:
sorry, but this causes https://github.com/webpack/webpack/issues/13865 to reappear. This does indeed give the "critical dependency", but it's actually on purpose as webpack has a bug and doesn't process conditional imports correctly

Haroenv:
When renderToString is passed as an argument, this code path can be bypassed, and the error no longer happens. Thanks for your contribution!

PR #3638, opened 2022-09-26
feat(rish): translations for widgets

aymeric-giraudet:
### [FX-1708](https://algolia.atlassian.net/browse/FX-1708)

**Summary**

- Make components in `widgets` accept an *optional* `translations` prop and forward it to the UI
- All components using `ShowMoreButton` can translate this button too
- Added translations to `RefinementList` for its internal searchbox and for 'no results'

**Result**

Added tests to each widget to check if the translation is taken into account, used mostly `getBy*` rather than snapshots to create less noise in tests


packages/react-instantsearch-hooks-web/src/widgets/Breadcrumb.tsx
```diff
@@ -12,13 +12,16 @@ type UiProps = Pick<
...
+  Omit<UseBreadcrumbProps, 'separator'> & {
+    translations?: UiProps['translations'];
+  };
 
 export function Breadcrumb({
   attributes,
   rootPath,
   separator,
   transformItems,
+  translations = { root: 'Home' },
```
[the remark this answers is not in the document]
aymeric-giraudet: Good call, I thought from a purely translation perspective where if you use this for another language you anyway will have to override all but if it's for alternative texts in English too that would be better indeed

packages/react-instantsearch-hooks-web/src/widgets/Pagination.tsx
```diff
@@ -28,6 +28,18 @@ export function Pagination({
...
+    first: '‹‹',
+    previous: '‹',
+    next: '›',
+    last: '››',
+    page: (currentPage: number) => String(currentPage),
+    ariaFirst: 'First',
+    ariaPrevious: 'Previous',
+    ariaNext: 'Next',
+    ariaLast: 'Last',
+    ariaPage: (currentPage: number) => `Page ${currentPage}`,
```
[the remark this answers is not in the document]
aymeric-giraudet: I'd say it depends on why builders use `translations`, if it's primarily for translating into another language it will just make code more complicated with having to add a switch or a map or whatever conditional

Haroenv (approved):

sarahdayan (changes requested):
A lot of the internal names we picked initially have changed (see the *[*Full breakdown**](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/translations-api-rish.md#full-breakdown) of the RFC). Can you change them in the PR?

packages/react-instantsearch-hooks-web/src/ui/__tests__/ShowMoreButton.test.tsx
```diff
@@ -4,9 +4,19 @@ import React from 'react';
 
 import { ShowMoreButton } from '../ShowMoreButton';
 
+const defaultTranslations = {
+  showMore: 'Show more',
+  showLess: 'Show less',
```
sarahdayan, line 9: Not a fan, because if you want a single text (e.g., "Toggle") you'll have to put it twice.

aymeric-giraudet: But the default for us are 2 strings, I guess users wanting one string for both is a marginal use case while most users just want 2 different strings ?

Haroenv (approved):
👍 no text changed and matches rfc!

aymeric-giraudet:
Thanks @Haroenv ! It matches RFC except for `showMore` which I made into two labels rather than a function, there's a little discussion further up :D

packages/react-instantsearch-hooks-web/src/ui/__tests__/ShowMoreButton.test.tsx
sarahdayan: I don't know about that assumption, we don't really have a good way to know.

In any case, if we want two string translations they should be called `showMoreButtonText` and `showLessButtonText` to fit the [naming scheme](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/translations-api-rish.md#naming) we chose. But then again, there's kind of a mismatch between that and our CSS classes because there's no "show less" button.

I still believe the function is the best pick because the code to write is really straightforward, and the translation key matches a single element, like the others.

If users do internationalization with this API, they will have a lot of logic as well, so I'm not worried that having to write one ternary will be considered a lot of work.

aymeric-giraudet: One last thing I'm wondering is, let's say the developer wants to just display "Toggle", there would be no way for users to know what state the button is in, we would either have to also add a CSS class or an icon, or two strings like "Toggle (⬆)" and "Toggle (⬇️)"

Also in the case of a function should the argument be an object with a single `isShowingMore`  boolean or just directly this boolean ? It's an object in the RFC

aymeric-giraudet: Makes sense, I'll do the same for the functions in `<Pagination>` then :)

packages/react-instantsearch-hooks-web/src/ui/ShowMoreButton.tsx
```diff
@@ -2,13 +2,30 @@ import React from 'react';
...
   isShowingMore,
+  translations,
   ...props
 }: ShowMoreButtonProps) {
   return (
-    <button {...props}>{isShowingMore ? 'Show less' : 'Show more'}</button>
+    <button {...props}>
+      {isShowingMore
+        ? translations.showLessButtonText
+        : translations.showMoreButtonText}
```
Haroenv, line 28: I agree, with the new wording it seems like there's a separate show less button, so I'd make this `showMoreButtonText(isShowingMore)`, probably object arguments for everything (even though it bloats bundlesize)

packages/react-instantsearch-hooks-web/src/ui/__tests__/HierarchicalMenu.test.tsx
```diff
@@ -412,4 +417,23 @@ describe('HierarchicalMenu', () => {
...
+            return isShowingMore ? 'Show less brands' : 'Show more brands';
+          },
+        }}
+      />
+    );
+
+    expect(
+      getByRole('button', { name: 'Show more brands' })
+    ).toBeInTheDocument();
+  });
```
sarahdayan, line 438: Should we re-render with `showMore=false` to test the other side of the ternary?

packages/react-instantsearch-hooks-web/src/ui/__tests__/RefinementList.test.tsx
```diff
@@ -645,4 +650,20 @@ describe('RefinementList', () => {
...
+        showMoreButtonText({ isShowingMore }) {
+          return isShowingMore ? 'Show less brands' : 'Show more brands';
+        },
+      },
+    });
+    const { getByRole } = render(<RefinementList {...props} />);
+
+    expect(
+      getByRole('button', { name: 'Show more brands' })
+    ).toBeInTheDocument();
```
sarahdayan, line 667: Should we re-render with `showMore=false` to test the other side of the ternary?

packages/react-instantsearch-hooks-web/src/ui/__tests__/RefinementList.test.tsx
```diff
@@ -645,4 +650,20 @@ describe('RefinementList', () => {
...
+  });
```
sarahdayan, line 668: Should we test the translations for the SFFV parts?

packages/react-instantsearch-hooks-web/src/ui/__tests__/Menu.test.tsx
```diff
@@ -285,4 +290,23 @@ describe('Menu', () => {
...
+          showMoreButtonText({ isShowingMore }) {
+            return isShowingMore ? 'Show less brands' : 'Show more brands';
+          },
+        }}
+      />
+    );
+
+    expect(
+      getByRole('button', { name: 'Show more brands' })
+    ).toBeInTheDocument();
```
sarahdayan, line 310: Should we re-render with `showMore=false` to test the other side of the ternary?

packages/react-instantsearch-hooks-web/src/widgets/HierarchicalMenu.tsx
```diff
@@ -68,6 +72,12 @@ export function HierarchicalMenu({
     canToggleShowMore,
     onToggleShowMore: toggleShowMore,
     isShowingMore,
+    translations: {
+      showMoreButtonText(options) {
```
sarahdayan, line 76: Nitpick (so feel free to disregard), but `isShowingMore` could be destructured.

packages/react-instantsearch-hooks-web/src/widgets/RefinementList.tsx
```diff
@@ -94,6 +107,16 @@ export function RefinementList({
     }
   }
 
+  const mergedTranslations = {
+    resetButtonTitle: 'Clear the search query.',
+    submitButtonTitle: 'Submit the search query.',
```
sarahdayan, line 112: We didn't add periods in the RFC, and I don't think it fits well with button labels. wdyt?

```suggestion
    resetButtonTitle: 'Clear the search query',
    submitButtonTitle: 'Submit the search query',
```

packages/react-instantsearch-hooks-web/src/widgets/RefinementList.tsx
```diff
@@ -109,16 +132,22 @@ export function RefinementList({
...
+          resetButtonTitle: mergedTranslations.resetButtonTitle,
         }}
       />
     ),
     noResults:
-      searchable && isFromSearch && items.length === 0 && 'No results.',
+      searchable &&
+      isFromSearch &&
+      items.length === 0 &&
+      mergedTranslations.noResults,
```
sarahdayan, line 144: Interesting, we forgot this translation in the RFC.

Do you think `noResults` is fine? If we get closer to the naming scheme we chose, maybe `noResultsText` is more appropriate, wdyt?

packages/react-instantsearch-hooks-web/src/widgets/__tests__/RefinementList.test.tsx
```diff
@@ -1350,4 +1352,69 @@ describe('RefinementList', () => {
...
+          <RefinementList
+            attribute="brand"
+            showMore
+            translations={translations}
+          />
+        </InstantSearchHooksTestWrapper>
+      );
+
+      expect(
+        await findByRole('button', { name: 'Show more brands' })
```
sarahdayan, line 1379: Can we test the other side of the ternary by interacting with the widget?

packages/react-instantsearch-hooks-web/src/widgets/HierarchicalMenu.tsx
aymeric-giraudet: I did not destructure it cause it would shadow the one in the scope above and our ESLint config treats it as an error.
I could destructure and rename but it would be even more verbose

sarahdayan: Ah, didn't notice the variable already existed.

Current code is fine then 🙂

packages/react-instantsearch-hooks-web/src/ui/ShowMoreButton.tsx
```diff
@@ -9,11 +9,9 @@ export type ShowMoreButtonTranslations = {
   /**
    * Alternative text for the show more button.
    */
-  showMoreButtonText: string;
-  /**
-   * Alternative text for the show less button.
-   */
-  showLessButtonText: string;
+  showMoreButtonText: (showMoreButtonTextOptions: {
```
Haroenv:
```suggestion
  showMoreButtonText: (state: {
```

```suggestion
  showMoreButtonText: (options: {
```

```suggestion
  showMoreButtonText: (data: {
```

It may make more sense being the same for every translation

aymeric-giraudet: Yeah I changed it to options later, same for Pagination

packages/react-instantsearch-hooks-web/src/ui/__tests__/RefinementList.test.tsx
aymeric-giraudet: It's not part of the UI component, this one just has translations for the "show more" button, it's the widget that handles searchbox and no results

packages/react-instantsearch-hooks-web/src/widgets/__tests__/Pagination.test.tsx
```diff
@@ -2013,4 +2013,91 @@ describe('Pagination', () => {
...
+            nextPageItemText: 'Next',
+            previousPageItemText: 'Previous',
+            pageItemText: ({ currentPage, nbPages }) =>
+              `#${currentPage}/${nbPages}`,
+          }}
+        />
+      </InstantSearchHooksTestWrapper>
+    );
+
+    expect(container).toMatchInlineSnapshot(`
```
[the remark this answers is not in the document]
aymeric-giraudet: Seems like [testing playground](https://testing-playground.com/gist/15e4470940095d53259514607a518100/aec35e59532ee93c940e57fd899028b7e09421e7) recommends `getByRole` for links, so I used that.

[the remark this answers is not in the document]
sarahdayan: Even better!

packages/react-instantsearch-hooks-web/src/ui/ShowMoreButton.tsx
sarahdayan: `data` > `options` IMO but it's no big deal since this isn't a named one and we can change it whenever 🙂

packages/react-instantsearch-hooks-web/src/ui/ShowMoreButton.tsx
```diff
@@ -2,13 +2,31 @@ import React from 'react';
...
+export type ShowMoreButtonTextOptions = {
+  /**
+   * Whether the widget is showing more items or not.
+   */
+  isShowingMore: boolean;
+};
+
+export type ShowMoreButtonTranslations = {
+  /**
+   * Alternative text for the show more button.
```
sarahdayan, line 17:
```suggestion
   * Alternative text for the "Show more" button.
```

Haroenv (approved):

sarahdayan (approved):
LGTM! Feel free to merge after the suggestion is applied 🙂

Great job 🚀

PR #3648, opened 2022-10-05
chore(examples): update next dep example

FabienMotte:
**Summary**

This PR updates the `next` dependency of the next example to the last version.

**Result**

It should fix the Webpack error we currently have on Codesanbox: https://codesandbox.io/s/github/algolia/react-instantsearch/tree/master/examples/next

image


Haroenv (approved):
I didn't see the error, but this is fine to do :)

PR #3653, opened 2022-10-13
feat(core): support react 18 strict mode

dhayab:
**Summary**

This PR makes React InstantSearch compatible with React 18's Strict Mode.

**Result**

Similar to what we did for React InstantSearch Hooks, cleanup logic is now scheduled instead of immediately executed on unmount, which makes them cancelable when React 18's Strict Mode remounts the components immediately.

[**Sandbox →**](https://codesandbox.io/s/react-instantsearch-app-forked-12ivo6?file=/package.json)


Haroenv (approved):

FabienMotte (approved):
👏

PR #3657, opened 2022-10-19
docs(example-ssr): use cjs imports

Haroenv:
**Summary**

The example is ran using node without esm, and webpack also isn't set up for proper esm, so this errors in modern node (14+) if we use esm.

Setting up for proper esm (type module etc) wouldn't be representative for most users, so we'll fall back to cjs

CR-2214

**Result**

using cjs imports in hooks-ssr example

### DIFF
--- a/examples/hooks-ssr/src/App.js
+++ b/examples/hooks-ssr/src/App.js
@@ -10,8 +10,13 @@ import {
   RefinementList,
   SearchBox,
 } from 'react-instantsearch-hooks-web';
-import { simple } from 'instantsearch.js/es/lib/stateMappings';
-import { history } from 'instantsearch.js/es/lib/routers';
+// because this is ran on node without type: "module" set in the package.json
+// we need to use commonjs instead of esm.
+// If you use ESM in Node, you can use esm instead.
+// import { simple } from 'instantsearch.js/esm/lib/stateMappings';
+// // import { history } from 'instantsearch.js/esm/lib/routers';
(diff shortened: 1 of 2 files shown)


examples/hooks-ssr/src/App.js
```diff
@@ -10,8 +10,13 @@ import {
   RefinementList,
   SearchBox,
 } from 'react-instantsearch-hooks-web';
-import { simple } from 'instantsearch.js/es/lib/stateMappings';
-import { history } from 'instantsearch.js/es/lib/routers';
+// because this is ran on node without type: "module" set in the package.json
+// we need to use commonjs instead of esm.
+// If you use ESM in Node, you can use esm instead.
+// import { simple } from 'instantsearch.js/esm/lib/stateMappings';
+// // import { history } from 'instantsearch.js/esm/lib/routers';
```
dhayab, line 17:
```suggestion
// If you use ESM in Node, you can rely on these import statements instead:
// import { simple } from 'instantsearch.js/es/lib/stateMappings';
// import { history } from 'instantsearch.js/es/lib/routers';
```

dhayab (changes requested):

examples/hooks-ssr/src/App.js
Haroenv: good catch lol, thanks @dhayab

dhayab (approved):

PR #3658, opened 2022-10-19
feat(getServerState): allow users to inject renderToString

### LINKED ISSUE #3633: Could not import ReactDOMServer
## 🐛 Bug description

```
ERROR	Error: Could not import ReactDOMServer.
    at file:///var/task/node_modules/.pnpm/react-instantsearch-hooks-server@6.33.0_fzyfqr3mixyxftdbf7mmfo2jb4/node_modules/react-instantsearch-hooks-server/dist/es/getServerState.js:166:13
    at async getServerSideProps (/var/task/site/.next/server/pages/algolia.js:98:25)
    at async Object.renderToHTML (/var/task/node_modules/.pnpm/next@12.3.0_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/server/render.js:512:20)
    at async doRender (/var/task/node_modules/.pnpm/next@12.3.0_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/server/base-server.js:642:38)
    at async cacheEntry.responseCache.get.isManualRevalidate.isManualRevalidate (/var/task/node_modules/.pnpm/next@12.3.0_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/server/base-server.js:747:28)
    at async /var/task/node_modules/.pnpm/next@12.3.0_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/server/response-cache/index.js:80:36 {  page: '/algolia'}
RequestId: 3de89872-78cc-4717-adcf-7b57a96a1de9 Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

I'm using react-instantsearch-hooks-server with Next.js (vercel/commerce) and I have this error only in production (Vercel), in local dev environment it works as expected.
The error started to appear with React 18, with 17 it worked.

## 🔍 Bug reproduction
I tried to reproduce the bug in a codesandbox but it doesn't happen.
I've setup a repo with an example at https://github.com/slackerzz/vercel-commerce
It is a clean vercel/commerce with a route [`/algolia`](https://vercel-commerce-pnpm.vercel.app/algolia) where i use the ssr provider.
The code is availabel at https://github.com/slackerzz/vercel-commerce/blob/main/site/pages/algolia.tsx
The index used is from https://codesandbox.io/s/github/algolia/react-instantsearch/tree/master/examples/hooks-next

In a local environment running `pnpm dev` the `/algolia` route works as expected:

Schermata 2022-09-23 alle 14 40 16

After deploying the app to Vercel the same route returns a 500 error

## Environment

- React InstantSearch Hooks version: 6.33
- React version: 18.2.0
- Next.js version: 12.3.1

Haroenv:
**Summary**

There are some cases where the combination of trying to make sure renderToString doesn't end up in a browser bundle, being runnable on esm + cjs, react 17 and 18, .js extension etc. blows up.

One of those is pnpm/vercel removing "unused" packages.

**Result**

This PR introduces a new argument `renderToString` to `getServerState` so you can inject the dependency yourself, meaning the import is within your own code and won't be purged.

```js
import { renderToString } from 'react-dom/server';
await getServerState(<App/>, renderToString);
await getServerState(<App/>, import('react-dom/server').then(mod => mod.renderToString));
```

fixes #3633
closes #3618
see vercel/next.js#40067
FX-1869

### DIFF
--- a/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
+++ b/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
@@ -114,8 +114,19 @@ describe('ReactDOMServer imports', () => {
     await expect(
       getServerState(<App />)
     ).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"Could not import ReactDOMServer."`
+      `"Could not import ReactDOMServer. You can provide it as an argument: getServerState(<Search />, ReactDOMServer.renderToString)."`
     );
+  });
+
+  test('does not throw if user provides their own renderToString', async () => {
+    const searchClient = createSearchClient({});
+    const { App } = createTestEnvironment({ searchClient });
+
+    const serverState = await getServerState(<App />, (element) =>
+      jest.requireActual('react-dom/server').renderToString(element)
+    );
+
+    expect(serverState.initialResults).toEqual(expect.any(Object));
   });
 });
 
--- a/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
+++ b/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
@@ -118,7 +118,7 @@ describe('ReactDOMServer imports', () => {
     );
   });
 
-  test('does not throw if user provides their own renderToString', async () => {
+  test('calls the provided renderToString function', async () => {
     const searchClient = createSearchClient({});
     const { App } = createTestEnvironment({ searchClient });
 
--- a/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
+++ b/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
@@ -122,9 +122,10 @@ describe('ReactDOMServer imports', () => {
     const searchClient = createSearchClient({});
     const { App } = createTestEnvironment({ searchClient });
 
-    const serverState = await getServerState(<App />, (element) =>
-      jest.requireActual('react-dom/server').renderToString(element)
-    );
+    const serverState = await getServerState(<App />, {
+      renderToString: (element) =>
+        jest.requireActual('react-dom/server').renderToString(element),
+    });
 
     expect(serverState.initialResults).toEqual(expect.any(Object));
   });
--- a/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
+++ b/packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
@@ -114,7 +114,7 @@ describe('ReactDOMServer imports', () => {
     await expect(
       getServerState(<App />)
     ).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"Could not import ReactDOMServer. You can provide it as an argument: getServerState(<Search />, ReactDOMServer.renderToString)."`
+      `"Could not import ReactDOMServer. You can provide it as an argument: getServerState(<Search />, { renderToString })."`
     );
   });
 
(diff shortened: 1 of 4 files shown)


aymeric-giraudet (approved):

francoischalifour:
Don't we want to be more cautious and have it in a param object?

```js
await getServerState(<App/>, { renderToString });
```

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -8,19 +8,21 @@ import {
...
 import type { ReactNode } from 'react';
-import type { renderToString as reactRenderToString } from 'react-dom/server';
 import type {
   InstantSearchServerContextApi,
   InstantSearchServerState,
 } from 'react-instantsearch-hooks';
 
 type SearchRef = { current: InstantSearch | undefined };
 
+export type RenderToString = (element: JSX.Element) => unknown;
```
dhayab, line 18:
```suggestion
export type RenderToString = (element: JSX.Element) => string;
```

packages/react-instantsearch-hooks-server/src/getServerState.tsx
```diff
@@ -8,19 +8,21 @@ import {
...
+
 /**
  * Returns the InstantSearch server state from a component.
  */
 export function getServerState(
-  children: ReactNode
+  children: ReactNode,
+  renderToString?: RenderToString
```
dhayab, line 25: Should we add a comment to make it mandatory next major version?

dhayab (commented):
Neat idea!

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: I left unknown as you should be able to use this argument with renderToStream or something as well, because we don't use the string

dhayab: `renderToReadableStream()` will probably create issues for example. But as long as its typed as a synchronous method it should be fine indeed.

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: should there also be a warning if it's not given (knowing that it does work in most cases)?

dhayab: I was not for it initially, but maybe it's best to warn early since it will be a quick adjustment on those who update the library, and will remove the issues outlined in the description.

packages/react-instantsearch-hooks-server/src/getServerState.tsx
Haroenv: I was more thinking if there's in theory another method that traverses the tree, but doesn't render, it is also sufficient for this method

packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
```diff
@@ -114,9 +114,21 @@ describe('ReactDOMServer imports', () => {
...
+    const searchClient = createSearchClient({});
+    const { App } = createTestEnvironment({ searchClient });
+
+    const serverState = await getServerState(<App />, {
+      renderToString: (element) =>
+        jest.requireActual('react-dom/server').renderToString(element),
+    });
+
+    expect(serverState.initialResults).toEqual(expect.any(Object));
+  });
```
francoischalifour, line 131: Here we could rather test that the provided `renderToString` is called when calling `getServerState(<App />, { renderToString })`.

Asserting that it doesn't throw doesn't add as much value.

packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx
```diff
@@ -114,9 +114,25 @@ describe('ReactDOMServer imports', () => {
     await expect(
       getServerState(<App />)
     ).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"Could not import ReactDOMServer."`
+      `"Could not import ReactDOMServer. You can provide it as an argument: getServerState(<Search />, { renderToString })."`
     );
   });
+
+  test('does not throw if user provides their own renderToString', async () => {
```
francoischalifour, line 121: It better describes what the test does:

```suggestion
  test('calls the provided renderToString function', async () => {
```

francoischalifour (approved):

dhayab (approved):

PR #3665, opened 2022-10-27
fix(PoweredBy): add margin between text and Algolia logo

FabienMotte:
**Summary**

This PR adds a margin between text and logo to the `<PoweredBy>`  widget component for React InstantSearch.

As you can see on [this CodeSandbox](https://codesandbox.io/s/react-instantsearch-app-forked-nzgn0t), there is no margin:

image

We don't have this issue in the [RIS Storybook](https://react-instantsearch.netlify.app/storybook/?path=/story/poweredby--default) because a margin is added by the [Algolia theme](https://github.com/algolia/instantsearch-specs/blob/3044c2b7c99e7637403c8a2ea2ff70ec4d112f95/src/styles/themes/algolia.scss#L464-L466) while the CodeSanbox above uses the Satellite theme.

**Result**

A `margin-left: 0.3rem` is added on the SVG, overriding the class property added by the Algolia theme if present (like on Storybook).

image

**Questions**

Should we remove the class property in the [Algolia theme](https://github.com/algolia/instantsearch-specs/blob/3044c2b7c99e7637403c8a2ea2ff70ec4d112f95/src/styles/themes/algolia.scss#L464-L466) since it will now be overridden by the style property?


Haroenv (approved):
not ideal, but better than leaving it as is I think

sarahdayan (changes requested):
I don't think we should hardcode styles in the component, this is a CSS concern and should belong to the themes.

FabienMotte:
> I don't think we should hardcode styles in the component, this is a CSS concern and should belong to the themes.

@sarahdayan Another solution I see: should we delete the "Search by" text in RIS and use the PoweredBy logo (that includes the text in the SVG) so we don't have the spacing issue anymore and RIS is aligned with the [InstantSearch specs](https://instantsearch-css.netlify.app/widgets/powered-by/)?

**Note:** I need to update the specs for the PoweredBy widget since the logo has changed

Haroenv:
That would be a breaking change for people who have changed the text though, as it would revert to the English / original text. I think we should do either this or nothing (nothing being a fine option)

FabienMotte:
@Haroenv Indeed, I forgot that part is translatable. thanks!

Closing this PR in favor of [instantsearch-specs#127](https://github.com/algolia/instantsearch-specs/pull/127)

PR #3690, opened 2022-11-28
wip: introduce parentIndexId

Haroenv:
This would be useful for cases like https://github.com/algolia/react-instantsearch/discussions/3689#discussioncomment-4251317 where the cascade is working against your implementation.

The reason this is WIP is because I think without having a bare root (without the workaround in App.tsx) you still need to prevent searches on the main index and that's still fairly "ugly". If InstantSearch would have an option to prevent mainIndex from searching (so actually having an array of root indices instead of a single one) it would be much more useful

### DIFF
--- a/examples/hooks/App.tsx
+++ b/examples/hooks/App.tsx
@@ -7,6 +7,7 @@ import {
   Configure,
   ClearRefinements,
   CurrentRefinements,
+  DynamicWidgets,
   HierarchicalMenu,
   Highlight,
   Hits,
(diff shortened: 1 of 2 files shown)


examples/hooks/App.tsx
```diff
@@ -61,119 +62,131 @@ export function App() {
...
-            <ClearRefinements />
-            <CurrentRefinements
-              transformItems={(items) =>
-                items.map((item) => {
-                  const label = item.label.startsWith('hierarchicalCategories')
-                    ? 'Hierarchy'
-                    : item.label;
+            <Index
+              indexName="autocomplete_demo_products_query_suggestions"
+              parentIndexId="instant_search"
```
Haroenv, line 154: more ideally `parentIndexId={null}` would behave as another mainIndex, but not sure how that can be done. It could of course be implemented later as well

francoischalifour: I think we can resolve `null` to `useInstantSearchContext().indexName`.

Haroenv: yep, null could be mainIndex, but it wouldn't really solve the issue here (needing to have a manual index at the root with different indexId), as that mainIndex would probably have widgets as well, negating the point of parentId a bit (I could almost only imagine use cases where you want _no_ parameters to be inherited)

packages/react-instantsearch-hooks/src/lib/useIndex.ts
```diff
@@ -37,3 +43,39 @@ export function useIndex(props: UseIndexProps) {
 
   return indexWidget;
 }
+
+function useParentIndex({ parentIndexId }: UseIndexOwnProps) {
+  const physicalParentIndex = useIndexContext();
+  const search = useInstantSearchContext();
+
+  if (parentIndexId) {
```
Haroenv, line 51: if we want to support null == mainIndex:

```suggestion
  if (parentIndexId === null) {
    return searchMainIndex;
  }
  if (parentIndexId) {
```

examples/hooks/App.tsx
francoischalifour: I wonder how easier the mental model would be if we require an `<Index>` within `<InstantSearch>`.

Because we can't really portal an `<Index>` to an empty Helper in the current state. The root is shared among all `<Index>`.

(I see this problem a bit like Element and Fragment.)

Haroenv: Yep, that's what I had in mind as well, possibly that could be detected when the InstantSearch instance doesn't receive an indexName prop, you render Index manually, and thus will be able to have adjacent root indices

